PR #105, opened 2015-06-26
Add gpio control and related codes.

seanshpark:


seanshpark:
Close this as new push has lost with this one

Issue #118, opened 2015-07-02
Match JerryScript types


seanshpark:
With latest jerry, we need to fix warnings

1) 
```
iotjs_module_console.cpp:55:34: warning: invalid conversion from 
‘bool (*)(const JRawObjectType*, const JRawValueType*, iotjs::JRawValueType*, const JRawValueType*, uint16_t) {aka 
bool (*)(const ecma_object_t*, const jerry_api_value_t*, jerry_api_value_t*, const jerry_api_value_t*, short unsigned int)}’ to ‘iotjs::JHandlerType {
aka bool (*)(const ecma_object_t*, const jerry_api_value_t*, jerry_api_value_t*, const jerry_api_value_t*, unsigned int)}’ [-fpermissive]
```

2) 
```
iotjs_binding.h:128:8: warning:   
initializing argument 2 of ‘void iotjs::JObject::SetMethod(const char*, iotjs::JHandlerType)’ [-fpermissive]
```

seanshpark:
close issue as #119 landed

PR #119, opened 2015-07-06
Change character type to uint8_t

seanshpark:
This commit is to apply latest character type change in JerryScript as uint8_t
- Update to latest JerryScript
- add new jschar type as uint8_t and change char types to this
- disable JObject(const char*) constructor which makes problem
- update JHandlerType for new jerry_api_length_t

Related issue #118

### DIFF
--- a/src/iotjs_binding.h
+++ b/src/iotjs_binding.h
@@ -27,6 +27,7 @@ typedef jerry_external_handler_t JHandlerType;
 typedef jerry_object_free_callback_t JFreeHandlerType;
 typedef jerry_api_object_t JRawObjectType;
 typedef jerry_api_value_t JRawValueType;
+typedef jerry_api_length_t JRawLengthType;
 
 
 
@@ -61,7 +62,7 @@ class JObject {
   explicit JObject(double v);
 
   // Creates a javascirpt number object.
-  explicit JObject(const char* v);
+  explicit JObject(const jschar* v);
 
   // Creates a object from `JRawObjectType*`.
   // If second argument set true, then ref count for the object will be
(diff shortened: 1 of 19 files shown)


src/iotjs_util.cpp
```diff
@@ -45,19 +65,19 @@ char* ReadFile(const char* path) {
 }
 
 
-char* AllocBuffer(size_t size) {
-  char* buff = static_cast<char*>(malloc(size));
+jschar* AllocBuffer(size_t size) {
```
ILyoan: `AllocBuffer` is not for creating string buffer, it is for creating byte buffer(creating `size` of octets). Thus changed prototype is not correctly describe the purpose of this function.

src/iotjs_util.cpp
```diff
@@ -45,19 +65,19 @@ char* ReadFile(const char* path) {
...
+  jschar* buff = static_cast<jschar*>(malloc(size));
   memset(buff, 0, size);
   return buff;
 }
 
 
-char* ReallocBuffer(char* buffer, size_t size) {
-  return static_cast<char*>(realloc(buffer, size));
+jschar* ReallocBuffer(jschar* buffer, size_t size) {
```
ILyoan: Hero also.

src/iotjs_util.cpp
```diff
@@ -45,19 +65,19 @@ char* ReadFile(const char* path) {
...
+  return static_cast<jschar*>(realloc(buffer, size));
 }
 
 
-void ReleaseBuffer(char* buffer) {
+void ReleaseBuffer(jschar* buffer) {
```
ILyoan: Here also

src/iotjs_types.h
```diff
@@ -0,0 +1,26 @@
...
+ */
+
+
+#ifndef IOTJS_TYPES_H
+#define IOTJS_TYPES_H
+
+
+typedef uint8_t jschar;  /* follow jerry_api_char_t type which is uint8_t */
+
+#define _JSCT (jschar*)
```
ILyoan: Looks like this macro is tempting writing codes which does not look follow C grammar. for example
```
foo(_JSCT iotjs);
bar(_JSCT "bar");
```

What about function like macro such as:
```
#define JSCT(a) static_cast<jschar*>(a)

...

foo(JSCT(iotjs));
bar(JSCT("bar"));
```

src/iotjs_module_buffer.h
```diff
@@ -42,16 +42,16 @@ class BufferWrap : public JObjectWrap {
...
   int Compare(const BufferWrap& other) const;
 
-  size_t Copy(char* src, size_t len);
-  size_t Copy(char* src, size_t src_from, size_t src_to, size_t dst_from);
+  size_t Copy(jschar* src, size_t len);
+  size_t Copy(jschar* src, size_t src_from, size_t src_to, size_t dst_from);
 
  protected:
-  char* _buffer;
+  jschar* _buffer;
```
ILyoan: `Buffer` is for array of bytes, not for javascript string. Thus `_buffer` should be a type representing array of bytes.

src/iotjs_binding.h
```diff
@@ -150,10 +151,10 @@ class JObject {
   // Returns pontiner to null terminated string contents of string object.
   // Returned pointer should be released using `ReleaseCString()` when it become
   // unnecessary.
-  char* GetCString();
+  jschar* GetCString();
```
ILyoan: This function should return type of C String not javascript string.

ILyoan: Or I think we should rename it

src/iotjs_types.h
seanshpark: Ok, looks better

src/iotjs_binding.h
seanshpark: Rename GetCString() to GetByteString() seems nice.

ILyoan:
@seanshpark  
I think we need to introduce kind of String wrapper.
Before it, let's go with this.

Let's make it merged

seanshpark:
OK

PR #120, opened 2015-07-08
Remove js wrapper

ILyoan:
* Remove duplicated module wrapper of javascript side.
* Use `JSON.parse()` for parsing JSON file.


seanshpark:
LGTM

Issue #122, opened 2015-07-08
Fix for nuttx build


seanshpark:
Fix make when updating jerry: https://github.com/Samsung/jerryscript/pull/330
* limit heap for nuttx is needed

seanshpark:
PR #125

seanshpark:
close as PR landed

Issue #110, opened 2015-06-30
Hardware control APIs


seanshpark:
As of today, nodejs is on the discuss to support [hardware](https://github.com/nodejs/hardware) and libuv is ready with [uv_device_t PR](https://github.com/libuv/libuv/pull/379). 
For IoT.js, how are we going to support hardware control and manage devices?
First PR #105 needs some more adjustments and we need to discuss how to provide the API and manage H/W boards and devices. [GPIO-control-ideas](https://github.com/Samsung/iotjs/wiki/GPIO-control-ideas) is prepared to draw everyones sketch.
Please add comments and write down to WiKi page.

seanshpark:
As wrote in PR #113, process.iotjs can be used for distinguishing between iotjs and others. 
it may also have other properties, such as running product class, H/W board, and network or other environments that is needed to make application to decide better.
If this seems to be a bad idea, please add comments.

ILyoan:
I think we may need API reference for device control.
And for new module features will be introduced later on, how about consensus of API first and then implementation?

ILyoan:
https://github.com/Samsung/iotjs/wiki/IoT.js-API:-GPIO

ILyoan:
What type of mode could be exist for a pin of GPIO?

As far as I know there is only one; direction. Direction of a pin could be either input or output at one time thus rw pin is not possible right?

Is there any other type of mode besides direction?

seanshpark:
I've prepared a short page https://github.com/Samsung/iotjs/wiki/GPIO-control-ideas few days ago, and pin mode is described in summary. If you jump to link http://coactionos.com/embedded%20design%20tips/2013/10/21/Tips-Understanding-Microcontroller-Pin-Input-Output-Modes/ there is more. Hope this helps.

ILyoan:
@seanshpark The page you referred are only saying there are input and output mode(direction). Then can we assume that there are no other type of mode?

seanshpark:
@ILyoan , If you let mode to be input/output, one more thing is `floating` that is something like `disconnected`.

> Is there any other type of mode besides direction?

Besides direction, to the pin, 
* Pull up or down or open-drain(open collector)
* Rise up or down edge detection
* input in digital or analog

ILyoan:
Let's summarize what pin configuration are commonly definable so we can design our GPIO API more usable for anyone.

I see configurable options are:

### Direction

#### input
* floating
* pull-up
* pull-down

#### output
* open-drain
[...]

PR #125, opened 2015-07-14
Add arm-linux build for RPi2 and fix for nuttx build

seanshpark:

also fix for nuttx build, as jerry has removed nuttx and fixed to use external.


cmake/config/arm-linux.cmake
```diff
@@ -0,0 +1,42 @@
...
+set(CMAKE_SYSTEM_NAME Linux)
+set(CMAKE_SYSTEM_PROCESSOR armv7l-hf)
+
+set(EXTERNAL_CMAKE_C_COMPILER arm-linux-gnueabihf-gcc)
+set(EXTERNAL_CMAKE_CXX_COMPILER arm-linux-gnueabihf-g++)
+
+CMAKE_FORCE_C_COMPILER(${EXTERNAL_CMAKE_C_COMPILER} GNU)
+CMAKE_FORCE_CXX_COMPILER(${EXTERNAL_CMAKE_CXX_COMPILER} GNU)
+
+set(DEVICE_DEPENDS "linux-arm-rpi2")
```
ILyoan: Isn't this file for "arm-linux" ? 
Specifying "rpi2" in this file seems not appropriate.

tools/build.py
```diff
@@ -275,6 +275,8 @@ def build_libuv():
         # libuv is using gyp. run the system according to build target.
         if opt_target_arch() == 'arm' and opt_target_os() =='nuttx':
             check_run_cmd('./nuttx-configure', [opt_nuttx_home()])
+        elif opt_target_arch() == 'arm' and opt_target_os() =='linux':
+            check_run_cmd('./armlinux-configure')
```
ILyoan: Where is configure file?

ILyoan: Ah, ok I see this is for uv configuration. so the file should have been in the uv source tree. That good.

src/device/linux-arm-rpi2/gpioctl_linux-arm-rpi2.cpp
```diff
@@ -0,0 +1,295 @@
...
+  uint8_t gpionum = rpi2_get_gpio(portpin & 0xff);
+  snprintf(gpio_file_buff, GPIO_RPI2_MAX_PATH, gpio_valfmt, gpionum);
+  int fd = open_device(gpio_file_buff, false);
+  if (fd < 0) {
+    DLOG("RPi2 GPIO, failed to read %d", gpionum);
+    return IOTJS_GPIO_READFAIL;
+  }
+
+  int ret = 0, rd;
+  rd = read(fd, gpio_file_buff, 1);
```
ILyoan: Here also `read(fd, pdata, sizeof(*pdata))` ?

src/device/linux-arm-rpi2/gpioctl_linux-arm-rpi2.cpp
```diff
@@ -0,0 +1,295 @@
...
+
+  snprintf(gpio_file_buff, GPIO_RPI2_MAX_PATH, gpio_valfmt, gpionum);
+  int fd = open_device(gpio_file_buff, true);
+  if (fd < 0) {
+    DLOG("RPi2 GPIO, failed to write %d", gpionum);
+    return IOTJS_GPIO_WRITEFAIL;
+  }
+
+  int ret = 0;
+  ssize_t written = write(fd, data ? val_1 : val_0, 1);
```
ILyoan: Why do we have to write only "0" or "1" instead of given data.
Are there problems with `write(fd, &data, sizeof(data));`?

src/device/linux-arm-rpi2/gpioctl_linux-arm-rpi2.cpp
```diff
@@ -0,0 +1,295 @@
...
+#define GPIO_RETRY_WAIT           1000      // in microseconds
+
+int GpioControlInst::open_device(const char* devicefile, bool wmode) {
+  int fd = -1;
+  int retry = GPIO_OPEN_RETRY;
+  int sleepvalue = GPIO_RETRY_WAIT;
+  int flags = wmode ? O_WRONLY : O_RDONLY;
+
+  // creating virtual file for device control takes some time
+  while (fd < 0 && retry > 0) {
```
ILyoan: I think it takes too long time to give up retrying. it's almost 8 min?

ILyoan: Ooops, sorry it takes 8/1000 min.

ILyoan: It's about 0.5 secs. Isn't it too short ?

cmake/config/arm-linux.cmake
seanshpark: some where we need to specify rpi2 cause of gpio

src/device/linux-arm-rpi2/gpioctl_linux-arm-rpi2.cpp
seanshpark: https://www.kernel.org/doc/Documentation/gpio/sysfs.txt
'0' or '1' is used for port writing.

src/device/linux-arm-rpi2/gpioctl_linux-arm-rpi2.cpp
seanshpark: I think 0.5 sec for RPi2 is ok, for other boards there will be signal specification that we can specify.

cmake/config/arm-linux.cmake
ILyoan: That true but I don't think this file is correct place.

ILyoan:
I think we need to discuss GPIO API in #110.
After reach a consensus, development of GPIO will be much straightforward.

ILyoan:
We have gpio api for example, 'gpio.write(pin, data, callback).
GPIO backend for stm32 write given `data`. but backend for RPI2 write '0' if the given `data` is 0 or write '1'. I think this behavior does not guarantee for user consistency among various hardware.

cmake/config/arm-linux.cmake
seanshpark: ok, do you have a place in mind? I can change.

seanshpark:
> GPIO backend for stm32 write given data

yes, for STM32, explicit set data to 0 or 1 seems better to match digital I/O. the code is as is for analog value (0~255) also, that may not be same as RPi2. 
right now I'm not sure about the values and my thought is to change while use case development.

seanshpark:
Changed this PR to reflect #122(https://github.com/Samsung/jerryscript/pull/330) and https://github.com/Samsung/jerryscript/pull/345.
GPIO related implemetation(#110) will be in next PR

cmake/config.cmake
```diff
@@ -14,6 +14,24 @@
...
 cmake_minimum_required(VERSION 2.8)
 
+string(TOLOWER ${CMAKE_SYSTEM_NAME} CFG_SYS_NAME)
+string(TOLOWER ${CMAKE_SYSTEM_PROCESSOR} CFG_SYS_PROCESSOR)
+
+if(${CFG_SYS_NAME} STREQUAL "linux")
+  if(${CFG_SYS_PROCESSOR} STREQUAL "x86" OR
+     ${CFG_SYS_PROCESSOR} STREQUAL "x86_64")
+    set(DEVICE_DEPENDS "linux-x86")
+  elseif(${CFG_SYS_PROCESSOR} STREQUAL "armv7l-hf")
```
ILyoan: "armv7l-hf" just tells us the cpu architecture, we could not determine that this is Raspberry PI2 ?

cmake/config.cmake
```diff
@@ -14,6 +14,24 @@
...
+    set(DEVICE_DEPENDS "linux-arm-rpi2")
```
ILyoan: What about just use "arm-linux"?
Are there special implementation expected for Raspberry PI 2?

cmake/config.cmake
```diff
@@ -14,6 +14,24 @@
 
 cmake_minimum_required(VERSION 2.8)
 
+string(TOLOWER ${CMAKE_SYSTEM_NAME} CFG_SYS_NAME)
+string(TOLOWER ${CMAKE_SYSTEM_PROCESSOR} CFG_SYS_PROCESSOR)
+
+if(${CFG_SYS_NAME} STREQUAL "linux")
+  if(${CFG_SYS_PROCESSOR} STREQUAL "x86" OR
+     ${CFG_SYS_PROCESSOR} STREQUAL "x86_64")
+    set(DEVICE_DEPENDS "linux-x86")
```
ILyoan: "x86-linux" is preferable? as many system use <arch-os-abi> target triple.

cmake/config.cmake
seanshpark: yes, we need board dependent implementation. at this time there is only one arm-linux and it's RPi2, but later there might be others. I'm not sure about how to extend this right now. I hope you can give a way to do this.

cmake/config.cmake
seanshpark: yes, gpio implementation  and it will be board dependent.

seanshpark: name arm-linux looks better.

ILyoan: I thought previous GPIO implemention for RPi2 was using GPIO sysfs supported by linux kernel. If it was right, it just depends on OS not board ?

seanshpark:
board specific names are removed. when implementation differs we can add compile directives.

ILyoan:
@seanshpark  OK thanks. I see numbers likely are different from board to board even if we use the same implementation for different boards so we need board directive as you mentioned. I'll modify build script for adding board directive.

ILyoan:
@seanshpark I think last commit is enough to be merged. as we can adjust trivial issues from following updates. Let's make it merged.
Thanks

PR #127, opened 2015-07-20
Add target board build directive

ILyoan:
This PR is to add "target_board" directive for build system.

* "build.py" provides an option "target_board" to get target board.
* `TARGET_BOARD` will be set by the given directive, if not fed `"unknown"`.
* Applications are able to get the board by accessing `process.iotjs.board` .


seanshpark:
LGTM

PR #129, opened 2015-07-21
gpio implementation discussion with code

seanshpark:
this PR is not for real but for gpio implementation discussion.
please leave comments.
* related issue: Implement GPIO class,  #128

PR #131, opened 2015-07-23
Start of GPIO implementation based on GPIO API

seanshpark:
* applied comments from #129
* setPin() API ready, others are blank
* x86 native dummy code with setPin() for testingy, others are blank
    
### DIFF
--- a/src/device/x86-linux/gpioctl_x86_linux.cpp
+++ b/src/device/x86-linux/gpioctl_x86_linux.cpp
@@ -13,25 +13,133 @@
  * limitations under the License.
  */
 
-#if defined(__LINUX__)
-
 #include "iotjs_def.h"
 #include "iotjs_objectwrap.h"
 #include "iotjs_module_gpioctl.h"
 
+//
+// this is a dummy emulator code for x86.
+// please add new 'device' file for your purpose.
+// use timer with 1msec delay to implement async callback.
+// real gpio implementation would be better to use threading
+//
 
 namespace iotjs {
 
-/*
- * Use default dummy for linux-x86
- */
+
+// type for delayed callback
+typedef enum {
+  DELAYEDCALL_SETPIN = 1,
+} DelayedCallType;
+
+
+// base class for delayed callback
+struct DelayedCallBase {
+  DelayedCallType type;
+  uv_timer_t timer;
+};
+
+
+// setspin delayed callback class
+struct DelayedCallSetpin : public DelayedCallBase {
+  GpioCbDataSetpin* data;
+  GpioSetpinCb aftersetpin;
+  int result;
+};
+
+
+static void timerHandleTimeout(uv_timer_t* handle) {
+  uv_timer_stop(handle);
+
+  DelayedCallBase* base;
+  base = reinterpret_cast<DelayedCallBase*>(handle->data);
+  switch (base->type) {
+    case DELAYEDCALL_SETPIN : {
+      DelayedCallSetpin* dcsp =
+                            reinterpret_cast<DelayedCallSetpin*>(base);
+      GpioCbDataSetpin* setpin =
+                            reinterpret_cast<GpioCbDataSetpin*>(dcsp->data);
+
+      DDDLOG("x86 linux gpio dummy setpin w/cb pin(%d), dir(%d), mode(%d)",
+             setpin->pin, setpin->dir, setpin->mode);
+
+      dcsp->result = (setpin->pin >= 0) ? 0 : GPIO_ERR_INVALIDPARAM;
+      dcsp->aftersetpin(dcsp->data, dcsp->result);
+      delete dcsp;
+      break;
+    }
+    default:
+      IOTJS_ASSERT(false);
+      break;
+  }
+}
+
+
+//-----------------------------------------------------------------------------
+
+class GpioControlImpl : public GpioControl {
+public:
+  explicit GpioControlImpl(JObject& jgpioctl);
+
+  virtual int Initialize(void);
+  virtual void Release(void);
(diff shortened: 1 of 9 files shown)


src/device/x86-linux/gpioctl_x86_linux.cpp
```diff
@@ -13,25 +13,128 @@
...
 namespace iotjs {
 
-/*
- * Use default dummy for linux-x86
- */
+
+// type for delayed callback
+typedef enum {
+  DELAYEDCALL_SETSPIN = 1,
+} dc_type;
```
ILyoan: Type name should be UpperCamelCase
https://github.com/Samsung/iotjs/wiki/Coding_Style_Guideline#type-names

src/js/gpio.js
```diff
@@ -47,47 +51,140 @@ GPIO.release = function() {
...
+    result = gpioctl.ERR_INVALIDPARAM;
   }
   else {
-    return err;
+    result = gpioctl.setPort(portNumber, dircode, mode, callback);
+  }
+  if (result < 0) {
+    err = new GpioError(result, gpioctl.errMessage(result));
+    if (util.isFunction(callback)) {
+      callback(err);
```
ILyoan: Isn't this resulting double callback ?

src/js/gpio.js
```diff
@@ -47,47 +51,140 @@ GPIO.release = function() {
...
+  }
+};
+
+
+GPIO.writePort = function(portNumber, value, callback) {
+  var result = gpioctl.writePort(portNumber, value, callback);
+  if (result < 0) {
+    err = new GpioError(result, gpioctl.errMessage(result));
+    if (util.isFunction(callback)) {
+      callback(err);
```
ILyoan: Isn't this resulting double callback ?

src/js/gpio.js
```diff
@@ -47,47 +51,140 @@ GPIO.release = function() {
...
 
+GPIO.readPort = function(portNumber, callback) {
+  var result = 0;
+  if (!util.isFunction(callback)) {
+    result = gpioctl.ERR_INVALIDPARAM;
+    throw new GpioError(result, gpioctl.errMessage(result));
+  }
+  result = gpioctl.readPort(portNumber, callback);
+  if (result < 0) {
+    callback(err, false);
```
ILyoan: Isn't this resulting double callback ?

src/device/x86-linux/gpioctl_x86_linux.cpp
```diff
@@ -13,25 +13,128 @@
...
+}
+
+
+// SetPin for callback is provided
+int GpioControlImpl::SetPin(gpio_setpin_t* setpin_data, gpio_setpin_cb cb) {
+  DDDLOG("x86 linux gpio dummy setpin w/cb pin(%d), dir(%d), mode(%d)",
+         setpin_data->pin, setpin_data->dir, setpin_data->mode);
+
+  Environment* env = Environment::GetEnv();
+  DelayedCallSetSpin* delay = new DelayedCallSetSpin;
```
ILyoan: Why do we have to this delayed callback here? even this function is dummy so didn't actually set pin configuration.

ILyoan: Nowhere deleting DelayedCallSetSpin?

src/device/x86-linux/gpioctl_x86_linux.cpp
```diff
@@ -13,25 +13,128 @@
...
+  delay->type = DELAYEDCALL_SETSPIN;
+  delay->data = setpin_data;
+  delay->cb = cb;
```
ILyoan: Here nothing hold callback function object so this function could be turned garbage?

src/device/x86-linux/gpioctl_x86_linux.cpp
seanshpark: yes, thank you for pointing out.

src/js/gpio.js
seanshpark: creating "GpioError" in JS would be easy. so in native code it just retursn error code in case of error and don't call `callback`.

src/js/gpio.js
seanshpark: same here. native code does not call `callback` in case of error, just return err code.

src/js/gpio.js
seanshpark: same here.

src/device/x86-linux/gpioctl_x86_linux.cpp
seanshpark: * `callback` is needed to be called for dummy also but does not do any settings. delayed callback to emulate next tick.
* not deleted? need to fix this.

src/js/gpio.js
ILyoan: What if the error occur during async process? this implementation only cover sync mode.

We can create `GpioError` object like this:
```
function cb(errcode) {
  var err = null;
  if (errcode != 0) {
    err = new GpioError(errcode, gpioctl.errMessage(errcode));
  }
  callback(err);
}
gpioctl.setPort(portNumber, direction, mode, cb);
```

src/device/x86-linux/gpioctl_x86_linux.cpp
seanshpark: `cb` is `AfterSetPin` so it's not and object. callback function object is in `SetPinWrap* wrap` object of `JHANDLER_FUNCTION SetSpin` handler.

ILyoan: alright. that's good.
If it is always AfterSetPin, could we rename the variable or comment it. so that it would be easier to see?

seanshpark:
please review, more than a week has passed,

Issue #133, opened 2015-07-24
Printing empty string("") via console result segmentation fault.


ILyoan:
```
console.log("");
```

Related https://github.com/Samsung/jerryscript/issues/439

ILyoan:
Closing as #134 landed.

PR #134, opened 2015-07-24
Add test for issue #133

ILyoan:
Add test reproducing issue #133.

This test marked as "SKIP" since this test fail right now. should be fixed.


seanshpark:
LGTM

Issue #137, opened 2015-07-26
console.log fails for anything other than a non-empty string


roryrjb:
Aside from https://github.com/Samsung/iotjs/issues/133, `console.log` fails for anything other than a string; see script below, I've included the expected output as displayed by __Node v0.10.40__, __Node v0.12.7__ and __io.js v2.4.0__, and the latest build of __IoT.js__ from the `master` branch.

```javascript
// statement                    node/iojs         iotjs

console.log('hello world')      // hello world    hello world
console.log(1)                  // 1              uncaughtException: undefined
console.log(9 + 10)             // 19             uncaughtException: undefined
console.log(null)               // null           uncaughtException: undefined
console.log(true)               // true           uncaughtException: undefined
console.log(false)              // false          uncaughtException: undefined
console.log(undefined)          // undefined      uncaughtException: undefined
console.log(0.09)               // 0.09           uncaughtException: undefined
console.log(NaN)                // NaN            uncaughtException: undefined
console.log(function () {})     // [Function]     uncaughtException: undefined
console.log({})                 // {}             uncaughtException: undefined
console.log([])                 // []             uncaughtException: undefined
```

ILyoan:
@roryrjb  Thanks for reporting it.
At this time `console.log()` only prints string and fails for any other arguments.
We are preparing equivalent functionality with node.js. but we have to wait implementation of `String.prototype.replace()` of JerryScripts. After that it'll print all this request.

roryrjb:
@ILyoan thanks for your response. Keep up the good work :+1:

ILyoan:
Close as #156 landed.
Printing result is different from node.js since it requires not a few implementation.
Feel free to reopen if needed.

PR #139, opened 2015-07-27
Add test for issue #137. console.log() for non-string value.

ILyoan:
Add test for issue #137.
Marked as "SKIP" since this test fails at this time, should be fixed.


seanshpark:
LGTM

PR #140, opened 2015-07-27
initial commit for http module

chunseoklee:
1. simple support for GET, POST, HEAD method
2. no support for UPGRADE, CONNECT method
3. no buffering in outgoing message

### DIFF
--- a/src/iotjs_module_httpparser.cpp
+++ b/src/iotjs_module_httpparser.cpp
@@ -1,0 +1,465 @@
+/* Copyright 2015 Samsung Electronics Co., Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+
+#include "iotjs_def.h"
+#include "iotjs_module_httpparser.h"
+#include "http_parser.h"
+#include "iotjs_objectwrap.h"
+#include "iotjs_module_buffer.h"
+#include <string.h>
+#include <stdlib.h>
+#include <stdio.h>
+
+
+namespace iotjs {
+
+
+#define JSETPROPERTY( container_, fname_, vname_ )                      \
+  do {                                                                  \
+    JObject jobj(vname_ );                                              \
+    container_.SetProperty(fname_, jobj);                               \
+  } while(0)
+
+
+// If # of header fields == HEADER_MAX, flush header to JS side.
+// This is weired : # of maximum headers in C equals to HEADER_MAX-1.
+// This is because , OnHeaders cb, we increase n_fields first,
+// and check whether field == HEADER_MAX.
+// ex) HEADER_MAX 2 means that we can keep at most 1 header field/value
+// during http parsing.
+// Increse this to minimize inter JS-C call
+#define HEADER_MAX 10
+
+
+
+class HTTPParserWrap : public JObjectWrap {
+public:
+  explicit HTTPParserWrap(JObject& parser_, http_parser_type type)
+    : JObjectWrap(parser_) {
+    Initialize(type);
+    parser.data = this;
+  }
+
+  ~HTTPParserWrap() {
+  }
+
+  void Initialize(http_parser_type type);
+
+  // http-parser callbacks
+  int OnMessageBegin() {
+    url.MakeEmpty();
+    status_msg.MakeEmpty();
+    return 0;
+  }
+
+  int OnUrl(const char* at, size_t length) {
+    url.Append(at, length);
+    return 0;
+  }
+
+  int OnStatus(const char* at, size_t length) {
+    status_msg.Append(at, length);
+    return 0;
+  }
+
+  int OnHeaderField(const char* at, size_t length) {
+
+    if (n_fields == n_values) {
+      n_fields++;
+      // values and fields are flushed to JS
+      // before corresponding OnHeaderValue is called.
+      if (n_fields == HEADER_MAX) {
+        Flush(); // to JS world
+        n_fields = 1;
+        n_values = 0;
+      }
+      fields[n_fields-1].MakeEmpty();
+    }
+    IOTJS_ASSERT(n_fields == n_values + 1);
+    fields[n_fields-1].Append(at, length);
+
+    return 0;
+  }
+
+  int OnHeaderValue(const char* at, size_t length) {
+    if (n_fields != n_values) {
+      n_values++;
+      values[n_values-1].MakeEmpty();
+    }
+
+    IOTJS_ASSERT(n_fields == n_values);
+
+    values[n_values-1].Append(at, length);
+
+    return 0;
+  }
+
+  JObject makeHeader(){
+    // FIXME: This must be impl. as JS Array
+    JObject jheader;
+    JSETPROPERTY(jheader, "length", (n_values)*2);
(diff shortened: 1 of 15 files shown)


src/js/http_server.js
```diff
@@ -0,0 +1,264 @@
...
+  201 : 'Created',
+  202 : 'Accepted',
+  203 : 'Non-Authoritative Information',
+  204 : 'No Content',
+  205 : 'Reset Content',
+  206 : 'Partial Content',
+  207 : 'Multi-Status',               // RFC 4918
+  300 : 'Multiple Choices',
+  301 : 'Moved Permanently',
+  302 : 'Moved Temporarily',
```
brendanashworth: These status codes are not correct, see [here](http://www.iana.org/assignments/http-status-codes/http-status-codes.xhtml).

Mischanix: The goal here is to mimic node.js's http module, where this dictionary is copied from:  https://github.com/joyent/node/blob/4d2a1a993e575c8ca201b961dc741d24ae3b7efb/lib/_http_server.js#L42

brendanashworth: Okay, I understand :) I just want to note that they'll be changed in iojs/node 3.x to conform to the IANA ones.

ILyoan: I think "302 Moved Temporarily" was HTTP/1.0.
For we are serving HTTP/1.1, I think following the specification will be better choice:
http://tools.ietf.org/html/rfc7231#page-49

src/iotjs_module_httpparser.cpp
```diff
@@ -0,0 +1,566 @@
...
+      char* heap_str = new char[len];
+      memcpy(heap_str, str, len);
+      str = heap_str;
+      saved_heap = true;
+    }
+  }
+
+  void UpdateStr(const char* newstr, int length) {
+    if (str == NULL) {
+      str = newstr;
```
ILyoan: Is this safe?
`newstr` could be local buffer?

chunseoklee: ```newstr``` is a part of local buffer being parsed and handled by http-parser.  It is safe since these strings are flushed to JS side before parsing is done. 
Anyway, I will replace StringHelper with String soon.

chunseoklee:
Updated as the follows
1. String instead of StringHelper, which is removed.
2. Add some comments

src/iotjs_module_httpparser.cpp
```diff
@@ -0,0 +1,483 @@
...
+    }
+    IOTJS_ASSERT(n_fields == n_values + 1);
+    fields[n_fields-1].Append(at, length);
+
+    return 0;
+  }
+
+  int OnHeaderValue(const char* at, size_t length) {
+    if (n_fields != n_values) {
+      n_values++;
```
ILyoan: Is there no chance that `n_fields` exceeds `HEADER_MAX` ?

src/iotjs_module_httpparser.cpp
```diff
@@ -0,0 +1,483 @@
...
+    JObject func = jobj.GetProperty("OnHeadersComplete");
+    IOTJS_ASSERT(func.IsFunction());
+
+    // URL
+    JArgList argv(1);
+    JObject info;
+
+    // HEADERS
+    if (flushed) {
+      Flush();
```
ILyoan: Why flush headers only if some header information had already been flushed.

src/iotjs_module_httpparser.cpp
```diff
@@ -0,0 +1,483 @@
...
+    }
+    else {
+      JSETPROPERTY(info, "headers", makeHeader());
+      if ( parser.type == HTTP_REQUEST) {
+        IOTJS_ASSERT(!url.IsEmpty());
+        JSETPROPERTY(info, "url", url.data());
```
ILyoan: Looks like `url` is only set when nothing has been flushed yet. It is right ?

src/iotjs_module_httpparser.cpp
```diff
@@ -0,0 +1,483 @@
...
+  parser->cur_jbuf = NULL;
+  parser->cur_buf = NULL;
+  parser->cur_buf_len = 0;
+
+
+  if (parser->had_exception) {
+    return false;
+  }
+
+  if (nparsed != buf_len) {
```
ILyoan: Isn't this an exception case?

src/iotjs_module_httpparser.cpp
```diff
@@ -0,0 +1,483 @@
...
+  int nparsed = http_parser_execute(&(parser->parser), &settings,
+                                    buf_data, buf_len);
+
+  parser->cur_jbuf = NULL;
+  parser->cur_buf = NULL;
+  parser->cur_buf_len = 0;
+
+
+  if (parser->had_exception) {
+    return false;
```
ILyoan: I think we need to throw some Error object.

src/iotjs_module_httpparser.cpp
```diff
@@ -0,0 +1,483 @@
...
+
+JHANDLER_FUNCTION(Close, handler) {
+  IOTJS_ASSERT(handler.GetArgLength() == 0);
+  IOTJS_ASSERT(handler.GetThis()->IsObject());
+
+  JObject* jparser = handler.GetThis();
+  HTTPParserWrap* parser =
+    reinterpret_cast<HTTPParserWrap*>(jparser->GetNative());
+
+  delete parser;  // FIXME: used in FreeParser.
```
ILyoan: I don't think deleting parser instance here is safe. This may lead double free.
Fortunately it looks like there is no place where `parser.close()` is called.

src/js/http_client.js
```diff
@@ -0,0 +1,204 @@
...
+    var res = req.res;
+    res.on('end', function() {
+      res.emit('close');
+    });
+    res.push(null);
+  }
+
+  if (parser) {
+    parser.finish();
+    delete parser;
```
ILyoan: I see 
```
delete parser
```
is the same as 
````
delete this.parser
```` 
and here `this` is `socket` so this line reads as
```
delete socket.parser
```

but `req` is also holding `parser` with `req.parser`, so parser will not turn into garbage from this line.

ILyoan: And there are similar code around this patch.

ILyoan: Sorry I was confused with `delete` operator. So my interpretation was wrong.
But, anyway, this line does not make `parser` garbage.

src/iotjs_module_httpparser.cpp
chunseoklee: ```values``` and ```fields``` are flushed to JS in ```OnHeaderField``` function just before corresponding ```OnHeaderValue``` is called. So, it is not.

src/iotjs_module_httpparser.cpp
chunseoklee: In previous ```Flush()``` call, ```url``` was already flushed to JS side. However, if not ```flushed```, all headers and ```url``` start to flush here.

src/iotjs_module_httpparser.cpp
chunseoklee: Yes. It is true, here.

src/iotjs_module_httpparser.cpp
chunseoklee: Yes. ```parser.close()``` is used nowhere now. I will remove this.

src/js/http_client.js
chunseoklee: I will remove all the references to ```parser``` to make it GCed.

src/iotjs_module_httpparser.cpp
chunseoklee: I will delete the ```had_exception``` variable from ```HTTPParserWrap```. 
```had_exception``` is on when JS function call has issues. But it can be handled by ```MakeCallback``` throwing error routine.

chunseoklee:
@ILyoan Updated as commented.

ILyoan:
Could not thoroughly review this PR, This patch might have some glitches but I feel this is enough to be merged and at least we can fix problems if there are any.

chunseoklee:
Then, I will push this PR now, and follow up.

PR #144, opened 2015-07-28
Emitting end and finish event of duplex stream more properly

ILyoan:
This commit fixes two problems

1. In some situation 'end' event was not emitted from the first place.
2. In some situation 'end` event was lost because the socket closed before triggering 'end' listener.


chunseoklee:
LGTM lets push

PR #146, opened 2015-07-29
Add build script option '--cmake-option' to set cmake parameter

ILyoan:
From this PR we can pass cmake option like this:
```
./tools/build.py --cmake-option="-DCMAKE_EXE_LINKER_FLAGS=-static"
```


chunseoklee:
lets push

chunseoklee:
Merged

PR #147, opened 2015-07-29
Initializing process.argv

ILyoan:
Previously, `process.argv` only provided information about source path. and it was not an `Array`.

This patch will make `process.argv` an Array instance and contain other arguments as well as application path(argv[1]).

This patch contains some minor refactoring, most significant one of them is changing prototype of iotjs::Start() function: `Start(char* src)` to `Start(int argc, char** argv)`. Hence programs that use IoT.js as library should be modified (ex. nuttx entry program)

### DIFF
--- a/include/iotjs.h
+++ b/include/iotjs.h
@@ -25,7 +25,7 @@
 
 namespace iotjs {
 
-IOTJS_EXTERN_C int Start(char* src);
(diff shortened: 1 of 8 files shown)


chunseoklee:
LGTM lets push

PR #151, opened 2015-07-30
Enable 'AllowHalfOpen' option for server socket

ILyoan:
Server socket will be able to take 'allowHalfOpen' option and apply it when connections are established.


chunseoklee:
LGTM

PR #156, opened 2015-07-31
Fix printing console. now console module prints non-string value.

### LINKED ISSUE #137: console.log fails for anything other than a non-empty string

ILyoan:
Print console non-string value.
Results are not exactly the same with node.js.
Need discuss if we need to make it equivalent with node.js.

Fix #137, #133


chunseoklee:
LGTM lets push

chunseoklee:
Merged

PR #158, opened 2015-08-03
Add String `Append()` method.

ILyoan:
* Remove unused `LocalScope` and `LinkedList`.
* Add String `Append()`, `MakeEmpty()` method.
* Add String default constructor.


chunseoklee:
LGTM lets push

PR #159, opened 2015-08-04
EventEmitter.once fixed

chunseoklee:
small patch for EventEmitter.once : 
In http module, EventEmitter.once emit TypeError occasionally. This patch works fine with #140. 
In simple setting, it is failed to be reproduced.  


ILyoan:
@chunseoklee LGTM.

PR #167, opened 2015-08-10
Support caching non-native modules

### LINKED ISSUE #162: Caching non-native modules
Non-native modules are not cached now.
It may cause severe problem when the same module is loaded multiple times.

wateret:
Support caching non-native modules and add a test case.
This resolves #162 .

### DIFF
--- a/src/js/module.js
+++ b/src/js/module.js
@@ -27,6 +27,7 @@ function Module(id, parent) {
 module.exports = Module;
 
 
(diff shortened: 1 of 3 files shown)


seanshpark:
LGTM

seanshpark:
Current implementation may consume memory. Please add an optimisation issue to solve this.

chunseoklee:
LGTM

Issue #170, opened 2015-08-12
uv__finish_close: Assertion `handle->flags & UV_CLOSING' failed


seanshpark:
Sometimes there is uv__finish_close() assert, with httpd timer test codes.
Simple test code is as
```
var count = 0;

var tt = setInterval(function() {
  console.log("hello " + count);
  count+=1;

  var tick = 0;

  var timer = setInterval(function() {
    tick++;
    if (tick > 3) {
      clearInterval(timer);
      tick = 0;
    }
    console.log("timer test : " + count + " : " + tick);
  }, 500);
}, 1000);
```

32bit seems to reproduce better, 64bit linux doesn't seem to reproduce assert.

seanshpark:
For HandleWrap destruction, handle should be valid while processing `closing`.
Problem occurred when TimerWrap object was allocated to same address, by new, to TimerWrap object that was just deleted but still in process of `uv_close()` in `uv__finish_close()`.
As it's a new object, initialised by `uv_timer_init()` so `UV_CLOSING` is not set anymore making assert.
Fix #171 is to postpone object destruction for HandleWrap after `uv_close()` is finished.

PR #171, opened 2015-08-12
Delay HandleWrap destruction after handle is closed

seanshpark:
- for issue #170


chunseoklee:
LGTM

PR #172, opened 2015-08-13
Integer type correction using JObject(int v)

seanshpark:

JObject ctor for int is defined as `JObject(int v)`. In some platforms using int32_t fails.


ILyoan:
@seanshpark LGTM

Issue #168, opened 2015-08-11
Test 'test_net2.js' fail


seanshpark:
Only can see with 32bit linux
```
[ run_pass ] test_net2.js - unexpected stdout
[WRN] Attempt to close uninitialized or already closed handle
Hello IoT.js
```

tested with 687ec6ba531893390142fc54ab24afe5a30312f2 but seems been a while.

`unexpected stdout` is cause of 
```
export IOTJS_DEBUG_LEVEL=3
```
for debugging but `[WRN] Attempt to close uninitialized or already closed handle` needs investigation.

PR #174, opened 2015-08-17
Prevent duplicate socket destroy

ILyoan:
Fix for #168


seanshpark:
LGTM, no more #168 warnings

PR #182, opened 2015-08-18
Fixed misreferencing for socket state

ILyoan:
Fixed miss referencing for socket state.

net module of IoT.js maintain its state through `socket._socketState` object.


chunseoklee:
LGTM, lets push.

Issue #184, opened 2015-08-19
Implement socket.setKeepAlive()


ILyoan:
For #31

ILyoan:
Closing as #187 landed

PR #187, opened 2015-08-19
Add a method 'socket.setKeepAlive()'

wateret:
For #184.

Tested by capturing packets.


wateret:
Change function 'Socket.prototype.setKeepAlive' to handle the argument **enable** better.

ILyoan:
No problem on nuttx ?

ILyoan:
@wateret LGTM.

Seems like we already have problem with building on nuttx, it is not directly liked to this PR.

PR #196, opened 2015-08-25
Refactoring Gpio

chunseoklee:
1. unified GpioCbWrap
2. unified GpioCbData
3. unified After callback instead of AfterXXX


ILyoan:
It looks like `GpioCbWrap` is only for GPIO module. Why we have separate 'iotjs_gpiocbwrap.h' file? Is there a problem on having `GpioCbWrap` in 'iotjs_module_gpioctl.h` ?

ILyoan:
If I'm rigiht, `GpioCbWrap` is aggregating `GpioCbData` as `GpioCbWrap` allocates an instance of `GpioCbData` on its constructor and release it on its  destructor. In this case, we can just let `GpioCbWrap` own  `GpioCbData` as member value not by pointer. Furthermore, we might even not need to have `GpioCbData` class by just make `GpioCbWrap` have properties of it.

ILyoan:
Also I think `GpioCbWrap` could be a concrete class of `ReqWrap`. and we can consider renaming it to `GpioReqWrap`.

chunseoklee:
1. iotjs_gpiocbwrap.h/cpp were removed. ```GpioCbWrap``` and ```GpioCbData``` moved to iotjs_module_gpioctl.h
2. ```GpioCbWrap``` has ```GpioCbData``` as member variable, not as pointer now. 

I think  ```GpioCbData``` should be remained since it should be passed to ```gpioctrl->XXX(data, After)``` function as an args.

ILyoan:
@chunseoklee Let's get this merged at this level to go forward.

chunseoklee:
Merged

PR #204, opened 2015-08-31
Remove memory leak

ILyoan:
Fixed four types of memory leak:

* Cleanup fs request when request wrap goes away to prevent memory leak
* Fixed memory leak occurred whild reading source file.
* Fixed memory leak, not deleting handle wrap if it had been closed.
* Fix memory leak by exiting program before releasing memory

This PR also contains Jerry Update.


chunseoklee:
LGTM. 
@ILyoan  test_httpserver part in 6ece3aa commit  related to memory leak ?

ILyoan:
@chunseoklee No it is for make sure that all packet transitions properly completed.

ILyoan:
Merged

PR #208, opened 2015-08-31
http.get added

chunseoklee:


wateret:
What happens when we call *http.get()* with argument *option.method = 'POST'*
The request will be sent as GET or POST?

wateret:
LGTM

chunseoklee:
@wateret In node, request is sent as 'POST' without checking. There might be some options : ex. throw exception if 'POST' method is passed on http.get call.

chunseoklee:
Merged

PR #210, opened 2015-08-31
Reimplement js2c script in python version

wateret:
Reimplement js2c script in python version.
New script mininizes .js files when release mode by removing comments and whitespaces

Code size has decreased from 85391 to 55435(bytes).


chunseoklee:
LGTM.

PR #215, opened 2015-09-03
Rework on GPIO

ILyoan:
Rework on GPIO based on previous implementation.

* Reimplement gpio.js to meet GPIO API.
* Reimplement gpio module due to changes on gpio.js.
* Rename gpio module 'gpioctl' to 'gpio'.
* Rename 'device' directory to 'platform'
* Added GPIO implementation for linux generic.


chunseoklee:
How about emitting 'write' event on write complete? 
IMHO, distinction among event owner might be an issue.

ILyoan:
@chunseoklee It requires only simple work to adding 'write' event. 

Do you think providing that feature would make it easier to write GPIO program?
And what about 'read' event?

chunseoklee:
Sure,  it would be nice since cb hell is pain.

ILyoan:
@chunseoklee Then we need to review on GPIO API.

writePin has below prototype:
```
writePin(pinNumber: Number, value: Boolean, callback: Function(Error))
```
The callback function does not takes `pinNumber` and `value` since they are obvious.

If we decide to provide 'write' event, it would be like:
```
event: write (pinNumber: Number, value: Boolean) 
```

Are you OK with it?

chunseoklee:
@ILyoan Yes. Agree with event wih pinnumber  to recognize who is in charge of an event.

ILyoan:
@chunseoklee Opened issue to discuss about API 
# 217

ILyoan:
@chunseoklee Applied emitting 'writePin' and 'readPin' event after `writePin()` and `readPin()` operation respectively. callback parameter for `readPin()` become optional.

chunseoklee:
LGTM

Issue #205, opened 2015-08-31
Implement dns.lookup()


wateret:
For #200

wateret:
Closing as #216 landed

PR #216, opened 2015-09-03
Initial commit for dns module

wateret:
Created dns module.

APIs added
* dns.lookup() - for #205

### DIFF
--- a/src/iotjs_module.cpp
+++ b/src/iotjs_module.cpp
@@ -26,6 +26,7 @@
 #include "iotjs_module_timer.h"
 #include "iotjs_module_gpioctl.h"
 #include "iotjs_module_httpparser.h"
+#include "iotjs_module_dns.h"
 
 
 namespace iotjs {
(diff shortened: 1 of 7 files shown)


chunseoklee:
@wateret It it difficult to add test file for this feature ?

wateret:
Yes it is difficult to add tests.
Even testing with 'localhost', ip addresses other than '127.0.0.1' could be the answers.
I also found **gethostip** command in Linux which prints an ip address for a hostname but Mac OS doesn't have it.

ILyoan:
@wateret Looks like js file is missing.

wateret:
Just added it.

wateret:
Added  a commit.
Now argument *host* in *socket.connect()* is resolved with *dns.lookup()*. It emits 'lookup' event as well.

@chunseoklee With this commit, I was able to make a test case for dns module which creates a server and connects to it with hostname 'localhost' instead of '127.0.0.1

wateret:
Updated the test file to test 'lookup' event as well.

ILyoan:
+1

chunseoklee:
LGTM

PR #220, opened 2015-09-08
Added a test for stablizing net module. is marked as skip.

ILyoan:
Added a test for net module that current IoT.js does not correctly execute.

Marked as skip test but should be addressed.


wateret:
LGTM

Issue #228, opened 2015-09-09
Remove `jholder()` from ObjectWrap


ILyoan:
ObjectWrap is commonly used data structure in IoT.js but the property called '_jholder' is only used by TCP builtin.

I think it should be removed and TCP builtin need to be implemented in other way.

ILyoan:
Closing as #234 landed.

PR #234, opened 2015-09-14
Removed `jholder` from `ObjectWrap`.

ILyoan:
Fix for #228 

This patch accomopany with huge change in net module because implementation of previous net module and tcp builtin was based on `jholder`.


wateret:
LGTM

Issue #233, opened 2015-09-14
Kill child unit test process when timeout


ILyoan:
Current unit test checker halts without killing test processes when a test exceeds time limit.

Consequently resources opened by child test process remains not closed affecting following test.

ILyoan:
Close as fixed by #246

PR #246, opened 2015-09-17
Fix unexpected termination when timeout in check_test.py

wateret:
I temporarily added a test case *test_timeout.js* that takes forever.
In other words, this testcase always gets timed out.

Currently *check_test.py* terminates immediately after a timeout exception.
But with this patch even if a timeout exception is raised, it just reports timeout error for the test and keeps doing other tests.

The output is like
```
. . .
[ run_pass ] test_net3.js
[ run_pass ] test_timers2.js
[ run_pass ] test_timeout.js - timed out after 10 seconds

[ run_pass ] issue-133.js
[ run_pass ] issue-137.js
[ run_pass ] issue-198.js
[ run_pass ] issue-223.js
[ run_pass ] module_cache.js
[ run_pass ] test_require.js

[ run_pass ] Test Failed
  * test_timeout.js

Failed run_checktest
. . .
```

% *test_timeout.js* is not included in the commit.

### DIFF
--- a/tools/check_test.py
+++ b/tools/check_test.py
@@ -152,6 +152,7 @@ def run_test(arg):
     exitcode = 0
     attr_mismatched = False
     error_msg = ''
+    timeout_expired = False
 
     # run the test.
     try:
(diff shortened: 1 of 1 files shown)


wateret:
Related to #233, but not sure this will completely resolves it.

ILyoan:
LGTM.

PR #254, opened 2015-09-24
Guarantee a javascirpt object is alive during corresponding handle is open

ILyoan:
Changes:

* HandleWrap increases ref count for Javascript object when intialize. prevent GC.
* HandleWrap decreases ref count for Javascript object when close handle. allow GC.
* Close all unclosed handle just before clean up program.
* Rework on timer builtin.

### DIFF
--- a/src/iotjs.cpp
+++ b/src/iotjs.cpp
@@ -18,6 +18,7 @@
 #include "iotjs.h"
 #include "iotjs_js.h"
 #include "iotjs_string_ext.h"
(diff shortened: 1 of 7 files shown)


chunseoklee:
LGTM.
Also, ``` this._handleTemp = this._handle; ```(Server.prototype.close) can be removed by this patch ?

ILyoan:
@chunseoklee Yes. this is removed now.
Rebased on top of master.

wateret:
LGTM

PR #255, opened 2015-09-24
Changes for using libtuv and latest nuttx

seanshpark:
This change contains to use libtuv
* using tuv is optional for the moment within build.py  and it's off for default. use '--tuv' to turn on.
* copy output librarys to nuttx/lib if build for nuttx


chunseoklee:
LGTM

PR #261, opened 2015-10-02
Update Ext Magic String data

seanshpark:
Without magic string register, new data from and measured with `test_httpserver.js` file
```
  Peak allocated = 163868 bytes
  Peak waste = 1916 bytes
```

With magic string
```
  Peak allocated = 154956 bytes
  Peak waste = 1856 bytes
```

Diff `8,912`


chunseoklee:
LGTM

PR #264, opened 2015-10-05
Fix a bug in net.js

wateret:
This bug is caused by refactoring net.js

`this` object in `close` function should be substituted by `socket`.


ILyoan:
@wateret LGTM

Issue #274, opened 2015-10-27
To enable LTO for iotjs.


huxinx:
The current IoT.js is build without LTO.
From my watching, after enable LTO,
we get a iotjs binary with shrunk binary size and boosting performance.

Here is SunSpider 1.0.2 data I collected on ubuntu 15.04 32-bit, ATOM system,
first column is LTO ON, the second is LTO OFF data.                                        

3d-cube.js.......................1.475........9.782
access-binary-trees.js.....1.167........4.998
access-fannkuch.js.........3.803........26.549
access-nbody.js.............uncaughtException..uncaughtException
bitops-3bit-bits-in-byte.js.1.494........9.407
bitops-bits-in-byte.js	........1.827........13.518
bitops-bitwise-and.js........1.503........11.637
controlflow-recursive.js....1.828.........6.407
date-format-xparb.js........2.177.........5.432
math-cordic.js..................1.902........13.313
math-partial-sums.js.........uncaughtException..uncaughtException
math-spectral-norm.js......1.351.........8.024
string-fasta.js....................6.824........15.988

binary size ,
1.0M    iotjs.on  // LTO ON
2.5M    iotjs.off  // LTO OFF

Since LTO introduces so many benefits,  shall it be enabled ?

egavrin:
LTO build is a very useful feature. So, I'd strongly vote for adding this by default.
Could you please create pull request enabling LTO?

bzsolt:
We investigated this issue nowadays, and the conclusion was the same as you wrote.
I agree with @egavrin, it would be good to enable it.
I've created a PR for this, please check it: #275

egavrin:
@bzsolt great!

seanshpark:
Have you tried building and testing for NuttX / STM32F4-discovery?

huxinx:
I do not have STM32F4-discovery at this moment. For my experience, on x86, with gcc 4.9, just enable jerry_cmake_opt.append('-DENABLE_LTO=ON' is fine. With gcc 4.8, need to send extra flag '-flto' , other introduces link error. Maybe you can try to use another version gcc.

huxinx:
https://github.com/Samsung/iotjs/compare/master...huxinx:lto

@bzsolt @seanshpark @egavrin 
Here is my solution, add a switch for lto in build.py
the default config is still LTO=OFF,
turn on lto with --lto

huxinx:
Since 275 has been merged, please close this one.

PR #275, opened 2015-10-27
Enable LTO.

bzsolt:

### DIFF
--- a/cmake/iotjs.cmake
+++ b/cmake/iotjs.cmake
@@ -17,6 +17,11 @@ cmake_minimum_required(VERSION 2.8)
 file(GLOB LIB_IOTJS_SRC ${SRC_ROOT}/*.cpp
                         ${SRC_ROOT}/platform/${PLATFORM_DESCRIPT}/*.cpp)
 
+# LTO
+if("${ENABLE_LTO}" STREQUAL "ON")
+  set(CFLAGS "${CFLAGS} -flto")
+endif()
+
 set(LIB_IOTJS_CFLAGS ${CFLAGS})
 set(LIB_IOTJS_INCDIR ${TARGET_INC}
                      ${INC_ROOT}
--- a/tools/build.py
+++ b/tools/build.py
@@ -486,8 +486,8 @@ def build_libjerry():
         # prepare cmake command line option.
         jerry_cmake_opt = [JERRY_ROOT]
 
-        # set lto off.
-        jerry_cmake_opt.append('-DENABLE_LTO=OFF')
+        # set lto on.
+        jerry_cmake_opt.append('-DENABLE_LTO=ON')
 
         # tool chain file.
         jerry_cmake_opt.append('-DCMAKE_TOOLCHAIN_FILE=' +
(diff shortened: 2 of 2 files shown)


egavrin:
Please add "Related issue: #274" for the commit message.

bzsolt:
I missed that, thanks.

egavrin:
Looks good to me.

PS. I suppose it's better to wait for approve from someone else too.

seanshpark:
Need to test for Nuttx / STM32F4-discovery.

chunseoklee:
During nuttx build, the following msg is shown:

LD: nuttx
/home/twoflower/usr/dev/harmony/nuttx/nuttx/lib/libjerrycore.a: error adding symbols: Bad value
make[1]: *** [nuttx] Error 1

It seems that LTO makes some jerry symbols not readable from nuttx linker. Do someone has any solution ?

huxinx:
I do not have STM32F4-discovery at this moment. For my experience, on x86, with gcc 4.9, just enable jerry_cmake_opt.append('-DENABLE_LTO=ON' is fine. With gcc 4.8, need to send extra flag '-flto' , other introduces link error. What is your gcc version?

huxinx:
Another update maybe necessary,
explicitly use gcc-ar and gcc-ranlib for LTO build when using GCC 4.9.

seanshpark:
@bzsolt , can you check with STM32F4-disco + NuttX ? or If you don't have any target boards, you can  check with compilation only and leave some comments about the results here.
And can you use an option in `build.py` so that we can turn the switch? First we can use `--lto` option to turn it on and when we think it's ok for default then switch it on. If some one want's to turn it off he or she  can use `--no-lto` option to disable it.

bzsolt:
Of course, I've checked it with a board, and I've got the same linking error.

From the GCC's site:
>Link-time optimization does not work well with generation of debugging information. 
Combining -flto with -g is currently experimental and expected to produce unexpected results.

I've try to figure out what is the cause, I've tried to disable the debug options (such as `-g`, `-gdwarf-4`), but it doesn't help, so I haven't found any solution for this error yet.

ILyoan:
@bzsolt Remaking build.py is now ongoing.
You may find the changes in #281.
After the PR landed, we can turn on LTO with `build.py --jerry-lto`.

bzsolt:
The linker still fails.
I've tried the LTO build with different gcc version.
In case of gcc-4.8, I've got the above mentioned error.
With gcc-4.9, the following:
```
LD: nuttx
arm-none-eabi-ld: warning: cannot find entry symbol __start; defaulting to 08000000
stm32_vectors.o: In function `exception_common':
.../nuttx/arch/arm/src/chip/stm32_vectors.S:363: undefined reference to `up_doirq'
stm32_vectors.o: In function `_vectors':
stm32_vectors.S:(.vectors+0x4): undefined reference to `__start'
```

I' ve wanted to keep these symbols in object files with ```extern "C"```, ```__attribute__((used)))```, but the compiler disregarded these attempts.

In further, I think the all-in-one build will be a good substitute for LTO on NuttX.
Please check the details [here](https://github.com/Samsung/jerryscript/issues/626#issuecomment-156475847).

PR #276, opened 2015-10-29
gpio for nuttx

chunseoklee:

### DIFF
--- a/src/platform/arm-nuttx/iotjs_module_gpio-arm-nuttx.cpp
+++ b/src/platform/arm-nuttx/iotjs_module_gpio-arm-nuttx.cpp
@@ -18,6 +18,38 @@
 
 #include "iotjs_module_gpio.h"
 
+#include <unistd.h>
+#include <fcntl.h>
+#include <nuttx/gpio.h>
+#include <sys/ioctl.h>
+
+
+#define GPIO_MAX_PINNO 10
+
+
+
+// NUTTX on STM32F4 Gpio map in IoT.js //
+
+uint32_t gpioPin[GPIO_MAX_PINNO+1] = {
+  0,
+  (GPIO_PIN8   | GPIO_PORTA),
+  (GPIO_PIN10  | GPIO_PORTA ),
+  (GPIO_PIN15  | GPIO_PORTA ),
+  (GPIO_PIN11  | GPIO_PORTD ),
+  (GPIO_PIN3   | GPIO_PORTA ),
+  (GPIO_PIN5   | GPIO_PORTA ),
+  (GPIO_PIN14  | GPIO_PORTB ),
+  (GPIO_PIN10  | GPIO_PORTB ),
+  (GPIO_PIN2   | GPIO_PORTC ),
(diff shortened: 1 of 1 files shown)


chunseoklee:
This patch does not work now since dns module is not supported by nuttx. I will reopen this patch after dns module is fixed.

seanshpark:
Using with latest nuttx and libtuv build is ok.
1) use latest nuttx as described in [Build-for-Nuttx-(new)](https://github.com/Samsung/iotjs/wiki/Build-for-Nuttx-(new)) page
2) use `libtuv` with `--tuv` option in `build.py`

seanshpark:
As there are long code definitions from `nuttx/arch/arm/src/stm32/stm32_gpio.h ` file, I was thinking about removing them and include that particular file.

```
#include "../arch/arm/src/stm32/stm32_gpio.h"
```
This didn't work cause `arch/arm/src/stm32/chip/stm32f40xxx_pinmap.h` was including `"stm32_gpio.h"` and there is no given include search path.

So thought why not give include paths in the make file.
```
diff --git a/cmake/config.cmake b/cmake/config.cmake
index b29e519..e06daf9 100644
--- a/cmake/config.cmake
+++ b/cmake/config.cmake
@@ -32,6 +32,13 @@ elseif(${CFG_SYS_NAME} STREQUAL "external")
   endif()
 endif()
 
+if ("${PLATFORM_DESCRIPT}" STREQUAL "arm-nuttx")
+  set(TARGET_INC ${TARGET_INC} "${NUTTX_HOME}")
+  if ("${TARGET_BOARD}" STREQUAL "stm32f4disco")
+    set(TARGET_INC ${TARGET_INC} "${NUTTX_HOME}/arch/arm/src/stm32")
+  endif()
+endif()
+
 set(CC ${CMAKE_C_COMPILER})
 set(CXX ${CMAKE_CXX_COMPILER})
 
```

and change the include line like this
```
#include "stm32_gpio.h"
```

But again, when other boards should be included we may have to add (some or lots of) `#if defined(...) #elif defined(...)`.

So what do you and maybe others think about this?

chunseoklee:
@seanshpark updated as commented

seanshpark:
@chunseoklee thank you. one more thing. it's about `gpioPin[5]`, I can't see index limit check and I think there are more pins than 4, so, have any plans about this?

chunseoklee:
@seanshpark no special reason for that. I added 6 gpios to PR. Now, 10 gpios are available for nuttx-STM32.

seanshpark:
OK, LGTM

ILyoan:
@chunseoklee It looks like, we need more descriptor to help distinguishing among various boards.
Since this patch seems to work only for STM32F4 board. so I think "iotjs_module_gpio-arm-nuttx-stm32f4dis.cpp" would be better name. 
As we are remaking build.py and it will have target-board descriptor, Let's take look at the naming policy after build system landed.

chunseoklee:
This patch suspended due to low priority

PR #289, opened 2015-11-06
fix bug in build.py: implement external_include_dir option

jiangzidong:
build.py forget to use external_include_dir.


ILyoan:
@jiangzidong Great!
Look good to me.

Issue #272, opened 2015-10-21
Implement GPIO by libmraa, which can directly operate the board pins in Edison Board and others


jiangzidong:
IoTjs is designed for iot device, so the GPIO is one of the main features. Now the linux gpio has a general impl, which directly operate the gpios in kernel. While the mapping between kernel's gpio and physical gpio in board is difficult and varies with different embedded board, so we propose to import mraa into IoTjs and let mraa handle the mapping work. (https://github.com/intel-iot-devkit/mraa)

The codes are in https://github.com/jiangzidong/iotjs/tree/mraa_enable.
**Should I just directly create a pull request, and let's discuss there?** or stay in this issue? I think PR is easy to review the code.

1, We add a file : iotjs_module_gpio-linux-mraa.inl.h, which is parallel with iotjs_module_gpio-linux-general.inl.h. **The GPIO API is not changed**. Users can choose suitable implementation by the building option. 
```
./tools/build.py --mraa
```
2, Gpio-linux-mraa implement export, unexport, setpin, readpin, writepin with mraa api. Then the users can operate board gpio pin directly. For example, to turn on the LED in Edison Arduino Board.
```javascript
var gpio = require("gpio");
gpio.initialize(init_done);
function init_done() {
  print("init done!");
  gpio.setPin(13, "out", set_done);// the led is in the board pin 13
}
function set_done() {
  print("set done!");
   gpio.writePin(13, 1);
}
```
if the users operate the kernel gpio to turn on the led, they may need to refer the handbook and handle more than 5 gpios. see example in http://www.emutexlabs.com/project/215-intel-edison-gpio-pin-multiplexing-guide.

3, The libmraa is a **dinamic** libraries and is included in the official Edison/Galileo image. We add mraa as a submodule, and only build it for link. So the final `iotjs` doesn't have any mraa codes. It can run in the linux env with libmraa, such as Edison board. Building iotjs for Edison is easy. i686 version can run directly in Edison. In the host PC(no matter 32bit or 64bit), `./tools/build.py --mraa --target-arch=i686`, then copy the binary `iotjs` into Edison board.

seanshpark:
@jiangzidong , Thank you. I think it's nice to use `mraa` for gpio control. Hope to see others opinion also.
I think it's ok to discuss in this issue as it is created and when ready, make a PR referencing this.

[...]

PR #295, opened 2015-11-09
implement GPIO with libmraa (issue #272)

jiangzidong:
1, implement the external-static-lib and **externel-shared-lib** in build.py.
the externel shared lib is needed, because link-flag will put the `.so` in the wrong link order. see #272 for details.

2, implement GPIO with libmraa.
the building commands canbe:
`./tools/build.py --target-arch=i686 --compile-flag="-DUSING_MRAA" --external-include-dir="/home/jzd/iotm" --external-shared-lib="/home/jzd/iotm/libmraa.so"`
2.1: the header file and shared lib I used the command are taken from yocto linux in Edison board. And I put them into /home/jzd/iotm
2.2: USING_MRAA is a macro, which will affect the GPIO implementation in `iotjs_module_gpio-xxx-linux.cpp`. If it is defined, the iotjs will use mraa to implement GPIO. 

Mraa handles the complex mapping between board IO pin and kernel IO pin. If user want to handle the GPIO pin 13 on one board, he/she just operate the pin 13 by iotjs GPIO api. It makes the development more convenient.

example:
```javascript
var gpio = require("gpio");
gpio.initialize(init_done);
function init_done() {
  print("init done!");
  gpio.setPin(13, "out", set_done);// the led is in the board pin 13
}
function set_done() {
  print("set done!");
   gpio.writePin(13, 1);
}
```

### DIFF
--- a/cmake/iotjstuv.cmake
+++ b/cmake/iotjstuv.cmake
@@ -20,6 +20,8 @@ file(GLOB LIB_IOTJS_SRC ${SRC_ROOT}/*.cpp
 string(REPLACE ";" " " IOTJS_CFLAGS_STR "${IOTJS_CFLAGS}")
 string(REPLACE ";" " " IOTJS_LINK_FLAGS_STR "${IOTJS_LINK_FLAGS}")
 separate_arguments(EXTERNAL_INCLUDE_DIR)
+separate_arguments(EXTERNAL_STATIC_LIB)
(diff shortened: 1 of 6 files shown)


ILyoan:
@jiangzidong  Look good to me.
Thanks for the PR.

I'll merge it.

ILyoan:
merged.

PR #302, opened 2015-11-19
Implement Buffer.byteLength

chunseoklee:

### DIFF
--- a/src/js/buffer.js
+++ b/src/js/buffer.js
@@ -48,9 +48,39 @@ function Buffer(subject) {
 
 
 // Buffer.byteLength(string)
-Buffer.byteLength = function(str) {
-  // FIXME: Returns actual byte length of string not counts of characters.
-  return str.length;
+Buffer.byteLength = function(str, enc) {
+  var ret;
+  str = str + '';
+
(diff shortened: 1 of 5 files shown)


src/js/buffer.js
```diff
@@ -48,9 +48,29 @@ function Buffer(subject) {
 
 
 // Buffer.byteLength(string)
-Buffer.byteLength = function(str) {
-  // FIXME: Returns actual byte length of string not counts of characters.
-  return str.length;
+Buffer.byteLength = function(str, enc) {
```
ILyoan: As we are currently supporting only 'utf8' for string encoding. that I think we don't need to get additional parameter for distinguishing string encoding. Furthermore other than 'utf8' encoding should cause raising of an exception like `Error: unsupported encoding`.

chunseoklee:
@ILyoan updated as commented. 
utf8 is on default unless specific encoding option is given.

chunseoklee:
will be revised after this issue [https://github.com/Samsung/jerryscript/issues/722] is clear.

chunseoklee:
This patch suspended due to low priority

PR #303, opened 2015-11-20
Fix afterConnect function of net module

wateret:
- Do nothing when socket is already destroyed
- Fix resetSocketTimeout function gets wrong socket state

IoT.js-DCO-1.0-Signed-off-by: Hanjoung Lee hanjoung.lee@samsung.com

### DIFF
--- a/test/run_pass/test_net10.js
+++ b/test/run_pass/test_net10.js
@@ -1,0 +1,45 @@
+/* Copyright 2015 Samsung Electronics Co., Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


test/run_pass/test_net10.js
```diff
@@ -0,0 +1,45 @@
...
+ */
+
+var net = require('net');
+var assert = require('assert');
+
+var timedout = false;
+var connected = false;
+
+// Try connect to host that is not exist
+var socket1 = net.createConnection(11111, '89.89.89.89');
```
ILyoan: We could not assume there isn't a host on that IP.
Using local lookback(127.0.0.1) or [TEST-NET](https://en.wikipedia.org/wiki/IPv4#Special-use_addresses)(192.0.2.0) will be better.

RFC 5737, section 4 says:
> Addresses within the TEST-NET-1, TEST-NET-2, and TEST-NET-3 blocks SHOULD NOT appear on the public Internet and are used without any coordination with IANA or an Internet registry [RFC2050].

wateret:
@ILyoan Updated as commented.

ILyoan:
@wateret LGTM

chunseoklee:
+1

PR #310, opened 2015-11-24
Do not turn on `JERRY_FLAG_SHOW_OPCODE`

ILyoan:
There is a bug on pretty printer of JerryScript https://github.com/Samsung/jerryscript/issues/680 that prevents printing opcodes for certain tests.

IoT.js turns on 'show opcode' flag when memstat is needed. but 'show opcode' is not directly linked with memstat and even it prevent us from checking memstat due to ICE. That I think we don't need to turn on 'show opcode' for 'memstat'.

If 'show opcode' is really needed for other purpose, add a build option for it other than `ENABLE_JERRY_MEM_STATS`.

### DIFF
--- a/src/iotjs.cpp
+++ b/src/iotjs.cpp
@@ -39,7 +39,7 @@ static bool InitJerry() {
 
 #ifdef ENABLE_JERRY_MEM_STATS
   jerry_flag |= JERRY_FLAG_MEM_STATS;
-  jerry_flag |= JERRY_FLAG_SHOW_OPCODES;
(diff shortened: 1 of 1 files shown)


ILyoan:
On second thought, using process arguments seems better than build flags.

wateret:
@ILyoan I agree with your second thought. `JERRY_FLAG_SHOW_OPCODES` does not have to be a build flag.

wateret:
@ILyoan I'd like to ask why the code is commented out rather than removed.

ILyoan:
Closing this PR.
We are going to have command line arguments for this flags. #312 
I'll open another PR with it.

Issue #312, opened 2015-11-26
Add command line arguments for turn on and off flags.


ILyoan:
Some flags currently managed by build time options are able to be set by command lime arguments.

For example, when --jerry-memstat option for build.py is given, it makes Jerry with memstat feature on. and IoT.js will automatically turn on the flags as below.. 
```
 #ifdef ENABLE_JERRY_MEM_STATS
   jerry_flag |= JERRY_FLAG_MEM_STATS;
   jerry_flag |= JERRY_FLAG_SHOW_OPCODES;
 #endif
```

But I think it better to decide it at the time executing IoT.js via command line whether tuning on or off such flags.

For example,
```
$ iotjs --mam-stat --show-opcodes app.js
```

yichoi:
Closing. I found that this command line options work fine.

PR #313, opened 2015-11-26
Added IoT.js command line option '--memstat' and '--show-opcodes'

ILyoan:
Added IoT.js command line option '--memstat' and '--show-opcodes'

Removed `ENABLE_JERRY_MEM_STATS` compilation flag.

Related issue: #312

### DIFF
--- a/src/iotjs_env.cpp
+++ b/src/iotjs_env.cpp
@@ -1,0 +1,100 @@
+/* Copyright 2015 Samsung Electronics Co., Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "iotjs_env.h"
+
+#include <string.h>
+
+
+namespace iotjs {
+
+
+/**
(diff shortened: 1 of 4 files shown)


src/iotjs_env.cpp
```diff
@@ -0,0 +1,85 @@
...
+      _config.show_opcode = true;
+    } else {
+      fprintf(stderr, "unknown command line argument %s\n", argv[i]);
+      return false;
+    }
+    ++i;
+  }
+
+  // Remaining arguments are for application.
+  _argv = new char*[_argc + argc - i];
```
wateret: When does `_argv` get freed?

seanshpark:
Two options only work for debug build in JerryScript. In tools.py file,
```
    if option.jerry_memstat:
        option.buildtype = 'debug'
        option.no_check_test = True
```
is to make it debug build if enabled. For this change, is this still valid? 
Or I think it would nice for the user to see some message that it only works for debug build.

ILyoan:
@seanshpark  `JERRY_FLAG_MEM_STATS` works when jerry was built with 'mem_stats' feature. that is decided at build time. build.py will receive '--jerry-memstat' option to build jerry with the feature.
On the other hand new introducing command line arguments, '--memstat' and '--show-opcodes', are given when executing program.
We need to regard '-jerry-memstat' for build.py and '--memstat' for command line argument as independently.

If you give '--memstat' when running iotjs. but if jerry was not built with memstat feature. Jerry will print out following message:
```
Ignoring memory statistics option because of '!MEM_STATS' build configuration.
```

ILyoan:
@seanshpark And It looks like we don't need to impose jerry to be built for debug mode to enable memstat as we can see mem stats with even Jerry built for release mode. 
I'll remove that code changing build mode:
```
 if option.jerry_memstat:
        option.buildtype = 'debug'
```

ILyoan:
@wateret Updated PR.

wateret:
LGTM

PR #318, opened 2015-12-17
Separate the iotjs.js and other js modules in iotjs_js.h.

jiangzidong:
problem
---
`js2c.py` will turn all js module code and `iotjs.js` into const array in `iotjs_js.h`.

`iotjs.cpp` needs `iotjs.js`, `iotjs_module_process.cpp` needs other js modules, so they both include `iotjs_js.h`

Every const char array are stored in the .rodata section **twice**, which waste the storage resource, especially in embedded environment. e.g. for `System.map` in nuttx, 
```
jzd@jzd32:~/nuttx_ws/nuttx/nuttx$ cat iotjs/System.map | grep iotjs_s
0805e624 t _ZN5iotjsL7iotjs_sE
08083574 t _ZN5iotjsL7iotjs_sE
jzd@jzd32:~/nuttx_ws/nuttx/nuttx$ cat iotjs/System.map | grep gpio_s
080612ac t _ZN5iotjsL6gpio_sE
0807555c t _ZN5iotjsL6gpio_sE
```
solution
---
separate the iotjs_js.h into 2 files:
1. `iotjs_js.h` is only for `iotjs.js` 
2. `native_js.h` is for all other native js modules

So that no array will be multiple referenced.

result
---
in nuttx case, when the jerryheap is set to 64KB

||origin(kb)|after patch(kb)|decrease(%)|
---|---|---|---
whole flash size(nuttx.bin)|553|480|13.2|
iotjs own flash size (text + rodata + data)|425|352|17.1|

the iotjs own flash size is from `System.map`
**It can save more than 10% flash memory**

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@
 /build
 /build/*
 /src/iotjs_js.h
+/src/native_js.h
 /test/tmp/*
 
 # IDE related files
--- a/src/iotjs_module_process.cpp
+++ b/src/iotjs_module_process.cpp
@@ -15,7 +15,7 @@
 
 #include "iotjs_def.h"
 #include "iotjs_module_process.h"
-#include "iotjs_js.h"
+#include "native_js.h"
 
 #include <stdio.h>
 #include <stdlib.h>
--- a/tools/js2c.py
+++ b/tools/js2c.py
@@ -67,10 +67,16 @@ LICENSE = '''/* Copyright 2015 Samsung Electronics Co., Ltd.
  */
 '''
 
-HEADER = '''#ifndef IOTJS_JS_H
+HEADER1 = '''#ifndef IOTJS_JS_H
 #define IOTJS_JS_H
 namespace iotjs {
 '''
+
+HEADER2 = '''#ifndef NATIVE_JS_H
+#define NATIVE_JS_H
+namespace iotjs {
+'''
+
 FOOTER = '''}
 #endif
 '''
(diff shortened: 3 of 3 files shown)


ILyoan:
@jiangzidong Great!!
@chunseoklee What do you think about this?

chunseoklee:
After this patch, iotjs does not work on nuttx(STM32F4).

jiangzidong:
@chunseoklee **It can work on my nuttx** (STM32F407+BB, with the build guild https://github.com/Samsung/iotjs/wiki/Build-for-Nuttx-(new) ).

My build options are 
```
./tools/build.py --target-arch=arm --target-os=nuttx --nuttx-home=/home/jzd/nuttx_ws/nuttx/nuttx --target-board=stm32f4dis --jerry-heaplimit=64 --buildtype=release
```

chunseoklee:
In my STM board, 
``` console.log("console");  ``` fails.

**Comment by jiangzidong** [comment]
@chunseoklee but mine works well. Can the original iotjs work correctly on your board?

```
nsh> cat t.js
console.log("console")

nsh> iotjs t.js
console
```

chunseoklee:
@jiangzidong binary on 2ca7fe19e297 works fine on STM board. It seems awkward.  Need to figure out.

jiangzidong:
@chunseoklee What about we exchange our `nuttx.bin` via email, to check whether it is code, env or hardware's problem?

chunseoklee:
@jiangzidong bin you sent works well on nuttx. I will figure out the reason.

chunseoklee:
LGTM +1

wateret:
LGTM

@jiangzidong Nice catch! BTW, I had another thought creating `iotjs_js.cpp` and move initializations to that file. We may need to implement it after this PR has been merged as there is possibility that we could include `iotjs_js.h` or `native_js.h` in other sites.

jiangzidong:
@wateret Yes, I agree. I will work on the `iotjs_js.cpp` after this PR merges.

wateret:
Merged.

jiangzidong:
@wateret ,  PR #352  move the defination and initializations to the cpp file, while leave the declaration in the header

PR #325, opened 2016-01-04
js2c.py generate iotjs_js.h and iotjs_js.cpp

jiangzidong:
follow @wateret suggestion in #318 

js2c.py generate iotjs_js.h and iotjs_js.cpp, and put definations into the cpp file, put declaration into the header file.In this way, other cpp files can include the iotjs_js.h when they needs the module codes, and that actions will not waste flashsize.

In nuttx, the total size of nuttx.bin is same (481k) after this patch

### DIFF
--- a/src/iotjs_module_process.cpp
+++ b/src/iotjs_module_process.cpp
@@ -15,7 +15,7 @@
 
 #include "iotjs_def.h"
 #include "iotjs_module_process.h"
-#include "native_js.h"
+#include "iotjs_js.cpp"
 
 #include <stdio.h>
 #include <stdlib.h>
--- a/.gitignore
+++ b/.gitignore
@@ -3,7 +3,7 @@
 /build
 /build/*
 /src/iotjs_js.h
-/src/native_js.h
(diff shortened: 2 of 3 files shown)


src/iotjs_module_process.cpp
```diff
@@ -15,7 +15,7 @@
 
 #include "iotjs_def.h"
 #include "iotjs_module_process.h"
-#include "native_js.h"
+#include "iotjs_js.cpp"
```
wateret: I think this line should be 

```cpp
#include "iotjs_js.h"
```

wateret:
LGTM +1

LaszloLango:
LGTM

zherczeg:
LGTM

LaszloLango:
@jiangzidong, sorry for the long delay. Please add your [DCO](https://github.com/Samsung/iotjs/wiki/IoT.js-Developer's-Certificate-of-Origin-1.0) to the commit message and rebase to the current master, so we can land your PR.

PR #330, opened 2016-02-01
Fix string length computation in iotjs::WrapEval

akiss77:
The iotjs::String constructor automatically allocates at least
size+1 bytes for the data (that +1 byte for the terminating zero),
so there is no need for the +1 on the caller side.

Moreover, strcpy can work just as fine as strcat in this case,
since caller knows where the first string ends, so there is no need
to make strcat figure that out.


akiss77:
As motivated by @zherczeg in Samsung/jerryscript#813

LaszloLango:
LGTM

zherczeg:
LGTM

PR #332, opened 2016-02-02
Make license checking future proof

akiss77:
Instead of checking for a fixed year in the copyright message,
allow any year or interval.

IoT.js-DCO-1.0-Signed-off-by: Akos Kiss akiss@inf.u-szeged.hu

PR #333, opened 2016-02-02
Add mem_stats.sh tool

akosthekiss:
Adapted from jerry's mem_stats.sh. The following changes have been
made:
* IoT.js does not ask for "separate" mem stats from jerry and so
  statistics are only printed at the end of execution. Thus, there
  is only one "Peak Heap" column.
* IoT.js has a slightly different command line syntax to trigger
  mem stats (--memstat after the js file, not --mem-stats before
  it).
* Some IoT.js "pass tests" (which are used for memory benchmarking
  for the time being) require to be executed from their containing
  directory. Thus, the readlink command line utility is used to
  retrieve the canonical path of some files and then some directory
  changes happen.
* Because of the strict 80 characters-per-line limit, some lines
  are broken.

Note: Max RSS computation is done via the rss-measure.sh script of
jerry, which is not copied/adapted but used from within
deps/jerry/tools.)

### DIFF
--- a/tools/mem_stats.sh
+++ b/tools/mem_stats.sh
@@ -1,0 +1,124 @@
+#!/bin/bash
+
+# Copyright 2015-2016 Samsung Electronics Co., Ltd.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS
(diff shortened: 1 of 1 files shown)


akiss77:
Depends on #332

LaszloLango:
Could you add usage info at least in a comment?

LaszloLango:
@akiss77, this PR looks inactive for a long time. Do you have any plan to update? We have to close it if you don't.

akiss77:
Sorry for the delay. I've added usage/help, printed when no command line arguments are given. Then updated to master.

LaszloLango:
@akiss77, thanks for the update. LGTM

zherczeg:
LGTM

PR #334, opened 2016-02-02
Fix valgrind memcheck issues

akiss77:
I've run iotjs through valgrind with its own "pass tests". It has reported some memory issues which are fixed in this PR's patches.

Depends on #332


LaszloLango:
LGTM

zherczeg:
LGTM

PR #342, opened 2016-03-04
fixes an incompatibility with nodejs, when a module is named a direct…

toshas:
…ory, which has index.js, but does not have package.json with main field set


LaszloLango:
@toshas, sorry for the late response. Could you add a new test for this? (Maybe into the `test/run_pass/require1` folder?)

yichoi:
No updates of test case for a long time. If needed open it again

PR #344, opened 2016-03-23
Update JerryScript dependency

LaszloLango:

### DIFF
--- a/src/iotjs.cpp
+++ b/src/iotjs.cpp
@@ -55,13 +55,16 @@ static bool InitJerry(Environment* env) {
   InitJerryMagicStringEx();
 
   // Do parse and run to generate initial javascript environment.
-  if (!jerry_parse((jerry_api_char_t*)"", 0)) {
+  jerry_api_object_t *err_obj_p = NULL;
+  if (!jerry_parse((jerry_api_char_t*)"", 0, &err_obj_p)) {
     DLOG("jerry_parse() failed");
+    jerry_api_release_object (err_obj_p);
     return false;
   }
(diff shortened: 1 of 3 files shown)


src/iotjs_module_process.cpp
```diff
@@ -130,7 +130,7 @@ static JResult WrapEval(const char* source) {
   int len2 = strlen(source);
   int len3 = strlen(wrapper[1]);
 
-  String code("", len1 + len2 + len3 + 1);
+  String code("", len1 + len2 + len3);
```
akiss77: There is already a fix related to this issue in a pending PR (#330). As far as I can remember, it does not even need an updated jerry. I'd suggest (or ask, actually) to get that landed first.

LaszloLango: Sorry, I don't want to stole your credit. :) Of course I'll wait #330 to land first. This patch is currently a WIP patch, not ready for merge, but may help someone who want to use the latest version of JerryScript.

akiss77: @LaszloLango never thought so :) I pointed at that PR because it shows that the change can also be applied to current iot.js, and because that PR does a bit more than simply removing the +1 from the String constructor.

LaszloLango:
The PR is ready for review. @akiss77, I removed the changes of #330 from this PR. But that change is still needed for fresh Jerry. That should be merged first.

zherczeg:
LGTM

akiss77:
Pr #330 got merged. LGTM.

PR #346, opened 2016-03-25
Add Travis CI support to IoT.js

akiss77:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,18 @@
+language: cpp
+
+os: linux
+dist: trusty
+sudo: required
+
+before_install:
+  - deps/jerry/tools/apt-get-install-deps.sh
+
+install:
+
+script: "tools/build.py $OPTS"
+
+env:
+  - OPTS="--buildtype=debug"
+  - OPTS="--buildtype=release"
+  - OPTS="--jerry-lto"
+  - OPTS="--no-snapshot"


LaszloLango:
LGTM

akiss77:
Updated to current master. And travis is on already, so it already checked this :)

.travis.yml
```diff
@@ -0,0 +1,18 @@
...
+
+install:
+
+script: "tools/build.py $OPTS"
+
+env:
+  - OPTS="--buildtype=debug"
+  - OPTS="--buildtype=release"
+  - OPTS="--jerry-lto"
+  - OPTS="--no-snapshot"
```
zherczeg: No snapshot?

akiss77: Yes. That's a valid command line argument of `tools/build.py`. It disables the use of the snapshot feature. In this case, it runs a test job to check that variant as well. (Was that what you asked?)

akiss77: Yes. That's a valid command line argument of `tools/build.py`. It disables the use of the snapshot feature. In this case, it runs a test job to check that variant as well. (Was that what you asked?)

akiss77: Yes. That's a valid command line argument of `tools/build.py`. It disables the use of the snapshot feature. In this case, it runs a test job to check that variant as well. (Was that what you asked?)

zherczeg: So it runs  both with and without snapshot? That is ok.

akiss77: Yes.

zherczeg:
LGTM

PR #353, opened 2016-05-04
Compiling for Edison generated duplicated symbols

sergioamr:
Including the .cpp instead of the header generated duplicated symbols when compiling for Edison.
follow @wateret suggestion in #325

Also adding extra information about how to compile for Edison.

Added clarification about the folder to set mraa.h
The Intel license headers are not the same as the Samsung ones,
so the --no_check_tidy is required if added into the iotjs project folder.

Original Edison libmraa.so requires libft4222.so symbols to link in ubuntu.

Hope it doesn't break other builds, I've been trying to check against the other examples.

Signed-off-by: Martinez Rodriguez, Sergio <sergio.martinez.rodriguez@intel.com>

### DIFF
--- a/docs/Build-for-Edison-with-mraa.md
+++ b/docs/Build-for-Edison-with-mraa.md
@@ -29,6 +29,7 @@ In host PC
 ```
 cd mraa
 scp root@{edison ip}:/usr/lib/libmraa.so .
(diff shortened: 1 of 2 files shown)


LaszloLango:
LGTM

zherczeg:
LGTM

PR #354, opened 2016-05-12
Jerry update

LaszloLango:

### DIFF
--- a/cmake/config/arm-linux.cmake
+++ b/cmake/config/arm-linux.cmake
@@ -1,4 +1,4 @@
-# Copyright 2015 Samsung Electronics Co., Ltd.
+# Copyright 2015-2016 Samsung Electronics Co., Ltd.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -15,7 +15,7 @@
 include(CMakeForceCompiler)
 
 set(CMAKE_SYSTEM_NAME Linux)
-set(CMAKE_SYSTEM_PROCESSOR arm)
+set(CMAKE_SYSTEM_PROCESSOR armv7l-hf)
 
 set(EXTERNAL_CMAKE_C_COMPILER arm-linux-gnueabihf-gcc)
 set(EXTERNAL_CMAKE_CXX_COMPILER arm-linux-gnueabihf-g++)
--- a/src/iotjs_binding.cpp
+++ b/src/iotjs_binding.cpp
@@ -256,7 +256,6 @@ JObject JObject::GetProperty(const char* name) {
       _obj_val.u.v_object,
       reinterpret_cast<const jerry_api_char_t*>(name),
       &res);
-  IOTJS_ASSERT(is_ok);
   return JObject(&res);
 }
 
--- a/tools/build.py
+++ b/tools/build.py
@@ -1,6 +1,6 @@
 #!/usr/bin/env python
 
-# Copyright 2015 Samsung Electronics Co., Ltd.
+# Copyright 2015-2016 Samsung Electronics Co., Ltd.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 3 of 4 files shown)


seanshpark:
> 5,630 src/native_js.h

Is it OK to add this file?

LaszloLango:
@seanshpark, no. Accidentally added. I've updated the PR.

seanshpark:
can you check with arm-linux (raspberry pi 2) build?

LaszloLango:
It builds, but bails out when I try to run. I have to investigate what the problem is.

seanshpark:
ok, thank you.

akiss77:
I'm all in for the bump, with two requests:
- Could you bump jerry even further? I'd go with current jerry master.
- build.py needy a copyright year bump.

LaszloLango:
@akiss77, sure. The current master would be better, but at first we must find out why IoT.js fails on RPi2 after the update.

LaszloLango:
@akiss77, @seanshpark, I've updated the PR. Please check. It worked on RPi2.

I built with the following command:
`tools/build.py --clean --target-arch=arm --target-board=rpi2 --compile-flag="-static" --link-flag="-static"`

tools/build.py
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env python
 
-# Copyright 2015 Samsung Electronics Co., Ltd.
+# Copyright 2016 Samsung Electronics Co., Ltd.
```
akiss77: This should be 2015-2016

deps/jerry
```diff
@@ -1 +1 @@
-Subproject commit cc23c225ea7c03bfb0fd4f58352ab517ba26dc9e
+Subproject commit c7d33e98871187ad4777defc928233702c96b5bf
```
akiss77: There are newer commits in master again :)

src/iotjs_binding.cpp
```diff
@@ -256,7 +256,6 @@ JObject JObject::GetProperty(const char* name) {
       _obj_val.u.v_object,
       reinterpret_cast<const jerry_api_char_t*>(name),
       &res);
-  IOTJS_ASSERT(is_ok);
```
akiss77: I've run into this earlier but couldn't track down why this assert was here and why it's not needed anymore. Actually, it's a blocker now. Any idea why's that?

LaszloLango: This caused me a headache too. There where a bug in the Jerry API: https://github.com/Samsung/jerryscript/pull/1042

Currently the assert fails on `iotjs_module_tcp.cpp:86`

`JObject::GetProperty` returns with the result of the `jerry_api_get_object_field_value` call. Every GetProperty() checks the return value in runtime, so they are expecting that the property might not exist. There is no description for `JObject::GetProperty`. If we leave the assert there, then you can call the `GetProperty` with existing property names. IMHO, this is a too strict limit.

akiss77: I agree. Thanks for the details.

LaszloLango:
I've updated to the latest Jerry (9f29cc1) and tested on RPi2.

cmake/config/arm-linux.cmake
```diff
@@ -15,7 +15,7 @@
 include(CMakeForceCompiler)
 
 set(CMAKE_SYSTEM_NAME Linux)
-set(CMAKE_SYSTEM_PROCESSOR arm)
+set(CMAKE_SYSTEM_PROCESSOR armv7l-hf)
```
akiss77: Sorry for the late hit, I didn't realize before that this file also needs a copyright year update.

akiss77:
After the file header update, LGTM

LaszloLango:
@seanshpark, please review.

akiss77:
(Note: my suggestion is that after the second LGTM arrives, you do a bump of the jerry hash again. Not to get stuck in an infinite loop, you have my LGTM for that in advance, as long as only the hash changes.)

seanshpark:
LGTM, push

PR #356, opened 2016-05-18
Put dependencies(Linux) into a file and add an ARM-build to Travis CI

bzsolt:

### DIFF
--- a/tools/apt-get-install-deps.sh
+++ b/tools/apt-get-install-deps.sh
@@ -1,0 +1,19 @@
+#!/bin/bash
+
+# Copyright 2016 Samsung Electronics Co., Ltd.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+sudo apt-get update -q
+sudo apt-get install -q -y \
+    cmake g++
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,8 @@ dist: trusty
 sudo: required
 
 before_install:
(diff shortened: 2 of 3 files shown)


LaszloLango:
LGTM

tools/apt-get-install-deps.sh
```diff
@@ -0,0 +1,21 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+sudo apt-get update -q
+sudo apt-get install -q -y \
+    gyp cmake build-essential \
```
akiss77: `cmake` is already installed by `deps/jerry/tools/apt-get-install-deps.sh`
where are `gyp` and `build-essential` used?

bzsolt: I know that ```cmake``` is already installed, but I think it would be good to put every IoT.js dependencies into a single script (not only for testing). What do you think?
These are the required packages in the Wiki, I didn't check them. Should I?

akiss77:
`cmake` is sort of OK, iot.js itself uses it, not only its dependencies, so listing it makes the project more self-contained.
`build-essential` is a bit weird to me, I would expect that to be installed wherever development happens. Moreover, it's just a shorthand for "gcc, g++, libc-dev, make, dpkg-dev", out of which list we surely don't use the first and the last for iot.js itself (the code base being c++, and not doing any debian packaging), and even make is used transitively only by cmake (we have no handwritten Makefiles in iot.js).
`gyp` is a complete mistery to me. I haven't seen it used anywhere in the project.
So, for the above mentioned tools, I'd rather have an explicit list of what's really needed for local development.

As for the cross toolchain, I'm not compeletely sure that it wouldn't deserve its own script, but as even jerry mixes local and cross development dependencies in its install script, it's ok to have it here for now.

bzsolt:
I've created a virtual machine (Ubuntu 14.04.04) to check what you mention.
The `libc-dev` package were already installed. Only the `g++` package is the real dependency.
The `gyp` is optional indeed, the `http-parser` submodule can be build with it. But this currently uses `cmake`.

akiss77:
LGTM

LaszloLango:
still LGTM

PR #358, opened 2016-06-01
Don't rely on JS source being terminated with NL

akiss77:
Currently, it causes a "statement expected" error if last line of
JS code to be executed is a comment but not terminated with a NL.
This patch fixes that by always appending a NL character.

(Note: I wanted to add a regression test to `test/run_pass/issue` but couldn't as our tidyness checker does not let files be added to the code base with no NL at EOF...)


LaszloLango:
LGTM

seanshpark:
LGTM

PR #375, opened 2016-08-29
Jerry update

LaszloLango:

### DIFF
--- a/src/iotjs.cpp
+++ b/src/iotjs.cpp
@@ -20,7 +20,6 @@
 #include "iotjs_string_ext.h"
 #include "iotjs_handlewrap.h"
 
-#include "jerry.h"
 #include "jerry-api.h"
 #include "jerry-port.h"
 
@@ -39,37 +38,40 @@ Environment* Environment::_env = NULL;
  */
 static bool InitJerry(Environment* env) {
   // Set jerry run flags.
-  uint32_t jerry_flag = JERRY_FLAG_EMPTY;
+  uint32_t jerry_flag = JERRY_INIT_EMPTY;
 
   if (env->config()->memstat) {
-    jerry_flag |= JERRY_FLAG_MEM_STATS;
+    jerry_flag |= JERRY_INIT_MEM_STATS;
   }
 
   if (env->config()->show_opcode) {
-    jerry_flag |= JERRY_FLAG_SHOW_OPCODES;
+    jerry_flag |= JERRY_INIT_SHOW_OPCODES;
   }
 
   // Initialize jerry.
-  jerry_init((jerry_flag_t) jerry_flag);
+  jerry_init((jerry_init_flag_t) jerry_flag);
 
   // Set magic strings.
   InitJerryMagicStringEx();
 
   // Do parse and run to generate initial javascript environment.
-  jerry_api_object_t *err_obj_p = NULL;
-  if (!jerry_parse((jerry_api_char_t*)"", 0, &err_obj_p)) {
+  jerry_value_t parsed_code = jerry_parse((jerry_char_t*)"", 0, false);
+  if (jerry_value_has_error_flag(parsed_code)) {
     DLOG("jerry_parse() failed");
-    jerry_api_release_object (err_obj_p);
+    jerry_release_value(parsed_code);
     return false;
   }
 
-  jerry_api_value_t err_val;
-  if (jerry_run(&err_val) != JERRY_COMPLETION_CODE_OK) {
+  jerry_value_t ret_val = jerry_run(parsed_code);
+  if (jerry_value_has_error_flag(ret_val)) {
     DLOG("jerry_run() failed");
-    jerry_api_release_value (&err_val);
+    jerry_release_value(parsed_code);
+    jerry_release_value(ret_val);
     return false;
   }
 
+  jerry_release_value(parsed_code);
+  jerry_release_value(ret_val);
   return true;
 }
 
@@ -93,7 +95,7 @@ static void CleanupModules() {
 static bool RunIoTjs(JObject* process) {
   // Evaluating 'iotjs.js' returns a function.
 #ifndef ENABLE_SNAPSHOT
-  JResult jmain = JObject::Eval(String(iotjs_s, iotjs_l), false, false);
+  JResult jmain = JObject::Eval(String(iotjs_s, iotjs_l), false);
 #else
   JResult jmain = JObject::ExecSnapshot(iotjs_s, iotjs_l);
 #endif
--- a/src/iotjs_binding.cpp
+++ b/src/iotjs_binding.cpp
@@ -23,49 +23,35 @@ namespace iotjs {
 
 
 #define JVAL_IS_NULL(val_p) \
-    ((val_p)->type == JERRY_API_DATA_TYPE_NULL)
+    jerry_value_is_null(*val_p)
 
 #define JVAL_IS_UNDEFINED(val_p) \
-    ((val_p)->type == JERRY_API_DATA_TYPE_UNDEFINED)
+    jerry_value_is_undefined(*val_p)
 
 #define JVAL_IS_STRING(val_p) \
-    ((val_p)->type == JERRY_API_DATA_TYPE_STRING)
+    jerry_value_is_string(*val_p)
 
 #define JVAL_IS_OBJECT(val_p) \
-    ((val_p)->type == JERRY_API_DATA_TYPE_OBJECT)
+    jerry_value_is_object(*val_p)
 
 #define JVAL_IS_FUNCTION(val_p) \
-    (JVAL_IS_OBJECT(val_p) && jerry_api_is_function((val_p)->u.v_object))
+    jerry_value_is_function(*val_p)
 
 #define JVAL_IS_BOOLEAN(val_p) \
-    ((val_p)->type == JERRY_API_DATA_TYPE_BOOLEAN)
+    jerry_value_is_boolean(*val_p)
 
 #define JVAL_IS_NUMBER(val_p) \
-    (((val_p)->type == JERRY_API_DATA_TYPE_FLOAT32) || \
-     ((val_p)->type == JERRY_API_DATA_TYPE_FLOAT64) || \
-     ((val_p)->type == JERRY_API_DATA_TYPE_UINT32))
+    jerry_value_is_number(*val_p)
 
 #define JVAL_TO_BOOLEAN(val_p) \
-    (val_p)->u.v_bool
+    jerry_get_boolean_value(*val_p)
 
 #define JVAL_TO_NUMBER(val_p) \
-    ((val_p)->type == JERRY_API_DATA_TYPE_FLOAT32 ? \
-       static_cast<double>((val_p)->u.v_float32) : \
-     (val_p)->type == JERRY_API_DATA_TYPE_FLOAT64 ? \
-       static_cast<double>((val_p)->u.v_float64) : \
-     static_cast<double>((val_p)->u.v_uint32))
-
-
-
-JRawObjectType* GetGlobal() {
-  return jerry_api_get_global();
-}
-
+    jerry_get_number_value(*val_p)
 
 
 JObject::JObject() {
-  JRawObjectType* new_obj = jerry_api_create_object();
-  _obj_val = JVal::Object(new_obj);
+  _obj_val = jerry_create_object();
   _unref_at_close = true;
 }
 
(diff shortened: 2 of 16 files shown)


src/iotjs_binding.cpp
```diff
@@ -139,142 +115,126 @@ JObject::~JObject() {
...
                       bool direct_mode,
                       bool strict_mode) {
-  JRawValueType res;
-  jerry_completion_code_t ret = jerry_api_eval(
-      reinterpret_cast<const jerry_api_char_t*>(source.data()),
-      source.size(),
-      direct_mode,
-      strict_mode,
-      &res);
+  (void) direct_mode;
```
akosthekiss: Why do we need the `direct_mode` argument if it is unused?

src/iotjs_binding.cpp
```diff
@@ -139,142 +115,126 @@ JObject::~JObject() {
 }
 
 
-static JRawValueType jnull_val = JVal::Null();
-static JObject jnull(&jnull_val);
-JObject& JObject::Null() {
-  return jnull;
+JObject JObject::Null() {
+  return JObject(JVal::Null(), false);
```
akosthekiss: This creates new a null JObject over and over again. Previous implementation used a singleton pattern. It seems a reasonable approach for null. (Even though the implementation of the pattern might be improved / made nicer.) What is the reason behind changing it?

src/iotjs_binding.cpp
```diff
@@ -139,142 +115,126 @@ JObject::~JObject() {
...
 }
 
 
-static JRawValueType jundefined_val = JVal::Undefined();
-static JObject jundefined(&jundefined_val);
-JObject& JObject::Undefined() {
-  return jundefined;
+JObject JObject::Undefined() {
+  return JObject(JVal::Undefined(), false);
```
akosthekiss: Ditto for undefined.

tools/build.py
```diff
@@ -518,23 +527,14 @@ def build_libjerry(option):
...
-    target_libjerry_ms = {
-        'target_name': target_libjerry_ms_name,
-        'output_path': join_path([build_home, 'jerry-core',
-                                  'lib%s.a' % target_libjerry_ms_name]),
-        'dest_path': libjerry_output_path
-    }
-
     targets = []
     if option.jerry_memstat:
         targets.append(target_libjerry_ms)
```
akosthekiss: `target_libjerry_ms` has been removed, this `if` is superfluous

tools/build.py
```diff
@@ -443,17 +444,20 @@ def build_jerry(option):
     # Turn off LTO for jerry bin to save build time.
     cmake_opt.append('-DENABLE_LTO=OFF')
 
+    # Turn on snapshot
+    cmake_opt.append('-DFEATURE_SNAPSHOT_SAVE=ON')
+    cmake_opt.append('-DFEATURE_SNAPSHOT_EXEC=ON')
```
akosthekiss: Do we need snapshot execution? The jerry command line tool is only needed to _generate_ (i.e., save) snapshot, isn't it?

akosthekiss: Update: it seems that snapshot exec feature is missing from build_libjerry, however.

src/iotjs_binding.cpp
LaszloLango: jerry api is not available before the jerry_init() and static initializers run at startup, before the jerry_init(), so we cannot declare it as a static.

src/iotjs.cpp
```diff
@@ -39,37 +38,40 @@ Environment* Environment::_env = NULL;
...
   InitJerryMagicStringEx();
 
   // Do parse and run to generate initial javascript environment.
-  jerry_api_object_t *err_obj_p = NULL;
-  if (!jerry_parse((jerry_api_char_t*)"", 0, &err_obj_p)) {
+  jerry_value_t parsed_code = jerry_parse((jerry_char_t*)"", 0, false);
+  if (jerry_value_has_error_flag(parsed_code)) {
     DLOG("jerry_parse() failed");
-    jerry_api_release_object (err_obj_p);
+    jerry_release_value (parsed_code);
```
zherczeg: No need a space here.

src/iotjs.cpp
```diff
@@ -39,37 +38,40 @@ Environment* Environment::_env = NULL;
...
   }
 
-  jerry_api_value_t err_val;
-  if (jerry_run(&err_val) != JERRY_COMPLETION_CODE_OK) {
+  jerry_value_t ret_val = jerry_run(parsed_code);
+  if (jerry_value_has_error_flag(ret_val)) {
     DLOG("jerry_run() failed");
-    jerry_api_release_value (&err_val);
+    jerry_release_value (parsed_code);
+    jerry_release_value (ret_val);
```
zherczeg: It seems the style is mixing here :) I think we should follow IoT.js here.

src/iotjs_binding.cpp
```diff
@@ -85,49 +71,39 @@ JObject::JObject(bool v) {
 
 JObject::JObject(int v) {
   _obj_val = JVal::Number(v);
-  _unref_at_close = false;
+  _unref_at_close = true;
```
zherczeg: Is there a case where _unref_at_close is false? If not, just remove this variable.

src/iotjs_binding.cpp
zherczeg: But you can initialize at the first time and save it into soma global variable.

src/iotjs_binding.cpp
LaszloLango: Yes, for simple types (null, undefined, etc.)

src/iotjs_binding.cpp
```diff
@@ -139,142 +115,123 @@ JObject::~JObject() {
...
-  bool is_ok = jerry_api_get_object_field_value(
-      _obj_val.u.v_object,
-      reinterpret_cast<const jerry_api_char_t*>(name),
-      &res);
-  return JObject(&res);
+  JRawValueType prop_name = jerry_create_string(
+      reinterpret_cast<const jerry_char_t*>(name));
+  JRawValueType res = jerry_get_property(_obj_val, prop_name);
+  jerry_release_value(prop_name);
+  return JObject(res);
```
zherczeg: Can you create a JObject with a value with error flag? Even the original did not check is_ok. I think if there is an error we should change the ret value to undefined (by releasing the error).

src/iotjs_binding.cpp
```diff
@@ -344,19 +299,17 @@ JResult JObject::Call(JObject& this_, JArgList& arg) {
...
+  res = jerry_call_function(_obj_val, this_.raw_value(), val_args, val_argv);
 
   if (val_args) {
     delete [] val_args;
   }
 
-  JResultType type = is_ok ? JRESULT_OK : JRESULT_EXCEPTION;
+  JResultType type = jerry_value_has_error_flag(res)
+      ? JRESULT_EXCEPTION
+      : JRESULT_OK;
```
zherczeg: Wrong number of spaces.

src/iotjs_binding.cpp
```diff
@@ -344,19 +299,17 @@ JResult JObject::Call(JObject& this_, JArgList& arg) {
...
 
-  return JResult(&res, type);
+  return JResult(res, type);
```
zherczeg: I think the error flag must be removed here, because the type represents the error.

src/iotjs_binding.cpp
LaszloLango: AFAIK, this is the used style in IoT.js.

src/iotjs_binding.cpp
```diff
@@ -411,23 +362,17 @@ String JObject::GetString() {
...
-                                     * until jerry_cleanup is not called */
-                             &res);
-
-  IOTJS_ASSERT(ret == JERRY_COMPLETION_CODE_OK ||
-               ret == JERRY_COMPLETION_CODE_UNHANDLED_EXCEPTION);
+  JRawValueType res = jerry_exec_snapshot(
+      snapshot_p,
+      snapshot_size,
+      false); /* the snapshot buffer can be referenced
+               * until jerry_cleanup is not called */
```
zherczeg: Indentation again.

src/iotjs_module_fs.cpp
```diff
@@ -90,19 +91,20 @@ static void After(uv_fs_t* req) {
     fs_req->result = err; \
     After(fs_req); \
   } \
-  handler.Return(JObject::Null()); \
+  handler.Return(JVal::Null()); \
```
zherczeg: There is no need a \ at the end

src/iotjs_module_fs.cpp
```diff
@@ -90,19 +91,20 @@ static void After(uv_fs_t* req) {
...
   int err = uv_fs_ ## syscall(env->loop(), \
                               req_wrap.req(), \
                               __VA_ARGS__, \
                               NULL); \
   if (err < 0) { \
     JObject jerror(CreateUVException(err, #syscall)); \
     handler.Throw(jerror); \
-    return false; \
+    return; \
   } \
```
zherczeg: Ditto. I think we should remove these.

src/iotjs_module_tcp.cpp
```diff
@@ -282,7 +273,8 @@ void AfterWrite(uv_write_t* req, int status) {
   args.Add(JVal::Number(status));
 
   // Make callback.
-  MakeCallback(jcallback, JObject::Null(), args);
+  JObject undefined = JObject::Undefined();
```
zherczeg: I think if JObject::Undefined() would be static again, we could remove a lot of these local declarations.

src/iotjs_binding.cpp
zherczeg: I think in the new API we need to unref even the simple variables. I know that is a no-op at the memont, but still it can be changed in the future. So releasing everything is better.

src/iotjs_binding.cpp
zherczeg: There was some similar removed ? : code and they put ? and : in the same column.

src/iotjs_binding.cpp
akosthekiss: I think the following could work here:
```c++
JObject& JObject::Null() {
  static JObject null(JVal::Null(), false);
  return null;
}
```

This should do lazy initialization upon first call to the method (and return a reference, just like in previous version, which would enable a cleaner use of this method).

LaszloLango: @akosthekiss, thanks. Lazy initialization works fine for me. I'll update the patch.

src/iotjs_binding.cpp
LaszloLango: Yes, it is possible. I'll fix this.

src/iotjs_binding.cpp
LaszloLango: this is the IoT.js style

LaszloLango:
@akosthekiss, @zherczeg, thanks for the reviews. I've updated the PR.

src/iotjs.cpp
```diff
@@ -39,37 +38,40 @@ Environment* Environment::_env = NULL;
...
-  jerry_init((jerry_flag_t) jerry_flag);
+  jerry_init((jerry_init_flag_t) jerry_flag);
 
   // Set magic strings.
   InitJerryMagicStringEx();
 
   // Do parse and run to generate initial javascript environment.
-  jerry_api_object_t *err_obj_p = NULL;
-  if (!jerry_parse((jerry_api_char_t*)"", 0, &err_obj_p)) {
+  jerry_value_t parsed_code = jerry_parse((jerry_char_t*)"", 0, false);
```
zherczeg: Parse an empty string? What is the purpose?

LaszloLango: InitJerry will return with false if cannot parse and run it.

src/iotjs_binding.cpp
```diff
@@ -139,142 +115,132 @@ JObject::~JObject() {
...
+  return JObject((JRawValueType) jerry_get_global_object(), true);
 }
 
 
-JObject CreateError(const char* message, jerry_api_error_t error) {
-  return JObject(jerry_api_create_error(
-      error, reinterpret_cast<const jerry_api_char_t*>(message)));
+JObject CreateError(const char* message, jerry_error_t error) {
+  JRawValueType error_value = jerry_create_error(error,
+     reinterpret_cast<const jerry_char_t*>(message));
```
zherczeg: Isn't 4 space the indentation?

zherczeg:
LGTM

src/iotjs_binding.cpp
LaszloLango: yes, good catch

akosthekiss:
LGTM

Issue #335, opened 2016-02-03
Illegal memory read vulnerability in Buffer


akosthekiss:
The current implementation of buffers makes read access to data memory possible, which may raise security concerns.

```javascript
var buf = new Buffer(100);
var str = "x";

buf._builtin.write(str, 0, buf.length);

for (var i = 0; i < buf.length; i++) {
    console.log(i + ': "' + buf.toString(i, i + 1) + '"');
}
```

The above example accesses the `_builtin` member of `Buffer` objects and exploits the fact that the natively implemented `write` method (`JHANDLER_FUNCTION(Write)` in iotjs_module_buffer.cpp) performs very limited sanity checks on its arguments. Basically, all range checks are performed on js side (in `Buffer.prototype.write` in js/buffer.js), which the above example sidesteps. Thus, it will copy 100 bytes from the start of `str` into the buffer, from which it will be accessible afterwards. 100 is just an example, of course, we may read out and analyse "arbitrarily" large chunks of memory. This may expose sensitive data lingering in the memory.

E.g., the output on my box is:
```
0: "x"
1: ""
2: " "
3: "r"
4: "e"
5: "q"
6: "u"
7: "i"
8: "r"
9: "e"
10: ","
11: " "
12: "m"
13: "o"
14: "d"
15: "u"
16: "l"
17: "e"
18: ")"
19: " "
20: "{"
21: "
"
22: "v"
23: "a"
ICE: Assertion 'lit_is_utf8_string_valid (string_p, string_size)' failed at /home/akiss/devel/jerry/iotjs/deps/jerry/jerry-core/ecma/base/ecma-helpers-string.cpp(ecma_new_ecma_string_from_utf8):417.
Error: ERR_FAILED_INTERNAL_ASSERTION
```

That data from the 2nd index onward, I should not really have access to IMHO. (The ICE at the end seems to be unrelated -- but interesting.) Actually, valgrind shares my view and reports 98 invalid read errors for the above script.

PS: Generally, js side argument checking seems to be an issue in several functions in the buffer module, and in some other modules as well.

LaszloLango:
@akiss77, is this still valid?

akiss77:
@LaszloLango Confirmed. On latest master (0b7aa35385d4cbcaffa9a6a5338939becc663272), the output is still as follows:
```
0: "x"
1: ""
2: "u"
3: "n"
4: "c"
5: "t"
6: "i"
7: "o"
8: "n"
9: "("
10: "e"
[...]

PR #393, opened 2016-09-08
fix issue #335

yichoi:
In order to avoid unnecessary copy, adding length check code

### DIFF
--- a/src/iotjs_module_buffer.cpp
+++ b/src/iotjs_module_buffer.cpp
@@ -82,6 +82,17 @@ size_t BufferWrap::length() {
 }
 
 
+static size_t BoundRange(int index, size_t low, size_t upper) {
+  if (index < static_cast<int>(low)) {
+    return low;
+  }
+  if (index > static_cast<int>(upper)) {
+    return upper;
+  }
+  return index;
+}
+
+
 int BufferWrap::Compare(const BufferWrap& other) const {
   size_t i = 0;
   size_t j = 0;
@@ -193,6 +204,14 @@ JHANDLER_FUNCTION(Copy) {
   int src_start = handler.GetArg(2)->GetInt32();
   int src_end = handler.GetArg(3)->GetInt32();
 
+  dst_start = BoundRange(dst_start, 0, dst_buffer_wrap->length());
+  src_start = BoundRange(src_start, 0, src_buffer_wrap->length());
+  src_end = BoundRange(src_end, 0, src_buffer_wrap->length());
+
+  if (src_end < src_start) {
+    src_end = src_start;
+  }
+
   int copied = dst_buffer_wrap->Copy(src_buffer_wrap->buffer(),
                                      src_start,
                                      src_end,
@@ -216,6 +235,10 @@ JHANDLER_FUNCTION(Write) {
 
   BufferWrap* buffer_wrap = BufferWrap::FromJBufferBuiltin(*jbuiltin);
 
+  offset = BoundRange(offset, 0, buffer_wrap->length());
+  length = BoundRange(length, 0, buffer_wrap->length() - offset);
+  length = BoundRange(length, 0, src.size());
+
   size_t copied = buffer_wrap->Copy(src.data(), 0, length, offset);
 
   handler.Return(JVal::Number((int)copied));
@@ -232,8 +255,23 @@ JHANDLER_FUNCTION(Slice) {
 
   int start = handler.GetArg(0)->GetInt32();
   int end = handler.GetArg(1)->GetInt32();
+
+  if (start < 0) {
+    start += buffer_wrap->length();
+  }
+  start = BoundRange(start, 0, buffer_wrap->length());
+
+  if (end < 0) {
+    end += buffer_wrap->length();
+  }
+  end = BoundRange(end, 0, buffer_wrap->length());
+
+  if (end < start) {
+    end = start;
+  }
+
   int length = end - start;
-  JHANDLER_CHECK(length >= 0);
+  IOTJS_ASSERT(length >= 0);
 
   JObject jnew_buffer = CreateBuffer(length);
   BufferWrap* new_buffer_wrap = BufferWrap::FromJBuffer(jnew_buffer);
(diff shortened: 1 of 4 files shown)


LaszloLango:
Could you also add a regression test for the fix?

src/iotjs_module_buffer.cpp
```diff
@@ -216,7 +216,9 @@ JHANDLER_FUNCTION(Write) {
 
   BufferWrap* buffer_wrap = BufferWrap::FromJBufferBuiltin(*jbuiltin);
 
-  size_t copied = buffer_wrap->Copy(src.data(), 0, length, offset);
+  int length_min = src.size() < length ? src.size() : length;
```
LaszloLango: What is the behavior of Node.js in such cases? Wouldn't be better to throw an exception if length parameter is bigger than the length of the source?

yichoi:
Since it is possible for users to use Buffer.prototype._builtin, I think every boundary checks (which currently exist only in js part) should also exist in native part, too. To avoid duplication, I just moved the range check code into native part.
And after comparing the current code with Node.js implementation like @LaszloLango said, I found that current implementation doesn't throw proper exception on some invalid-ranged indexes, so I added some more RangeErrors. (But I think it is unnecessary to throw RangeError in native code, because IMHO it is implementation-specific decision.)

BTW I think maybe we should re-consider about Node.js compatibility.

test/run_pass/test_buffer_builtin.js
```diff
@@ -0,0 +1,107 @@
+/* Copyright 2015 Samsung Electronics Co., Ltd.
```
LaszloLango: Wrong copyright year

src/iotjs_module_buffer.cpp
```diff
@@ -82,6 +82,14 @@ size_t BufferWrap::length() {
 }
 
 
+static size_t boundRange(int index, size_t low, size_t upper)
+{
+  if (index < static_cast<int>(low)) return low;
+  if (index > static_cast<int>(upper)) return upper;
```
LaszloLango: I'm not sure what is the correct style, but this function looks odd to me. I think braces must be used in if statements and the left brace at the start of the function mustn't be on a separate line.

test/run_pass/test_buffer_builtin.js
```diff
@@ -0,0 +1,107 @@
...
+assert.equal(buff11.compare(buff10), 0);
+assert.equal(buff11.compare(buff12), -1);
+assert.equal(buff12.compare(buff11), 1);
+
+assert.equal(buff9._builtin.slice(-2, buff9.length).toString(), 'cd');
+assert.equal(buff9._builtin.slice(-3, -2).toString(), 'b');
+assert.equal(buff9._builtin.slice(0, -2).toString(), 'abcabcab');
+
+
+console.log(buff3.toString());
```
LaszloLango: leftover debug code?

yichoi: I copied that code from test_buffer.js, and I thought this code can be meaningful since it can check the result of toString (buf1) and slice (buf2).

LaszloLango: AFAIK, tools/build.py redirects the output of the tests and do not print anything into the console, so I think it is useless to write `console.log` in test files. If you'd like to validate `buff3.toString()`, then I suggest to change `console.log` to `assert.equal`.

src/iotjs_module_buffer.cpp
```diff
@@ -82,6 +82,17 @@ size_t BufferWrap::length() {
 }
 
 
+static size_t boundRange(int index, size_t low, size_t upper) {
```
LaszloLango: `boundRange` -> `BoundRange`

test/run_pass/test_buffer_builtin.js
yichoi: But https://github.com/Samsung/iotjs/blob/master/tools/check_test.py#L92 checks if @STDOUT exists in here: https://github.com/Samsung/iotjs/pull/393/files#diff-0f42bba455ef50958f24c64c145bfb91R17. It compares its console.log output, so if I remove the console.log, the test fails.

src/iotjs_module_buffer.cpp
```diff
@@ -216,6 +235,10 @@ JHANDLER_FUNCTION(Write) {
 
   BufferWrap* buffer_wrap = BufferWrap::FromJBufferBuiltin(*jbuiltin);
 
+  offset = boundRange(offset, 0, buffer_wrap->length());
```
LaszloLango: Isn't `BoundRange(offset, 0, buffer_wrap->length() - 1);` the correct call of this function?

test/run_pass/test_buffer_builtin.js
LaszloLango: I missed that part. OK then.

src/iotjs_module_buffer.cpp
yichoi: If `write` function is called from buffer.js, out-of-range offset will throw RangeError, so it's OK because this line cannot be called.

And if `write` function is called directly from _builtin.write, I think it is more reasonable not to write any bytes in the buffer, if input offset was out-of-range. 
If we use `buffer_wrap->length() - 1`, `buffer_wrap->Copy()` will write 1 byte into buffer even if the offset was invalid.

LaszloLango: The question was for every call of BoundRange. I'm worrying about the following corner case. If we have an example string "abcd" and call BoundRange like this

`BoundRande(4, 0, my_string->length());`

This will return 4, but my_String[4] is out of range. Did I missed something?

yichoi: Yes, my_String[4] is out of range. But if I understand your concern correctly, it is *intended* out-of-range.
For example, if we have code like below,
```
var buff = new Buffer("abcd");
buff._builtin.write("ef", 100, 2);
```
The offset `100` is invalid, and nothing should be written in `buff`.
It is accomplished by passing out-of-range offset `4` into below copy function. (we won't enter into the for loop - https://github.com/Samsung/iotjs/blob/master/src/iotjs_module_buffer.cpp#L111)

I think every use of `BoundRange` have this kind of boundary condition.

LaszloLango:
LGTM

src/iotjs_module_buffer.cpp
```diff
@@ -82,6 +82,17 @@ size_t BufferWrap::length() {
 }
 
 
+static size_t BoundRange(int index, size_t low, size_t upper) {
```
akosthekiss: The naming is a bit out-of-style: are we not starting function and method names with lower case letters, reserving upper case first letters for classes? (this would be `boundRange` then)

akosthekiss:
except for the style issue, the main content looks good

src/iotjs_module_buffer.cpp
LaszloLango: No. Grep for 'static void'  and you can see this is the correct style :) Did you find any example of the lowercase style? Is it mixed in the actual code base?

akosthekiss: You are right. I must have mixed up something. My bad, sorry.

akosthekiss:
LGTM

Issue #323, opened 2015-12-22
Segmentation Fault#When call fs.statSync or fs.stat as returning empty string which is path.


hyunjunekim:
If call fs.statSync or fs.stat as empty string which is path, raise `Segmentation fault`.
Example,
> var fs = require('fs');
> var stats1 = fs.statSync('');

Because on the binding function, get the null value through path String(iotjs_module_fs.cpp:293line).
> String path = handler.GetArg(0)->GetString();

If should call uv_fs_stat using null value which is path argument, occur to `Segmentation fault`.

glistening:
Thank you for reporting bug. I see same problem in current source. After investigating the source, I found there are same problems in fs.open and so on, too. First, I will make testcases issue-323.js using your report and put the issue-323 into our repository.

yichoi:
#396 landed. closing

PR #396, opened 2016-09-19
Fix Issue #323 (empty string as a file path)

sae-bom:
empty filename as an argument to 'fs' module should not make any segmentation faults. -Fix Issue #323


zherczeg:
LGTM

PR #401, opened 2016-09-20
Update submodule 'libtuv' to the latest

glistening:
Update submodule 'libtuv' to the latest version.
(It contains change for null check before strdup)


yichoi:
Merging this PR since Samsung/libtuv#31 landed

Issue #404, opened 2016-09-20
UnicodeDecodeError on running check_test.py


glistening:
When running tests with tools/check_test.py, in some systems it shows folowing error:

> [ run_pass ] test_http_get.js
multiprocessing.pool.RemoteTraceback:
"""
Traceback (most recent call last):
  File "/usr/lib/python3.4/multiprocessing/pool.py", line 119, in worker
    result = (True, func(*args, **kwds))
  File "/usr/lib/python3.4/multiprocessing/pool.py", line 44, in mapstar
    return list(map(*args))
  File "/home/yichoi/Working/iotjs/tools/check_test.py", line 139, in run_test
    test_attr = get_test_attribute(test)
  File "/home/yichoi/Working/iotjs/tools/check_test.py", line 97, in get_test_attribute
    join_path([RESOURCE_DIR, filename]))
  File "/home/yichoi/Working/iotjs/tools/check_test.py", line 76, in read_file_contents
    return f.read().encode().decode().strip()
  File "/usr/lib/python3.4/encodings/ascii.py", line 26, in decode
    return codecs.ascii_decode(input, self.errors)[0]
UnicodeDecodeError: 'ascii' codec can't decode byte 0xe2 in position 51: ordinal not in range(128)

In my case, I see this error on RPI2/raspbian. But not all RPI2/raspbian shows this error. Other RPI2/raspbian does not show this error. Other member found it on Ubuntu/x86_64.

it seems to depend on encoding setting in system's environment variable.
It would be better to modify check_test.py so that it runs well in any environment.

yichoi:
#405 landed. closing

PR #405, opened 2016-09-20
Fix UnicodeDecodeError on running check_test.py (#404)

glistening:
Set encoding type utf-8 explicitly to fix UnicodeDecodeError


yichoi:
LGTM

PR #406, opened 2016-09-21
Check DCO with Travis CI

yichoi:
Travis should check DCO on every commits.


yichoi:
Since we are already using "Squash and merge" button, (and it seems there's already a contributor who has different setting in github profile than in local git), I've adapted recent script in JerryScript.

akosthekiss:
LGTM

Issue #398, opened 2016-09-19
Build break for nuttx(with updated jerryscript)


haesik:
When I build iotjs for nuttx with updated source codes, I met a build break. 
It seems that the changed jerryscript does not support "arm" keyword of CMAKE_SYSTEM_PROCESSOR. 

-- The ASM compiler identification is GNU
-- Found assembler: /usr/bin/arm-none-eabi-gcc
System is unknown to cmake, create:
Platform/EXTERNAL to use this system, please send your config file to cmake@www.cmake.org so it can be added to cmake
-- CMAKE_SYSTEM_NAME         EXTERNAL
-- CMAKE_SYSTEM_PROCESSOR    arm
-- CMAKE_BUILD_TYPE          Release
-- JERRY_LIBC                ON
-- JERRY_LIBM                ON
-- JERRY_CMDLINE             ON
-- UNITTESTS                 OFF
-- PORT_DIR                  /home/haesik/harmony/iotjs/deps/jerry/targets/default
-- COMPILER_DEFAULT_LIBC     OFF
-- ENABLE_LTO                OFF
-- ENABLE_ALL_IN_ONE         ON
-- ENABLE_STRIP              ON
CMake Error at jerry-libc/CMakeLists.txt:30 (message):
  Machine architecture arm isn't supported

-- Configuring incomplete, errors occurred!
See also "/home/haesik/harmony/iotjs/build/arm-nuttx/release/deps/jerry/CMakeFiles/CMakeOutput.log".

[...]

PR #412, opened 2016-09-26
Fix Issue #398 (Build break for nuttx with updated jerryscript)

chokobole:
Fix Issue #398 

1. To fix error,
``` bash
redeclaration of C++ built-in type 'wchar_t'
```
I defined preprocessor variable CONFIG_WCHAR_BUILTIN when building IoT.js.
2. To synchronize build script in jerryscript side,  change word arm to armv7l.
3. To workaround 'cxa_guard_{acquire|release}', write JObject::init(), JObject::cleanup(), which is single point of entry and exit, in order to initialize and release undefined and null object.


src/iotjs_binding.cpp
```diff
@@ -116,13 +119,17 @@ JObject::~JObject() {
 
 
 JObject& JObject::Null() {
-  static JObject null(JVal::Null(), false);
-  return null;
+  if (null == NULL) {
+    null = new JObject(JVal::Null(), false);
```
akosthekiss: I'm a bit concerned about this approach. If I'm right, "undefined reference to `__cxa_guard_{acquire,release}'" is a compiler/linker/build system problem in the first place. These are C++ ABI functions to guarantee thread-safe initialization of local statics. If their implementation is missing then something is wrong there.

I see two cases:

1) If thread safety is not an issue, i.e., `JObject::Null` cannot be invoked from multiple threads in parallel, then we can use the "-fno-threadsafe-statics" compiler option and then the guard calls will not be generated. (This is still a not-that-nice workaround, but it should work.)

2) However, if multiple threads can invoke this method, then we have just sidestepped thread synchronization with the proposed change, and we may end up allocating multiple orphaned objects on the heap. And if we don't want that, then we will have to add our own locking around the if/new code path. (BTW, the here-allocated objects never get freed.)

chokobole: @akosthekiss I see.. Actually I found this when I tried to build nuttx, and then you occured to me there's possibility of linking problem with nuttx. I will check if there's other way to overcome this trouble then.

chokobole: @akosthekiss  I don't know much about NuttX, but hopefully I found [this commit](https://github.com/bherrera/NuttX/commit/15c1ace16cc4f7045ea28623936db23929631fea), after rebasing with his branch, i made it to build nuttx. I will write guild for this later on.

chokobole: @akosthekiss But lastly, I have just one question to ask you. As far as I know, JS is single thread language, and doens't it mean that this global Null and Undefined can only be accessed one by one??

akosthekiss: @chokobole I've taken a look at that commit. It's quite simple. It ensures static initialization in a single threaded context, but that's it, no thread safety. Didi it get accepted into Nuttx mainline?

As for threading in JS. Yes, JS is single threaded by default. (Although there are different approaches in different systems to allow some parallelization.) But... IoT.js is relying on libtuv, which _does_ use threading -- at least for the unix port, it surely uses pthreads. I'm not familiar enough with Nuttx to tell how things work there, but IoT.js should be OS agnostic, so we should consider all supported OSes (IMHO).

But I just bumped into an alternative potential solution, that can work on any OS and does not need static initializer guards. Roughly: `JObject` could have static `init` and `cleanup` functions that could instantiate/destroy `Null` and `Undefined` (on heap via `new` and `delete`), and `StartIoTjs` in iotjs.cpp could call them. This way no static initializer is needed, allocation surely happens only once, jerry api calls happen only after the api is available, and heap is deallocated when not needed. (This approach _does_ have the drawback, of course, that `JObject` as a class is only fully functional once its `init` method is called -- but AFAIK, that's all IoT.js-internal matter.)

chokobole: @akosthekiss Okay, if it is single point of entry and exit, then i totally agree with you, and I think I did as you said.  Is it correct??

PR #414, opened 2016-09-29
Implement to copy all the libraries in need to nuttx lib.

chokobole:
Before implementation, the following changes have been made:
* Refactor python script.
* Change code style to PEP8
* Remove the limit of 700 words to check license.(For the
* 'filesystem.py', which referencing to webkit)
* Skip '__init__.py' at check-tidy step


zherczeg:
LGTM

Issue #92, opened 2015-06-18
Automate Jerry external magic string registration.


ILyoan:
From #90, IoT.js registers commonly used string literal as external magic string of jerry engine.
Unfortunately we need to periodically maintain list of magic string by hand.

I think we can automate this by running a script at build process like "iotjs_js.h" file gereration.

yichoi:
Closing since #416 merged

PR #416, opened 2016-09-30
External magic string list should be generated automatically

yichoi:
Related issue: Samsung/jerryscript#922 Samsung/iotjs#92


zherczeg:
We have plans to implement a similar feature in JerryScript (external string generation during snapshot creation). Is this urgent to go in, or can it wait?

yichoi:
I also think it would be much better if jerryscript had this feature internally.

But please consider below requirements:
- IoT.js has multiple js files to generate snapshot
- Each snapshot can be loaded & executed separately
- Each js files may have duplicated strings

And my experimental result: IoT.js binary size gets much bigger with external magic string enabled. I think it's due to duplication of strings: snapshots have string literals inside, and external magic string array also have almost the same set of string literals. I think it would be great if we can remove this duplication. (maybe only for XIP-enabled devices?)

zherczeg:
In theory I would concatenate all js file, and run the external string generator once. This would enable to remove duplications.

zherczeg:
There was a time, when snapshots contained external string references. However, if the external string table changes, the snapshot is loaded incorrectly, and there is no way to test it during loading time. So this feature has been removed.

yichoi:
@zherczeg, could you please share your schedule about when to implement this feature in jerryscript?

zherczeg:
It is among the next steps but not yet scheduled. If you wish we can land this and change things after that task is done.

yichoi:
Thank you. Let's land first then remove after your implementation end.

zherczeg:
LGTM

Issue #49, opened 2015-05-26
Determine module search path order


chunseoklee:
Nodejs search for a module in the order of 
(1) ./node_modules
(2) $NODE_PATH/node_modules
(3) $HOME/node_modules
(4) $HOME/node_libraries
(5) (node execpath)/node/lib

We need to determine the search order w.r.t. multiple plaforms since enviroment or cwd is not available in small devices.

yichoi:
Closing since #417 merged.

Maybe we should re-consider the path with `node` in it. (in another issue)

PR #417, opened 2016-09-30
Cleanup module search path order

yichoi:
Now we can handle the case when some of cwd, $HOME, or $NODE_PATH is not valid.

Related issue: #49


zherczeg:
LGTM

Issue #357, opened 2016-05-25
latest code with wrong default linking library


gridnow:
hi,
I use lastest code but with error while linking:
[100%] Linking C executable debug.linux
/usr/bin/ld: skipping incompatible /usr/lib/gcc/x86_64-redhat-linux/4.8.5/../../
../libpthread.a when searching for -lpthread
/usr/bin/ld: skipping incompatible /lib/libpthread.a when searching for -lpthrea
d
/usr/bin/ld: skipping incompatible /usr/lib/libpthread.a when searching for -lpt
hread
/usr/bin/ld: cannot find -lpthread
/usr/bin/ld: skipping incompatible /usr/lib/gcc/x86_64-redhat-linux/4.8.5/../../
../libc.a when searching for -lc
/usr/bin/ld: skipping incompatible /lib/libc.a when searching for -lc
/usr/bin/ld: skipping incompatible /usr/lib/libc.a when searching for -lc
/usr/bin/ld: cannot find -lc
collect2: error: ld returned 1 exit status

I am using x64, but default libc path is /use/lib?   Old code is ok! but i do not remember its version after git pull

LaszloLango:
Could you send us your linux version and architecture of your workstation. What parameters are you using for the build?

gridnow:
centos7, x64 version
python 2.6.6
python tools/build.py without any parameter

Default lib path seems changed to x32,, but i am using x64

LaszloLango:
I suggest to try a clean release build at first
`tools/build.py --clean --buildtype=release`

Maybe it helps if you set your architecture explicitly with `--target-arch=x86_64`

gridnow:
It's strnge, with --clean i can compile and get executable file iotjs, but again without --clean, above errors occured.

yichoi:
Closing since PR #423 merged.

sae-bom:
The reason for the problem was a conflict which was happened between Jerry's binary build option and library build option.
It has been fixed by PR #423  
But in the future when you want to build JerryScript with compiler-default-libc(not jerry-libc) and static linking option, I recommend you to install some packages which contain 'static' libraries like pthread.a. Centos may not have those libraries basically.

PR #423, opened 2016-10-05
Separate Jerry binary buildroot from Jerry lib buildroot

sae-bom:
Because it can cause a conflict on CMakeCache.txt when host and target machine is same and both of them use same directory as a buildroot.
It can clear the issue #357


yichoi:
@sae-bom please rebase it again

PR #430, opened 2016-10-10
Rename MakeCallback into InvokeCallback

yichoi:
`MakeCallback()` function actually calls the callback. Since its naming can be misleading (it seems like it creates the callback object), I propose to change its name into `InvokeCallback()`.


yichoi:
It seems that it's following the name of nodejs equivalence. I'm closing issue until we resolve nodejs compatability issue.

PR #434, opened 2016-10-12
Set read size to 1023 when read files for nuttx compatibility

chokobole:


src/js/fs.js
```diff
@@ -221,8 +221,8 @@ fs.readFile = function(path, callback) {
 
   var read = function() {
     // Read segment of data.
-    var buffer = new Buffer(1024);
-    fs.read(fd, buffer, 0, 1024, -1, afterRead);
+    var buffer = new Buffer(1023);
+    fs.read(fd, buffer, 0, 1023, -1, afterRead);
```
yichoi: I don't think this value is related with above definition

src/iotjs_def.h
```diff
@@ -20,7 +20,7 @@
 
 #ifndef IOTJS_MAX_READ_BUFFER_SIZE
  #ifdef __NUTTX__
-  #define IOTJS_MAX_READ_BUFFER_SIZE 1024
+  #define IOTJS_MAX_READ_BUFFER_SIZE 1023
```
yichoi: Please add reference why 1023 is valid

LaszloLango:
Is it possible to write a test for this issue?

chokobole:
@LaszloLango I raised an issue about this..

PR #435, opened 2016-10-12
Add targets for nuttx-stm32f4 to support nuttx main repository.

chokobole:


mulderp:
thanks for that, I'll check later today to run the build. one small observation the directory is called `targtes` --> should probably be "targets" ?

yichoi:
PR #442 raised to fix typo directory name. Thanks

Issue #399, opened 2016-09-20
Null String allocates 1 byte heap memory


glistening:
With recently merged commit 54cd255979f5f1fb3e6ee38e2b56b99c60b48955 for fixing #323, we allocate 1 byte memory with malloc for every null string. Previously, null string is represented by setting _data to NULL. I think memory defragmentation should be considered more carefully in iot environment since it has small memory. Although this patch make it easy to remove many bugs relating non-nullchecked access, I think previous approach, i.e. without allocation, is more suit for iotjs. If maintainers think the commit seems good, at least, making null string as singleton would be better.

glistening:
As for now, I checked isssue #323 passes without the patch 54cd255. I will make a PR. If there is other reason for the change, Let me know.

In addition, 

 93   IOTJS_ASSERT(_size >= 0);
 94   IOTJS_ASSERT(_cap >= 0);
 95 
 96   if (_cap >= 0) {
 97     _data = AllocBuffer(_cap + 1);
 98   } else {
 99     _data = NULL;
100   }

line 98, 99, 100 seems strange. because line 94 says _cap is non-negative.

yichoi:
Closing since #444 merged

PR #444, opened 2016-10-13
Don't allocate 1 byte heap memory for Null String

glistening:
With recently merged commit #54cd255 for fixing #323, we allocate 1
byte memory with malloc for every null string. Previously, null string
is represented by setting _data to NULL. I think memory defragmentation
should be considered more carefully in iot environment since it has
small memory. Although this patch make it easy to remove many bugs
relating non-nullchecked access, I think previous approach, i.e. without
allocation, is more suit for iotjs. If maintainers think the commit
seems good, at least, making null string as singleton would be better.

I checked tc issue-323.js pass with null check in libtuv.

### DIFF
--- a/src/iotjs_util.cpp
+++ b/src/iotjs_util.cpp
@@ -93,7 +93,7 @@ String::String(const char* data, int size, int cap) {
   IOTJS_ASSERT(_size >= 0);
   IOTJS_ASSERT(_cap >= 0);
 
-  if (_cap >= 0) {
+  if (_cap > 0) {
     _data = AllocBuffer(_cap + 1);
   } else {
     _data = NULL;


yichoi:
Related with #399

LaszloLango:
You just reverted 54cd255 but I don't see what happened with the Segmentation fault previously reported. What is the result of the `test/run_pass/issue/issue-323.js`?

glistening:
As I wrote in last line of commit description, " I checked tc issue-323.js pass."

The issue #323 was already fixed with the commit Samsung/libtuv@bfc5089a415769d60e843b4359847adeca97c79a in libtuv.
libtuv submodule reference is also already updated in b90ed4cef2bf2c366d0aad1a7ae3463d87c7f068.

This commit is for removing allocation 1 byte heap memory for null string, not for fixing #323.

LaszloLango:
@glistening, thanks for the explanation. LGTM

Issue #436, opened 2016-10-12
Make IoT.js modules build-time configurable


yichoi:
Currently, every builtin modules are included in iotjs final binary.
However, some of the modules are not essential for every usage of IoT.js.
We should make it possible to select which module to include in binary at build-time.

yichoi:
Closing since https://github.com/Samsung/iotjs/pull/458 merged

PR #447, opened 2016-10-13
Refactoring: Separate module codes into `module` directory

yichoi:
First step to https://github.com/Samsung/iotjs/issues/436 : Make core logic independent of each module code


src/iotjs_binding_helper.cpp
```diff
@@ -0,0 +1,99 @@
...
+
+  JObject jexit(process->GetProperty("emitExit"));
+  IOTJS_ASSERT(jexit.IsFunction());
+
+  JArgList args(1);
+  args.Add(JVal::Number(code));
+
+  JResult jres = jexit.Call(JObject::Undefined(), args);
+  if (!jres.IsOk()) {
+    exit(2);
```
LaszloLango: It would be better to use defines for exit codes.

akosthekiss: As this is more of a code reorg patch, it's OK to keep original approaches IMO. (That said, I'd second a change in the future.)

src/iotjs_binding_helper.cpp
```diff
@@ -0,0 +1,99 @@
+/* Copyright 2016 Samsung Electronics Co., Ltd.
```
akosthekiss: As this code is originating from existing sources, better keep 2015-2016 copyright years

src/iotjs_binding_helper.h
```diff
@@ -1,4 +1,4 @@
-/* Copyright 2015 Samsung Electronics Co., Ltd.
+/* Copyright 2016 Samsung Electronics Co., Ltd.
```
akosthekiss: Same here: copyright 2015-2016

src/module/iotjs_module_process.h
```diff
@@ -0,0 +1,30 @@
+/* Copyright 2015 Samsung Electronics Co., Ltd.
```
akosthekiss: 2015-2016

yichoi:
@akosthekiss I've modified notices as per your comment

yichoi:
@LaszloLango Thanks for your comment. I will make another PR separately to reflect your comment.

yichoi:
@akosthekiss 2nd try :)

PR #450, opened 2016-10-13
Implement test driver from python driver to js driver 

chokobole:
* Skip some of tests, because there's no way to check process related events with this method.
* Change the way from comparison between console output between data to assert

Issue : on nuttx side, because of heap limit, the test driver can go through all the tests. So I split tests to several groups.


chokobole:
For now, every testcase has some additional function which plays a role that driver knows if the test endes or not. But this modification makes the developer feel ambiguous and difficult to write test cases. Accepting colleague's opinions which developers can write transparent test cases, implement more and pr later.

chokobole:
Accepting colleauge's opinion, I did few changes to original test cases
* Skip the test cases which are related to process events
* Don't compare exit code in process.on('exit', function () {}), because I overwrote process.exit to catch exceptions inside callback functions
```javascript
fs.exists('a.js', function () {
  assert.fail();
 })
````
* Make mock console object, so that prevent printing out while test driver is running.

File: test/run_fail/attrs.js [hunk 1]
```diff
@@ -0,0 +1,37 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+(function() {
+  var attrs = {
+    'process_exit1.js': {
+      skip: ['all'],
```
yichoi: Are you planning to re-enable process-related events? Otherwise the test coverage will be narrowed down after this patch get merged.

test/run_pass/issue/issue-133.js
```diff
@@ -13,6 +13,14 @@
  * limitations under the License.
  */
 
+if (this.driver) {
```
yichoi: If this code is valid for every testfiles, how about replacing it before & after evaluating every file contents in test driver?
(Object.defineProperty() with writable=false might be helpful, if you're worried about

test/run_pass/issue/issue-133.js
```diff
@@ -13,6 +13,14 @@
...
+  var originalConsole = console;
+  console = require('tools/commonJS/modules/console_wrapper');
```
yichoi: Is this `commonJS` related to [this CommonJS](http://www.commonjs.org/)? If not, it would be better to use other name to avoid confusion.

tools/check_test.js
```diff
@@ -0,0 +1,170 @@
+/* Copyright 2016 Samsung Electronics Co., Ltd.
```
yichoi: If we have to copy test files into IoT devices, I think all of the test driver files should be placed inside `test` directory, not separated in `test` and `tools` each.

tools/test_runner.js
```diff
@@ -0,0 +1,83 @@
...
+    } else {
+      if (!that.finished) {
+        that.spin(fail);
+      }
+    }
+  });
+}
+
+Runner.prototype.run = function() {
+  if (this.attr.fail) {
```
yichoi: I think we can replace console around here if it's necessary.

tools/check_test.js
```diff
@@ -0,0 +1,170 @@
...
+  return content;
+};
+
+Driver.prototype.finish = function() {
+  this.logger.message('finish all tests', this.logger.summary);
+}
+
+var driver = new Driver();
+
+process.exit = function () {
```
yichoi: Do we have to override it?
Couldn't we emit the exit code(`process.emitExit`) and do other stuffs like original process.exit() does,
and then call `driver.runner.finish`?
i.e. overriding just process.doExit, not whole process.exit.

There are so many tests that checks process exit code.

yichoi:
+1 on changing driver from python into js, since many IoT devices lack of python interpreter. My comments are mainly focused on not-touching original test files, and writing tests in more straightforward way.

test/run_pass/issue/issue-133.js
chokobole: Absolutely not, I just meant this is common modules for javascript. just like common direcotry for python. How about renaming 'common_js' and 'common_py' then?

tools/check_test.js
chokobole: Regarding this opinion, I don't totally agree, ``test`` is the directory which are for testfiles literally, and ``tools`` is the directory which are tools for utility such as check license, or run tests. Can we just move ``test`` and ``tools`` both? I don't think that is a big burden?

tools/test_runner.js
chokobole: Okay, I agree, thanks for the review!

tools/check_test.js
chokobole: The reason I don't compare exit code is that if the test runs failed in the middle, the process will be exit only once, and the exit code related whole tests are failed in that case. 
For example, ``testA`` is run passed, and this compares eixt code with 0.
```javascript
//testA
process.on('exit', function (code) {
  assert.equal(0,code);
});
```
But if the following test case B runs failed, and it is terminates with ``assert.fail`` which emits exit code other than 0, then even though test A runs passed but because this exit code, it looks failed.

test/nuttx-config.js
```diff
@@ -0,0 +1,122 @@
+/* Copyright 2016 Samsung Electronics Co., Ltd.
```
chokobole: okay, renaming is a good idea! but the problem still remains like below, I wrote this driver to run on the device, and device cannot run throughout whole tests, and I don't know what's the best idea to workaround. 
The reason I divided tests to 8 groups, isn't becauses there's a special reason, but just to be small enough to run on the device, and switching the group help developers run whole tests. Maybe we need discussions what we can do the most for this.

tools/check_test.js
yichoi: If that's the problem, then how about using `EventEmitter.prototype.removeListener`?

src/module/iotjs_module_testdriver.cpp
```diff
@@ -0,0 +1,77 @@
...
+    JObject jtimer = handler.GetArg(0)->GetProperty("handler");
+
+    TimerWrap* timer_wrap = reinterpret_cast<TimerWrap*>(jtimer.GetNative());
+    IOTJS_ASSERT(timer_wrap != NULL);
+    IOTJS_ASSERT(timer_wrap->jobject().IsObject());
+
+    bool has_active_reqs = uv__has_active_reqs(env->loop());
+    bool has_closing_handler = env->loop()->closing_handles != NULL;
+
+    int alive = !has_active_reqs && !has_closing_handler;
```
zherczeg: bool alive

test/run_pass/attrs.js
```diff
@@ -0,0 +1,202 @@
...
+ * limitations under the License.
+ */
+
+(function() {
+  var attrs = {
+    'test_cwd.js': {
+      skip: ['all'],
+      reason: "this test had @STDOUT=COMMAND[pwd], but there's no way to " +
+      "check out current directory path with js driver. So it skips " +
+      "temoporalily."
```
zherczeg: The file name is run_pass, but it seems these tests are failed. I think this is somewhat misleading.

test/nuttx-config.js
chokobole: @yichoi I renamed configure file as you mentioned.  Also, I made the driver that selects the testset by given argument. Please check this out!

src/module/iotjs_module_testdriver.cpp
chokobole: @zherczeg  I did, thanks for the review!

chokobole:
If you are agree to run tests with js driver then, I suggest a few things
* ``Exclude process.exit test`` : because it is single process test driver, I believe the purpose and the real way to check out whether this works is that we run process.exit and check the program really are terminated.
* ``Exclude process.nextTick`` : in order to test asynchronous test-caess, I already used process.nextTick for remaining callback to be executed. Mixed using process.nextTick makes me chaotic. I can't control the sequence of executions.
* ``Exclude process.cwd`` : is there a smart way to compare the output with the real current working directory?
* ``Move process_exitcode*.js to run_pass/`` : in order to keep concept, i did assert.fail() on purpose, and i think it's redundant.

test/run_pass/attrs.js
chokobole: @zherczeg As i left comments below, i hope to remove this one from our testcases.

yichoi:
Great work, LGTM.

It seems that we can enable `nextTick` tests by using `setTimeout` in test driver. However changing this in JavaScript file causes jerryscript assertion failure(ICE: Assertion 'ECMA_STRING_IS_REF_EQUALS_TO_ONE (string_p)' failed at deps/jerry/jerry-core/ecma/base/ecma-literal-storage.c(ecma_free_string_list):43.
Error: ERR_FAILED_INTERNAL_ASSERTION.) It can be either IoT.js issue or JerryScript issue, I'll open another issue to fix it.

Other than nextTick, I agree on the rest of your last comment.

test/run_pass/issue/issue-266.js
```diff
@@ -26,7 +26,6 @@ server.listen(port);
 
 server.on('connection', function(socket) {
   socket.on('data', function(data) {
-    console.log('data', data);
```
LaszloLango: I think should change to an assert to do the same as before

test/run_pass/test_console.js
```diff
@@ -13,9 +13,6 @@
  * limitations under the License.
  */
 
-/*
-  @STDOUT=FILE[test_console_stdout.txt]
```
LaszloLango: Where is the expected output test?

zherczeg:
LGTM

test/run_pass/test_console.js
chokobole: @LaszloLango Using js driver, I don't have a good idea to check out to compare console output with the real output which js made.

LaszloLango: I'd remove them from `function testset` at least.

LaszloLango:
@yichoi, the mentioned assertion is related to an unreleased string value. I think that will be an IoT.js issue.

yichoi:
@LaszloLango I will investigate assertion issue at IoT.js side first. If not resolved then I will investigate JerryScript side later.

PR #452, opened 2016-10-13
Remove redundant behavior from build script

chokobole:
Since the way of building nuttx is changed, these make and copy libraries commands are not required any more.


zherczeg:
LGTM

PR #459, opened 2016-10-14
Update document about Raspberry Pi.

nova0821:
- How to enable I2C interface on Raspberry Pi.
- How to cross compile on linux.

### DIFF
--- a/docs/Build-for-RPi2.md
+++ b/docs/Build-for-RPi2.md
@@ -3,10 +3,11 @@
 ### Prerequisite
 #### Linux
 
-Install arm linux cross compiler.
+Install arm linux cross compiler and i2c library.
 
 ``` bash
-sudo apt-get install gcc-arm-linux-gnueabihf g++-arm-linux-gnueabihf
+sudo apt-get install gcc-arm-linux-gnueabihf g++-arm-linux-gnueabihf libi2c-dev
+sudo mv /usr/arm-linux-gnueabihf/include/linux/i2c-dev.h /usr/arm-linux-gnueabihf/include/linux/i2c-dev.h.kernel
 ```
 
 #### macOS
--- a/docs/Setting-Raspberry-Pi-2.md
+++ b/docs/Setting-Raspberry-Pi-2.md
@@ -57,6 +57,27 @@ append your nameservers
    * proxy environment
    * apt-get proxy
 
+### Enable the I2C interface
+From the command line type:
+```bash
+sudo raspi-config
+```
+This will launch raspi-config utility.
+   * Select "9 Advanced Options"
+   * Select "A6 I2C" <br />
+
+The screen will ask you to enable I2C interface.
+   * Select "Yes"
+   * Select "Ok"
+   * Select "Finish" to return to the command line.<br />
+
+Install utilities
+```bash
+sudo apt-get update
+sudo apt-get install -y i2c-tools libi2c-dev
+```
+
+Reboot Raspberry Pi
 
 ### Reference
 * https://www.raspberrypi.org/documentation/installation/installing-images/linux.md


docs/Build-for-RPi2.md
```diff
@@ -3,10 +3,11 @@
 ### Prerequisite
 #### Linux
 
-Install arm linux cross compiler.
+Install arm linux cross compiler and i2c library.
 
 ``` bash
-sudo apt-get install gcc-arm-linux-gnueabihf g++-arm-linux-gnueabihf
+sudo apt-get install gcc-arm-linux-gnueabihf g++-arm-linux-gnueabihf i2c-tools libi2c-dev
+sudo mv /usr/arm-linux-gnueabihf/include/linux/i2c-dev.h /usr/arm-linux-gnueabihf/include/linux/i2c-dev.h.kernel
```
akosthekiss: Why is this renaming needed? I'd be reluctant to tamper with apt-installed files.

Moreover, is the installation of i2c-related packages mandatory? If yes, why is the macOS section not updated?

nova0821: 'libi2c-dev' package is mandatory. It include 'i2c-dev.h' file which has same name in Linux kernel. That's why renaming is needed. It rename original 'i2c-dev.h' to 'i2c-dev.h.kernel' in host file system, but not in arm include folder. That's why it is needed to change file name manually. If you have any other solution, please let me know.

I think 'i2c-tools' is not mandatory. I'll remove it.

akosthekiss: The two 'i2c-dev.h' files are in different directories, I guess. Isn't this rather an include path setup problem? Does the arm cross compiler pick up the native header or the native compiler finds the cross-installed header?

If the libi2c-dev package is mandatory, does that mean that mac builds will not work anymore?

nova0821: It's include path setup problem, but If we want to fix it completely, we need to copy include folder from raspberry pi file system to ubuntu and add option 'sysroot' when cross-compile. (maybe original 'i2c-dev.h' file in raspberry pi is renamed to 'i2c-dev.h.kernel' and new '
i2c-dev.h' is generated, so it will work fine) It will work with MacOS using this approach.
However, it is not possible to cross-compile until copying them from Raspberry Pi.

This approach is too complex, so I found another approach. It is simply rename header file in arm include folder, so cross-compiler will use header file in ubuntu file system. I also think it's not perfect approach, but it works.

nova0821: In addition, copying 'i2c-dev.h' file to IoT.js project will solve simply this problem, but its license is GPL. That's why I can't copy it.

akosthekiss: Because of licensing, you cannot push the header file to the repository. But can you not copy it to an arbitrary location, even to the project directory, as a build preparatory step? Does it matter where the preprocessor finds the header, from licensing perspective?

nova0821: @akosthekiss  I found another I2C library under MIT license. (https://github.com/roadnarrows-robotics/i2c)

I2c codes in Intel's mraa library(https://github.com/intel-iot-devkit/mraa/blob/master/src/i2c/i2c.c) also use it.

How do you think modify it for us? If we have our own source codes for I2C, it is not needed to install libi2c-dev package.

akosthekiss: Sounds good to me. (Note, I'm not that familiar with I2C implementations. All my above comments were along the lines of what's favourable from build system or user perspective.) The licensing of the linked code seems to fit the purposes of IoT.js, as far as I can tell.

yichoi:
Let's find alternative way

Issue #449, opened 2016-10-13
Add support for JavaScript Array in binding layer


yichoi:
- [ ] Add array support in binding layer
  - [ ] create array / check if object is array / ...
  - [ ] add support to easily iterate over array elements (esp. for sparse array, maybe in jerry?)
  - [ ] ... (TBD)
- [ ] Use array support in IoT.js
  - [ ] in `i2c` module, many APIs should use array as input parameter or callback argument
  - [ ] in `buffer` module, enable `new Buffer(array)`
  - [ ] in `http_incoming` module, IncomingMessage.headers should be array type
  - ... (TBD)

nova0821:
I think it will be better to support more APIs in binding layer. For example, push(add) and pop(remove) for numbered index and named index will be very helpful.

glistening:
I made a PR for array creating and checking for char array. Please see #460.

RedCarrottt:
Is there any progress in this issue? If anyone did not made a progress, I want to handle it.

glistening:
We intentionally don't put more time into this issue since we don't want to make APIs bigger. We want to keep minimal and necessary APIs.

glistening:
All the issues posted are in progress or in pending for some reason. Therefore I recommend you suggest your proposal in issue. You can refer to the guide in [Development-Process](https://github.com/Samsung/iotjs/wiki/Development-Process), especially in [Feature developmentprocess](https://github.com/Samsung/iotjs/wiki/Development-Process#feature-development-process).

glistening:
`buffer module, enable new Buffer(array)` - done
`in http_incoming module, IncomingMessage.headers should be array type` - landed in #776 
`add support to easily iterate over array elements` - We want to keep the API minimal. Let's discuss this item when we need this API.

akhilkedia:
@glistening Right now, there is no support to easily iterate over JS array elements on the C side.
This feature will be extremely useful in a OCF Things SDK I am currently working on.
Is support for this planned in the near future, or will it not be supported to keep APIs minimal?

LaszloLango:
@akhilkedia you can call [`jerry_get_array_length`](http://jerryscript.net/api-reference/#jerry_get_array_length) and [`jerry_get_property_by_index`](http://jerryscript.net/api-reference/#jerry_get_property_by_index) directly until the IoT.js binding hasn't done.

akhilkedia:
@LaszloLango Thanks!

PR #460, opened 2016-10-14
Add array creation and testing API in binding layer (#449)

glistening:
There was no binding API for creating and testing array.
This patch provides the basic API for char array.

### DIFF
--- a/src/iotjs_binding.cpp
+++ b/src/iotjs_binding.cpp
@@ -27,6 +27,12 @@ namespace iotjs {
 #define JVAL_IS_UNDEFINED(val_p) \
     jerry_value_is_undefined(*val_p)
 
+#define JVAL_IS_BOOLEAN(val_p) \
+    jerry_value_is_boolean(*val_p)
+
+#define JVAL_IS_NUMBER(val_p) \
+    jerry_value_is_number(*val_p)
+
 #define JVAL_IS_STRING(val_p) \
     jerry_value_is_string(*val_p)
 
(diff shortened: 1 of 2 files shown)


yichoi:
@glistening DCO is not included

glistening:
I am sorry. I included DCO just now.

And this patch includes following changes:
- I added blank lines between function definitions to make 2 line blanks consistently in iotjs_binding.cpp
- I changed the order of definition of JVAL_IS_XXX same to the declaration order in iotjs_binding.h

src/iotjs_binding.cpp
```diff
@@ -96,6 +101,21 @@ JObject::JObject(const String& v) {
 }
 
 
+JObject::JObject(uint32_t len, const char* data) {
+  _obj_val = jerry_create_array(len);
+
+  if (data != NULL) {
+    for (uint32_t i = 0; i < len; i++) {
+      jerry_value_t val = jerry_create_number(double(data[i]));
```
zherczeg: is this type casting works? It is normally (double) data[i]

LaszloLango: +1 to use C type cast

glistening: Yes. It works in my environment (gcc-4.8 for x86_64-linux and arm-linux). I've seen this style since sevral years ago. Probably it is a new syntax from new version C or C++. However, you're right. for portability, I will change to normal style. Thank you for your review.

zherczeg:
LGTM

PR #464, opened 2016-10-17
Implement I2C module's APIs: Scan, Read, ReadByte

nova0821:
- Implement I2C APIs for Linux platform (Scan, Read, ReadByte)
- Tested on Raspberry Pi

### DIFF
--- a/src/platform/iotjs_module_i2c-linux-general.inl.h
+++ b/src/platform/iotjs_module_i2c-linux-general.inl.h
@@ -67,7 +67,10 @@
 #define I2C_SMBUS_BLOCK_MAX 32
 #define I2C_SMBUS_READ 1
 #define I2C_SMBUS_WRITE 0
+#define I2C_NOCMD 0
+#define I2C_SMBUS_BYTE 1
 #define I2C_SMBUS_I2C_BLOCK_DATA 8
+#define I2C_MAX_ADDRESS 128
 
 
 namespace iotjs {
@@ -134,6 +137,17 @@ int I2cSmbusAccess(int fd, uint8_t read_write, uint8_t command, int size,
 }
 
 
+int I2cSmbusReadByte(int fd) {
+  I2cSmbusData data;
+  int result;
+
+  result = I2cSmbusAccess(fd, I2C_SMBUS_READ, I2C_NOCMD, I2C_SMBUS_BYTE, &data);
+
+  // Mask one byte from result (data.byte).
+  return result >= 0 ? 0xFF & data.byte : -1;
+}
+
+
 int I2cSmbusReadI2cBlockData(int fd, uint8_t command, uint8_t* values,
                              uint8_t length) {
   I2cSmbusData data;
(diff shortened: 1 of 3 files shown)


src/platform/iotjs_module_i2c-linux-general.inl.h
```diff
@@ -134,6 +136,16 @@ int I2cSmbusAccess(int fd, uint8_t read_write, uint8_t command, int size,
 }
 
 
+int I2cSmbusReadByte(int fd) {
+  I2cSmbusData data;
+  int result;
+
+  result = I2cSmbusAccess(fd, I2C_SMBUS_READ, I2C_NOCMD, I2C_SMBUS_BYTE, &data);
+
+  return result >= 0 ? 0x0FF & data.byte : -1;
```
LaszloLango: Please avoid using literals

src/platform/iotjs_module_i2c-linux-general.inl.h
```diff
@@ -212,6 +259,61 @@ void OpenWorker(uv_work_t* work_req) {
...
+  I2cReqData* req_data = i2c_req->req();
+
+  int result;
+  req_data->buf_data = AllocBuffer(128);
+  req_data->buf_len = 128;
+
+  for (int i = 0; i < 128; i++) {
+    ioctl(fd, I2C_SLAVE_FORCE, i);
+
+    if ((i >= 0x30 && i <= 0x37) || (i >= 0x50 && i <= 0x5F)) {
```
LaszloLango: ditto

zherczeg:
LGTM

src/platform/iotjs_module_i2c-linux-general.inl.h
nova0821: @LaszloLango It is just bit masking for data.byte.
Does it need to make value '0x0FF' to variable?

src/platform/iotjs_module_i2c-linux-general.inl.h
nova0821: @LaszloLango between 0x30 and 0x37, 0x50 and 0x5F is a range of read function here. Does it need to make them to variables? (# define SOMETHING)

128 is need to change to I2C_MAX_ADDRESS.

src/platform/iotjs_module_i2c-linux-general.inl.h
nova0821: @LaszloLango I'll fix it. It looks better to define value for 0x0FF

src/platform/iotjs_module_i2c-linux-general.inl.h
nova0821: ditto

src/platform/iotjs_module_i2c-linux-general.inl.h
yichoi: I think 0xFF is just for masking so not needed to be maintained as literals. @LaszloLango Is it necessary ?

src/platform/iotjs_module_i2c-linux-general.inl.h
yichoi: I think '128' is good to be maintained as literals however 0x30, 0x37, 0x50, 0x5F is originated from other I2C related code. It would be enough to add comments which describes the meaning of these values.

nova0821:
@LaszloLango @yichoi I changed '128' to 'I2C_MAX_ADDRESS' and added comments for another values. Please review again.

src/platform/iotjs_module_i2c-linux-general.inl.h
```diff
@@ -134,6 +137,17 @@ int I2cSmbusAccess(int fd, uint8_t read_write, uint8_t command, int size,
...
 
 
+int I2cSmbusReadByte(int fd) {
+  I2cSmbusData data;
+  int result;
+
+  result = I2cSmbusAccess(fd, I2C_SMBUS_READ, I2C_NOCMD, I2C_SMBUS_BYTE, &data);
+
+  // Mask one byte from result (data.byte).
+  return result >= 0 ? 0xFF & data.byte : -1;
```
LaszloLango: What is the return value of this function? 0 and -1 are type of `I2cError` or not?

nova0821: 0 is not returned value. (It is part of checking condition.)
Return values are (0xFF & data.byte) and -1. (It's not type of I2cError)

This function is modified from RoadNarrows-robotics I2C library, so I used original return value from it. All functions from I2C library return result value or -1 if it has an error.

LaszloLango:
LGTM

Issue #28, opened 2015-05-19
Complete fs module


ILyoan:
Complete implementation of [fs module](https://github.com/Samsung/iotjs/wiki/IoT.js-API:-File-System).

Required Items:
- [ ] ~~fs.access()~~
- [ ] ~~fs.accessSync()~~
- [x] ~~fs.exist() (deprecated)~~
- [x] ~~fs.existSync() (deprecated)~~
- [x] fs.stat()
- [x] fs.statSync()
- [ ] ~~fs.chmod()~~
- [ ] ~~fs.chmodSync()~~
- [ ] ~~fs.chown()~~
- [ ] ~~fs.chownSync()~~
- [x] fs.rename()
- [x] fs.renameSync()
- [x] fs.close()
- [x] fs.closeSync()
- [x] fs.open()
- [x] fs.openSync()
- [x] fs.read()
- [x] fs.readSync()
- [x] fs.readFile()
- [x] fs.readFileSync()
- [x] fs.write()
- [x] fs.writeSync()
- [x] fs.writeFile()
- [x] fs.writeFileSync()
- [ ] ~~fs.appendFile()~~
- [ ] ~~fs.appendFileSync()~~
- [ ] ~~fs.truncate()~~
- [ ] ~~fs.truncateSync()~~
- [x] fs.mkdir()
- [x] fs.mkdirSync()
- [x] fs.rmdir()
- [x] fs.rmdirSync()
- [x] fs.readdir()
- [x] fs.readdirSync()
- [x] fs.unlink()
- [x] fs.unlinkSync()

glistening:
I checked File System's basic APIs works in our main supporting platforms. So I close it.

PR #466, opened 2016-10-18
Implements fs.unlink(Sync) and fs.writeFile(Sync)

glistening:
This patch is for issue #28.

- Implements fs.writeFile and fs.writeFileAsync
- Implements fs.unlink and fs.unlinkSync
- Adds testcases for above file system APIs
- Add documentation for above APIs
- Fix wrong description in existing File System APIs

### DIFF
--- a/docs/IoT.js-API-File-System.md
+++ b/docs/IoT.js-API-File-System.md
@@ -30,14 +30,13 @@
  * `bytesRead: Number`
  * `buffer: Buffer`
 
-#### fs.readSync(fd, buffer, offset, length, position, callback)
+#### fs.readSync(fd, buffer, offset, length, position)
 
 * `fd: Int` - file descriptor.
 * `buffer: Buffer` - buffer that the data will be written to.
 * `offset: Number` - offset of the buffer where to start writing.
 * `length: Number` - number of bytes to read.
 * `position: Number` - specifying where to start read data from the file, if `null`, read from current position.
-
 
 #### fs.readFile(path[, options], callback)
 
@@ -51,7 +50,7 @@
 
 Reads entire file asynchronously.
 
-#### fs.readFileSync(path[, options], callback)
+#### fs.readFileSync(path[, options])
 
 * `path: String` - file path to be opened.
 * `options: Object` - options for the operation.
@@ -79,9 +78,29 @@ Reads entire file synchronously.
 * `length: Number` - number of bytes to write.
 * `position: Number` - specifying where to start write data to the file, if `null`, read from current position.
 
+#### fs.writeFile(path, data, [, options], callback)
+
+* `path: String` - file path that the `data` will be written
+* `data: Buffer` - buffer that contains data
+* `options: Object` - options for the operation
+* `callback: Function(err)` - callback function
+ * `err: Error`
+
+Writes entire `data` to the file specified by `path` asynchronously.
+
+#### fs.writeFileSync(path, data, [, options])
+
+* `path: String` - file path that the `data` will be written
+* `data: Buffer` - buffer that contains data
+* `options: Object` - options for the operation
+
+Writes entire `data` to the file specified by `path` synchronously.
+
 #### fs.close(fd, callback)
 
 * `fd: Int` - file descriptor.
+* `callback: Function(err) `
+ * `err: Error`
 
 Closes the file of fd asynchronously.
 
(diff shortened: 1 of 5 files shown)


src/module/iotjs_module_fs.cpp
```diff
@@ -332,6 +332,23 @@ JHANDLER_FUNCTION(Rmdir) {
 }
 
 
+JHANDLER_FUNCTION(Unlink) {
+  JHANDLER_CHECK(handler.GetThis()->IsObject());
+  JHANDLER_CHECK(handler.GetArgLength() >= 1);
```
yichoi: Is it better to express like "handler.GetArgLength() > 0" ?

chokobole:
This is not directly related to this commit, but as you did already, I left some, too.
While I was implementing the test driver, I also found
* In ``fs.open``, mode can be String, too! (if you check ``convertMode`` in src/fs.js_)
* ``fs.exists`` and ``fs.existsSync`` aren't specified there.
* callback of ``fs.write`` aren't marked the same with ``fs.read``, and i think bytesWrite is better to be correspondent to the one with ``fs.read``

docs/IoT.js-API-File-System.md
```diff
@@ -91,6 +110,20 @@ Closes the file of fd asynchronously.
 
 Closes the file of fd synchronously.
 
+#### fs.unlink(path, callback)
+
+* `path: String` - file path to be removed
+* `callback: Function(err) `
+ * `err: Error`
```
glistening: Above Left space has meaning. `err` is the sub item of `callback`. Without left space, `err` will be shown in the same level of `callback`. Please see current document in IoT.js-API-File-System.md. There are many usages of this pattern for describing callback's parameters.

src/js/fs.js
```diff
@@ -275,6 +275,76 @@ fs.readFileSync = function(path) {
...
+  checkArgString(path);
+  checkArgBuffer(data);
+
+  var fd = fs.openSync(path, 'w');
+  var len = data.length;
+  var bytesWritten = 0;
+
+  while (true) {
+    try {
+      var n = fs.writeSync(fd, data, bytesWritten, len - bytesWritten, 0);
```
nova0821: For the performance, declare variable outside of while.

src/js/fs.js
```diff
@@ -275,6 +275,76 @@ fs.readFileSync = function(path) {
...
 
 
+fs.writeFile = function(path, data, callback) {
+  checkArgString(path);
+  checkArgBuffer(data);
+  checkArgFunction(callback);
+
+  var fd;
+  var len;
+  var bytesWritten;
```
nova0821: It looks better to declare in one line.

src/module/iotjs_module_fs.cpp
glistening: Although I think that is problem of preference, if you want, I will replace all of the existing usage of >= 1 (e.g., `Close`, `Stat`, `Rmdir`) to > 0 in the iotjs_module_fs.cpp. Please let me know.

src/js/fs.js
```diff
@@ -275,6 +275,76 @@ fs.readFileSync = function(path) {
...
+      break;
+    }
+  }
+  fs.closeSync(fd);
+  return bytesWritten;
+};
+
+
 fs.mkdir = function(path, mode, callback) {
   if (typeof mode === 'function') callback = mode;
```
nova0821: We have function ``isFunction()`` in util.js. How do you think change checking function to use it?

glistening: OK. I will replace all above pattern with `isFunction()` (e.g., `mkdir`, `rmdir`, `unlink`)

src/js/fs.js
glistening: I cannot find explicit rule in IoT coding style guide. Although I think it is also the problem of taste. (there was a project that prohibits declaring several variables in one line. I can't remember the project name.). I will put them in one line. 

In addition, I will fix the year information.

glistening:
@chokobole, as you mentioned, there are many description in API reference. I suspect File System is the only case. I don't want to interleve this implementation patch with existing documentation errors. In my opinion, it could be an independent patch for updating API documentation.

src/js/fs.js
nova0821: I found a document related this, it says the performance is depends on JS engine(and almost same). I don't know which is faster in JerryScript, but if you want to leave it, please do it. (Don't need to change this line.)

LaszloLango:
Please fix travis failures

glistening:
I'm sorry for travis failure. I fixed it.

src/js/fs.js
zherczeg: Not relevant, all locals are globals from function perspective. So never write:

    for (var i = 0; i < 5; i++) for (var i = 0; i < 5; i++) {}

Because it will not work as in C

zherczeg:
LGTM

PR #473, opened 2016-10-20
Fix bugs of test driver

chokobole:
* On Nuttx, output path is set to 'mnt/sdcard' forcely
* On Nuttx, workaround bug of loading module
* Fix bug of wrong default values for options


zherczeg:
LGTM

Issue #439, opened 2016-10-12
Change IoT.js source code from C++ to C


yichoi:
In order to maximize portability for IoT devices, let's change IoT.js source code from c++ to c.

yichoi:
Done

PR #477, opened 2016-10-25
Re-implement String in c

yichoi:
Related issue: #439 

This is the first commit of c reimplementation. I propose this kind of file-by-file manner transition for productivity and cooperation. (Our code will look ugly in the middle of the transition, though.)

* I propose some common concepts for following works:

 - Coding style

    current style:
    ```
    namespace iotjs {
     class String {
       String();
       ~String();
       int Size();
     };
    }
    ```
    proposing c style:
    ```
    typedef struct {
    } IOTJS_VALIDATED_STRUCT(iotjs_string_t);
    iotjs_string_create();
    iotjs_string_destroy(iotjs_string_t* str);
    iotjs_string_size(iotjs_string_t* str);
    ```
 - Use nested struct to simulate private class member

   above struct will be extended as below:
    ```
    typedef struct {
    } iotjs_string_t_impl_t;
    typedef struct {
    #ifndef NDEBUG
      unsigned flag_create; /* it will hold some magic sequence for validation */
    #endif
      iotjs_string_t_impl_t unsafe; /* Only wizards will directly manipulate the members */
    } iotjs_string_t;
    ```
 - Use IOTJS_VALIDATED_STRUCT_* to simulate c++ method
 - Automate the validation of struct create/destroy (by using magic sequence check in debug mode)

* This patch also includes some string-only works and more modifications:
 - Reworked string so that empty string points some valid string
 - Remove .build.default.config and leave only default.config

Any opinion would be appreciated.


src/iotjs.cpp
```diff
@@ -98,7 +98,9 @@ static void CleanupModules() {
 static bool RunIoTjs(JObject* process) {
   // Evaluating 'iotjs.js' returns a function.
 #ifndef ENABLE_SNAPSHOT
-  JResult jmain = JObject::Eval(String(iotjs_s, iotjs_l), false);
+  iotjs_string_t code = iotjs_string_create_with_buffer((char*)iotjs_s,
```
zherczeg: In C pointers are usually preferred, iotjs_string_t* rather than iotjs_string_t. This could remain though, but looks a bit unusual for me.

src/iotjs_def.h
```diff
@@ -73,12 +73,72 @@
...
+  IOTJS_DECLARE_THIS(iotjs_classname_t, x);
+
+#else
+
+#define IOTJS_VALIDATED_STRUCT(iotjs_classname_t) \
+iotjs_classname_t##_impl_t; \
+typedef struct { \
+  unsigned _flag_create; \
+  iotjs_classname_t##_impl_t _unsafe; \
+} iotjs_classname_t;
```
zherczeg: What about using unions? That could reduce the memory consumption which is important on embedded systems.

zherczeg: No, it cannot be used if the first member is equal to IOTJS_INVALID_MAGIC_SEQUENCE. Ok, it cannot be done.

src/iotjs_string.c
```diff
@@ -0,0 +1,169 @@
...
+  return iotjs_string_create_with_size(data, strlen(data));
+}
+
+
+iotjs_string_t iotjs_string_create_with_size(const char* data, unsigned size) {
+  iotjs_string_t str;
+  IOTJS_VALIDATED_STRUCT_CONSTRUCTOR(iotjs_string_t, &str);
+
+  IOTJS_ASSERT(data != NULL);
+  IOTJS_ASSERT(strnlen(data, size) == size); // forbidden due to ambiguity
```
zherczeg: Do iotjs strings can not hold \0 characters? It is allowed in js.

src/iotjs_string.c
```diff
@@ -0,0 +1,169 @@
...
+
+  return str;
+}
+
+
+iotjs_string_t iotjs_string_create_with_buffer(char* buffer, unsigned size) {
+  iotjs_string_t str;
+  IOTJS_VALIDATED_STRUCT_CONSTRUCTOR(iotjs_string_t, &str);
+
+  IOTJS_ASSERT(buffer != NULL);
```
zherczeg: Do we need a strnlen(data, size) check here as well?

src/iotjs_string.c
```diff
@@ -0,0 +1,169 @@
...
+    iotjs_buffer_release(this->_data);
+  }
+}
+
+
+void iotjs_string_reserve(iotjs_string_t* str, unsigned capacity) {
+  IOTJS_VALIDATED_STRUCT_METHOD(iotjs_string_t, str);
+
+  if (this->_cap >= capacity)
+    return;
```
zherczeg: Is it not allowed to decrease capacity?

src/iotjs_string.h
```diff
@@ -0,0 +1,66 @@
...
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+
+typedef struct {
+  char* _data;
+  unsigned _size;
+  unsigned _cap;
+} IOTJS_VALIDATED_STRUCT(iotjs_string_t);
```
zherczeg: Underscore is C++ style, do we want to keep it?

zherczeg:
Fine. LGTM

src/iotjs_string.h
yichoi: @LaszloLango I think in this case every members are fixed-size data, not header. Anyway I moved the data to the end.
@zherczeg I agree with you. I removed underscores from member variable names. (And after that, I changed simulated-`this` into `_this` to avoid confusion with c++ `this`)

PR #485, opened 2016-10-27
Re-implement JVal, JArgList, JHandlerInfo into c-style struct

yichoi:
Related issue: #439

### DIFF
--- a/src/iotjs_binding.cpp
+++ b/src/iotjs_binding.cpp
@@ -70,34 +70,32 @@ JObject::JObject(const JObject& other) {
 
 
 JObject::JObject(bool v) {
-  _obj_val = JVal::Bool(v);
+  _obj_val = iotjs_jval_bool(v);
   _unref_at_close = false;
 }
 
 
 JObject::JObject(int v) {
-  _obj_val = JVal::Number(v);
+  _obj_val = iotjs_jval_number(v);
   _unref_at_close = true;
 }
 
 
 JObject::JObject(double v) {
-  _obj_val = JVal::Number(v);
+  _obj_val = iotjs_jval_number(v);
   _unref_at_close = true;
 }
 
 
 JObject::JObject(const char* v) {
   IOTJS_ASSERT(v != NULL);
-  _obj_val = jerry_create_string(reinterpret_cast<const jerry_char_t*>(v));
+  _obj_val = iotjs_jval_raw_string(v);
   _unref_at_close = true;
 }
 
 
 JObject::JObject(const iotjs_string_t& v) {
-  _obj_val = jerry_create_string_sz(
-      reinterpret_cast<const jerry_char_t*>(iotjs_string_data(&v)),
-                                            iotjs_string_size(&v));
+  _obj_val = iotjs_jval_string(&v);
   _unref_at_close = true;
 }
 
@@ -138,8 +136,8 @@ JObject::~JObject() {
 
 
 void JObject::init() {
-  _null = new JObject(JVal::Null(), false);
-  _undefined = new JObject(JVal::Undefined(), false);
+  _null = new JObject(iotjs_jval_null(), false);
+  _undefined = new JObject(iotjs_jval_undefined(), false);
 }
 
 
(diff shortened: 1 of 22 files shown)


src/iotjs_binding.cpp
```diff
@@ -472,178 +470,320 @@ bool JResult::IsException() const {
...
-    delete _argv[i];
+void iotjs_jargs_destroy(iotjs_jargs_t* jargs) {
+  IOTJS_VALIDATED_STRUCT_DESTRUCTOR(iotjs_jargs_t, jargs);
+
+  IOTJS_ASSERT(_this->argv == NULL || _this->argc > 0);
+  IOTJS_ASSERT(_this->argc <= _this->capacity);
+
+  if (_this->capacity > 0) {
+    for (int i = 0; i < _this->argc; ++i) {
+      delete _this->argv[i];
```
zherczeg: delete is not C. Is this intentional?

src/iotjs_binding.cpp
```diff
@@ -70,34 +70,32 @@ JObject::JObject(const JObject& other) {
 
 
 JObject::JObject(bool v) {
-  _obj_val = JVal::Bool(v);
+  _obj_val = iotjs_jval_bool(v);
```
LaszloLango: I'd remove these one line wrapper methods and use the jerryscript API here.

`_obj_val = jerry_create_boolean (v);`

So at other places.

zherczeg:
Reasonable.
LGTM

src/iotjs_binding.cpp
yichoi: Yes it is intentional. The type of `_this->argv[i]` is still `iotjs::JObject*` (c++ class), so I didn't removed the `delete` operator yet. I'll fix it after I change `class iotjs::JObject` into `struct iotjs_jobject_t`.

src/iotjs_binding.cpp
yichoi: Well, it is true that they have same meaning, but I didn't removed that wrapper function because:
* The explicit type of `_obj_val` is `JRawValueType`, not `jerry_value_t`. 
* I think compiler can inline the wrapper function. (checked at x86_64)
* We can add more common optimization in iotjs_jval_* (e.g. returning pre-declared value instead of calling jerry api, at least until we have LTO support also in iotjs)

yichoi: But I also think that there might be some redundancy here or somewhere.
I think old c++ binding layer had 2 purposes in general: 1. support C++ language, and 2. make binding easier(memory mgnt, etc). Since we are changing our code into c, if some code is written only for the first purpose, it should be removed. Please let me know if there are something more to remove.

yichoi:
Slight modification: changed remaining `JArgList` into c struct in nuttx-only code, and use static jerry_value_t for simple types instead of calling jerry api every time (at least until we have LTO in iotjs), as referred in [this comment](https://github.com/Samsung/iotjs/pull/485#discussion_r85461093)

zherczeg:
still LGTM

src/iotjs_binding.h
```diff
@@ -284,20 +311,25 @@ class JHandlerInfo {
...
   static JRawValueType name(const JRawValueType func_obj_val, \
                             const JRawValueType this_val, \
                             const JRawValueType args_p [], \
                             const JRawLengthType args_cnt) { \
     JRawValueType ret_val = jerry_create_undefined(); \
-    JHandlerInfo handler(func_obj_val, this_val, &ret_val, args_p, args_cnt); \
-    ___ ## name ## _native(handler); \
+    iotjs_jhandler_t jhandler; \
+    iotjs_jhandler_initialize(&jhandler, func_obj_val, this_val, &ret_val, \
+                                         args_p, args_cnt); \
```
yichoi: Updated in [this commit](https://github.com/Samsung/iotjs/pull/485/commits/828c02bce5cd693fd3f622db01a6a1ce0cba61f1)

yichoi:
@LaszloLango It is pending. Please review again.

PR #491, opened 2016-10-31
Enable TCP for NuttX

chokobole:
* On Nuttx Side, EOF doesn't come with the event when polled.  In order to workaround this issue, when called `Socket.prototype.end`,  I marked the last chunk with magic string, `\\e\\n\\d` to check if it is the last buffer.
* For this implementation, `writeAfterEnd` should be checked with only flag `state.ended`, because when ReadableStream is ended, it calls `onSocketEnd`. In the default mode, `state.allowHalfOpen` is false, it calls `destroySoon`, in the end, automatically calls `Socket.prototype.end`. Therfore, for example, it is possible when socket receiving `end` event, `Socket.prototype.end` was already called so that it might raise an error `write after end`.
screenshot_2016-11-01_21-12-36

### DIFF
--- a/src/iotjs_def.h
+++ b/src/iotjs_def.h
@@ -144,6 +144,7 @@ typedef struct { \
 #include <uv.h>
 #include <assert.h>
 #include <limits.h> /* PATH_MAX */
+#include <string.h>
 
 
 #endif /* IOTJS_DEF_H */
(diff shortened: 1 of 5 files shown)


zherczeg:
LGTM

PR #495, opened 2016-11-02
Update README.md about how to configure TCP on NuttX

chokobole:

### DIFF
--- a/targets/nuttx-stm32f4/README.md
+++ b/targets/nuttx-stm32f4/README.md
@@ -48,7 +48,7 @@ $ cd nuttx/tools
 $ ./configure.sh stm32f4discovery/usbnsh
 ```
 
-Now you can configure nuttx like either of below. For convenience, we provide built-in configure file for you.
+Now you can configure nuttx like either of below. For convenience, we provide built-in configure file for you. (This configure file is equipped with modules specified as `always`. For `optional` modules, you might follow instructions below.)
 ```bash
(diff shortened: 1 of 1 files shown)


targets/nuttx-stm32f4/README.md
```diff
@@ -86,6 +86,25 @@ Followings are the options to set:
 * Enable `Application Configuration -> System Libraries and NSH Add-ons -> IoT.js`
 * Enable all childs of `Application Configuration -> System Libraries and NSH Add-ons -> readline() Support` (for those who wants to use readline)
 
+##### 3.1. For those who want to use `net` module
```
nova0821: It looks it is additional option for ``net`` module. 
I think it is part of ``Followings are the options to set:``, but its numbering  is ``3.1``.

How do you think above options for ``Common`` and make ``For net module``, like:
- Common
 * Change `Build Setup -> Build Host Platform` from _Windows_ to [_Linux_|_OSX_]
 * Enable `System Type -> FPU support`
...
- For ``net`` module
 * Enable `System Type -> STM32 Peripheral Support -> Ethernet MAC`
 * Disable `System Type -> STM32 Peripheral Support -> USART2`
...

nova0821: In line 84, ``Network Device/PHT Support`` should be ``Network Device/PHY Support``.
Could you fix it with this commit?

PR #498, opened 2016-11-03
Fix a bug which generate assert when i2c api is called.

nova0821:
- After merging c string apis, it doesn't work.
- Modify string generate code and add detory code.


LaszloLango:
typo in commit message: `detory` -> `destroy`

nova0821:
@LaszloLango Thanks, I fixed it.

LaszloLango:
LGTM

PR #499, opened 2016-11-03
Fix build break on nuttx release mode

sae-bom:
When including <stdio.h>, stdio.h which is from nuttx OS uses stdbool.h
It makes conflicts with enum bool type in iotjs_def.h
To bypass it, I changed the order of included header files and type definition


zherczeg:
LGTM

PR #505, opened 2016-11-07
Fix readdir testcase failure

glistening:
I did not push some directories for readdir test.
My git add command was ignored since git does not permit
committing empty directory.
I add dummy files in previously empty directories.
    

chokobole:
@glistening  How about trying one of these suggestions either `touch .gitkeep` inside the empty diretory or add `!resources` in .gitignore?

yichoi:
+1 on `.gitkeep`

glistening:
Originally I want to describe the reason of seemingly dummy file explicitly and transparently. Perhaps I will forgot the intent after some time passes. However, at least two committers seem to want to make the file hidden and make name short. Also, I agree this change may be enough. Thank you for review.

zherczeg:
LGTM

PR #511, opened 2016-11-08
Implement UDP functions and Refactor TCP and UDP modules

chokobole:
* socket.getsockname
* socket.setBroadcast
* socket.setTTL
* socket.setMulticastTTL
* socket.setMulticastLoopback
* socket.addMembership
* socket.dropMembership


src/js/net.js
```diff
@@ -224,6 +238,11 @@ Socket.prototype.setKeepAlive = function(enable, delay) {
 };
 
 
+Socket.prototype.address = function() {
+  return this._getsockname();
+};
```
nova0821: How do you think rename function name ``_getsockname()`` to ``address``, because it looks just one line wrapper if there isn't any issue?

src/js/net.js
```diff
@@ -501,6 +520,18 @@ Server.prototype.listen = function() {
 };
 
 
+Server.prototype.address = function() {
+  if (this._handle && this._handle.getsockname) {
+    var out = {};
+    this._handle.getsockname(out);
+    // TODO(bnoordhuis) Check err and throw?
+    return out;
+  } else {
```
zherczeg: No need else here.

src/module/iotjs_module_tcp.cpp
```diff
@@ -468,6 +465,46 @@ JHANDLER_FUNCTION(SetKeepAlive) {
...
+      uv_inet_ntop(AF_INET, &a4->sin_addr, ip, sizeof ip);
+      port = ntohs(a4->sin_port);
+      iotjs_jval_set_property_string_raw(obj, "address", ip);
+      iotjs_jval_set_property_string_raw(obj, "family", "IPv4");
+      iotjs_jval_set_property_number(obj, "port", port);
+      break;
+    }
+
+    default:
+      iotjs_jval_set_property_string_raw(obj, "address", "");
```
zherczeg: Please add a break here.

src/module/iotjs_module_udp.cpp
```diff
@@ -307,52 +282,117 @@ JHANDLER_FUNCTION(Close) {
...
   IOTJS_ASSERT(!"Not implemented");
 
   iotjs_jhandler_return_null(jhandler);
+#endif
 }
 
 
-JHANDLER_FUNCTION(AddMembership) {
+void SetMembership(iotjs_jhandler_t* jhandler, uv_membership membership) {
+#if !defined(__NUTTX__)
```
glistening: This code assumes all OS except for nuttx works. But if iotjs is going to support other OS which is not similar to linux, we need to change all the `!defined(__NUTTX__)` to `!defined(__NUTTX__) && !defined(__OTHER_OS_)`. What about saying explicitly with currently supported OSes here (e.g. ` __linux__`) , not negative form.

src/module/iotjs_module_udp.cpp
```diff
@@ -307,52 +282,117 @@ JHANDLER_FUNCTION(Close) {
...
-JHANDLER_FUNCTION(GetSockeName) {
-  IOTJS_ASSERT(!"Not implemented");
+GetSockNameFunction(UdpWrap, udp_handle, uv_udp_getsockname);
 
-  iotjs_jhandler_return_null(jhandler);
+JHANDLER_FUNCTION(GetSockeName) {
+  DoGetSockName(jhandler);
 }
 
+#define X(fn)                                                                 \
```
zherczeg: I would prefer a longer name than X. EXEC perhaps?

chokobole: First time i dind't name it because its role is very simple to set options.  If you prefer longer name, then how about `IOTJS_UV_SET_SOCKOPT` ?

src/module/iotjs_module_udp.cpp
chokobole: IMHO, it is good to guard with more exceptional and self-explaining condition. And this API is supposed to work with os where running libuv fine, so the `__linux__` isn't enough.  and the same things happen when we start to support `darwin` or `window`.  So i am wondering what other's opinion are

nova0821:
LGTM

src/module/iotjs_module_udp.cpp
glistening: I agree.

src/module/iotjs_module_udp.cpp
zherczeg: Ok, lets use your suggestion.

chokobole: fixed naming issue! thanks for the review!

chokobole:
Fixed commit message
* Capitalize module names
* Modify module to modules

zherczeg:
LGTM

PR #514, opened 2016-11-08
Re-implement JObjectWrap, HandleWrap in C

yichoi:
Related issue: https://github.com/Samsung/iotjs/issues/439


src/iotjs_objectwrap.h
```diff
@@ -20,26 +20,30 @@
...
- public:
-  explicit JObjectWrap(const iotjs_jval_t* jobject);
-  virtual ~JObjectWrap();
+typedef struct {
+  iotjs_jval_t jobject;
+} IOTJS_VALIDATED_STRUCT(iotjs_jobjectwrap_t);
 
-  iotjs_jval_t* jobject();
+void iotjs_jobjectwrap_initialize(iotjs_jobjectwrap_t* jobjectwrap,
+                                 const iotjs_jval_t* jobject,
```
glistening: ditto

src/iotjs_handlewrap.c
```diff
@@ -0,0 +1,110 @@
+/* Copyright 2015 Samsung Electronics Co., Ltd.
```
glistening: 2015-2016

src/module/iotjs_module_buffer.cpp
```diff
@@ -24,9 +24,11 @@ namespace iotjs {
 
 
 BufferWrap::BufferWrap(const iotjs_jval_t* jbuiltin, size_t length)
-    : JObjectWrap(jbuiltin)
-    , _buffer(NULL)
+    : _buffer(NULL)
```
glistening: Use two space

yichoi:
minor coding style fixed. Thanks @glistening

src/iotjs_objectwrap.c
```diff
@@ -17,44 +17,30 @@
 #include "iotjs_objectwrap.h"
 
 
-namespace iotjs {
+void iotjs_jobjectwrap_initialize(iotjs_jobjectwrap_t* jobjectwrap,
+                                  const iotjs_jval_t* jobject,
+                                  uintptr_t jhandle,
```
glistening: `uintptr_t` is introduced in C99. What is our supporting C version? May I assume C99? There are also many use cases using fixed width integer types (e.g. `uint32_t`, ...), which is also introduced in C99, by including `<stdint.h>`. If we decide to support C99 after switching C++ to C, the code `typedef enum { false, true } bool;` in iotjs_def.h could be replaced with including `<stdbool.h>`

src/iotjs_handlewrap.h
```diff
@@ -41,31 +46,36 @@ namespace iotjs {
...
- public:
-  HandleWrap(const iotjs_jval_t* jobject, uv_handle_t* handle);
-  // jobject: Object that connect with the uv handle
 
-  virtual ~HandleWrap();
+typedef struct {
+  iotjs_jobjectwrap_t jobjectwrap;
+  uv_handle_t* handle;
+  OnCloseHandler on_close_cb;
+} IOTJS_VALIDATED_STRUCT(iotjs_handlewrap_t);
```
glistening: I looked up the definition of IOTJS_VALIDATED_STRUCT, and I found 
`typedef struct iotjs_classname_t { \  unsigned flag_create; \` in iotjs_def.h. 
`flag_create` is assignd to either IOTJS_VALID_MAGIC_SEQUENCE or IOTJS_INVALID_MAGIC_SEQUENCE, which is defined as 0xfee1c001, 0xfee1badd respectively. What about chainging `unsigned` to 32bit data type or poitner type, or making the magic number 16 bit?

src/iotjs_objectwrap.c
yichoi: Good point! JerryScript already has a dependency of C99 since it uses `stdbool.h` so It would be natural that IoT.js also follow C99 dependency. 
I am considering to replace `typedef enum { false, true } bool;` with `<stdbool.h>` but not this time :)

zherczeg:
LGTM

PR #529, opened 2016-11-10
ReqWrap should wrap only uv_req_t type, not some custom type

yichoi:


nova0821:
LGTM

PR #533, opened 2016-11-11
Re-implement ReqWrap and more modules in C, and rework validated struct

yichoi:
* Re-implement ReqWrap in C
* Re-implement Dns, Fs, Tcp in C
* Rework validated struct
  - Make private implementation (`iotjs_classname_t##_impl_t`) as its first member
  - To avoid extra work to find real starting address of the struct,
  - And to add more ASSERTs in Wrappers for safer execution

### DIFF
--- a/src/iotjs_reqwrap.h
+++ b/src/iotjs_reqwrap.h
@@ -1,4 +1,4 @@
-/* Copyright 2015 Samsung Electronics Co., Ltd.
+/* Copyright 2015-2016 Samsung Electronics Co., Ltd.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -22,7 +22,9 @@
 #include "iotjs_binding.h"
 
 
-namespace iotjs {
+#ifdef __cplusplus
+extern "C" {
+#endif
 
 
 // UV request wrapper.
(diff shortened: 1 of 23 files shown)


src/iotjs_reqwrap.h
```diff
@@ -22,59 +22,38 @@
...
 
-  // To retrieve javascript callback funciton object.
-  const iotjs_jval_t* jcallback();
 
-  // To retrieve pointer to uv request.
-  T* req();
+void iotjs_reqwrap_initialize(iotjs_reqwrap_t* reqwrap,
+                              const iotjs_jval_t* jcallback,
+                              uv_req_t* request,
+                              void* wrap);
```
nova0821: I think this line should be same line with ``uv_req_t* request, ``.

According to our coding style guideline - ``Return type should be on the same line as function name and parameters if it fits. If not, break between them aligned with the first argument.``

I think this rule reference to google's style guideline - https://google.github.io/styleguide/cppguide.html#Function_Declarations_and_Definitions

It shows some example, like below:

```c
ReturnType ClassName::ReallyLongFunctionName(Type par_name1, Type par_name2,
                                             Type par_name3) {
  DoSomething();
  ...
}
```

src/module/iotjs_module_fs.c
```diff
@@ -200,15 +275,12 @@ JHANDLER_FUNCTION(Read) {
     return;
   }
 
-  uv_buf_t uvbuf = uv_buf_init(reinterpret_cast<char*>(data + offset),
-                               length);
+  uv_buf_t uvbuf = uv_buf_init((char*)(data + offset), length);
```
glistening: Do we need char* casting here?

src/module/iotjs_module_fs.c
```diff
@@ -242,15 +314,12 @@ JHANDLER_FUNCTION(Write) {
...
```
glistening: ditto

yichoi:
fixed @nova0821 @glistening

hs0225:
LGTM

nova0821:
LGTM

PR #535, opened 2016-11-11
Use c99 option when compiling c files

yichoi:


zherczeg:
LGTM

PR #537, opened 2016-11-14
Travis should check nuttx build

yichoi:
Added automated nuttx build script and IoT.js precommit script

### DIFF
--- a/tools/precommit.py
+++ b/tools/precommit.py
@@ -1,0 +1,138 @@
+#!/usr/bin/env python
+
+# Copyright 2016 Samsung Electronics Co., Ltd.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+import argparse
+import sys
+import os
+from common_py import path
+from common_py.system.filesystem import FileSystem as fs
+from common_py.system.executor import Executor as ex
+from common_py.system.platform import Platform
+
+TESTS=['host', 'rpi2', 'nuttx', 'misc'] # TODO: support darwin
+BUILDTYPES=['debug', 'release']
+
+
+def parse_option():
+    parser = argparse.ArgumentParser(
+         description='IoT.js pre-commit script.',
+         epilog='If no arguments are given, runs full test.')
+    parser.add_argument('--test', choices=TESTS, action='append')
+    parser.add_argument('--buildtype', choices=BUILDTYPES, action='append')
+
+    option = parser.parse_args(sys.argv[1:])
+    if option.test is None:
+        option.test = TESTS
+    if option.buildtype is None:
+        option.buildtype = BUILDTYPES
+    return option
+
+
+def setup_nuttx_root(nuttx_root):
+    # Step 1
+    fs.maybe_make_directory(nuttx_root)
+    fs.chdir(nuttx_root)
+    if fs.exists('nuttx'):
+        fs.chdir('nuttx')
+        ex.check_run_cmd('git', ['pull'])
+        fs.chdir('..')
+    else:
+        ex.check_run_cmd('git', ['clone',
+                                 'https://bitbucket.org/nuttx/nuttx.git'])
+    if fs.exists('apps'):
+        fs.chdir('apps')
+        ex.check_run_cmd('git', ['pull'])
+        fs.chdir('..')
+    else:
+        ex.check_run_cmd('git', ['clone',
+                                 'https://bitbucket.org/nuttx/apps.git'])
+
+    # Step 2
+    fs.maybe_make_directory(fs.join(nuttx_root, 'apps', 'system', 'iotjs'))
+    for file in fs.listdir(fs.join(path.PROJECT_ROOT,
+                                   'targets', 'nuttx-stm32f4', 'app')):
+        fs.copy(fs.join(path.PROJECT_ROOT,'targets','nuttx-stm32f4','app',file),
+                fs.join(nuttx_root, 'apps', 'system', 'iotjs'))
+
+    # Step 3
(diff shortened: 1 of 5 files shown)


tools/precommit.py
```diff
@@ -0,0 +1,141 @@
...
+                    'nuttx',
+                    '.config.default'),
+            fs.join('.config'))
+
+
+def build_nuttx(nuttx_root, buildtype):
+    fs.chdir(fs.join(nuttx_root, 'nuttx'))
+    try:
+        nuttx_bin = fs.join(nuttx_root, 'nuttx', 'nuttx.bin')
+        if fs.exists(nuttx_bin):
```
chokobole: We don't need to check and remove `nuttx_bin`, make will overwrite for us.

tools/precommit.py
```diff
@@ -0,0 +1,141 @@
...
+            fs.remove(nuttx_bin)
+        if buildtype == "release":
+            ex.run_cmd('make', ['IOTJS_ROOT_DIR=' + path.PROJECT_ROOT, 'R=1'])
+        else:
+            ex.run_cmd('make', ['IOTJS_ROOT_DIR=' + path.PROJECT_ROOT, 'R=0'])
+
+        if fs.exists(nuttx_bin):
```
chokobole: I think it's better to check with `return code` which returns from `ex.run_cmd`

tools/precommit.py
```diff
@@ -0,0 +1,141 @@
...
+            return True
+        else:
+            print "Cannot create nuttx.bin"
```
chokobole: How about chaning error messasge to `Failed to build nuttx` like others such as jerry, libtuv, or IoT.js

tools/precommit.py
```diff
@@ -0,0 +1,141 @@
...
+    # Step 3
+    fs.chdir(fs.join(nuttx_root, 'nuttx', 'tools'))
+    ex.check_run_cmd('./configure.sh', ['stm32f4discovery/usbnsh'])
+    fs.chdir('..')
+    fs.copy(fs.join(path.PROJECT_ROOT,
+                    'targets',
+                    'nuttx-stm32f4',
+                    'nuttx',
+                    '.config.default'),
+            fs.join('.config'))
```
chokobole: is join needed?

tools/precommit.py
```diff
@@ -0,0 +1,141 @@
...
+                              '--target-board=stm32f4dis',
+                              '--jerry-heaplimit=78'])
+            if not build_nuttx(nuttx_root, buildtype):
+                ex.fail('nuttx ' + buildtype + ' build failed')
+
+    elif test == "misc":
+        build("debug", ['--no-snapshot', '--jerry-lto'])
+        build("debug", ['--iotjs-minimal-profile'])
+
+    else:
```
chokobole: I think we need a more delicate error message for `darwin` as long as we specify to support platform like above. ['host', 'rpi2', 'nuttx', 'darwin', 'misc']

tools/precommit.py
```diff
@@ -0,0 +1,141 @@
...
+
+
+def setup_nuttx_root(nuttx_root):
+    # Step 1
+    fs.maybe_make_directory(nuttx_root)
+    fs.chdir(nuttx_root)
+    if fs.exists('nuttx'):
+        fs.chdir('nuttx')
+        ex.check_run_cmd('git', ['pull'])
+        fs.chdir('..')
```
chokobole: For the future,  how about  do `git checkout $version` here?

tools/precommit.py
```diff
@@ -0,0 +1,141 @@
...
+
+import argparse
+import sys
+import os
+from common_py import path
+from common_py.system.filesystem import FileSystem as fs
+from common_py.system.executor import Executor as ex
+from common_py.system.platform import Platform
+
+TESTS=['host', 'rpi2', 'nuttx', 'darwin', 'misc']
```
chokobole: I think these kind of common variable should be moved somewhere like `common_py.system.platform`

tools/precommit.py
yichoi: Since our nuttx build guide doesnt' specify the exact version of nuttx, I think it would be better to make the buildbot to follow the latest version of nuttx.
When buildbot fails, that would be the right time to fix the nuttx version or update our nuttx build script.

tools/precommit.py
yichoi: I made this `TESTS` variable just to separate the tests into multiple categories so that our 5 buildbots in travis can run it in parallell. I think it can be changed if the number of buildbots changes, or if we decide to separate the tests in different manner, etc. Moreover, it doesnt' fit in any of our build configuration variable: target-board, target-os, target-arch. 
I think it would be better to keep it here.

tools/precommit.py
yichoi: I just removed the `darwin` option. Maybe someone can improve it later.

yichoi:
Updated the commit as @chokobole commented

tools/precommit.py
```diff
@@ -0,0 +1,138 @@
...
+            return True
+        else:
+            print 'Failed to build nuttx'
+            return False
+    except OSError as err:
+        print 'Failed to build nuttx: %s' % err
+        return False
+
+
+def build(buildtype, args):
```
chokobole: def build(buildtype, args=[])

tools/precommit.py
```diff
@@ -0,0 +1,138 @@
...
+
+
+option = parse_option()
+if os.getenv('TRAVIS') == None:
+    ex.check_run_cmd('tools/check_signed_off.sh')
+
+for test in option.test:
+    if test == "host":
+        for buildtype in option.buildtype:
+            build(buildtype, [])
```
chokobole: build(buildtype)

tools/precommit.py
```diff
@@ -0,0 +1,138 @@
...
+        fs.chdir('apps')
+        ex.check_run_cmd('git', ['pull'])
+        fs.chdir('..')
+    else:
+        ex.check_run_cmd('git', ['clone',
+                                 'https://bitbucket.org/nuttx/apps.git'])
+
+    # Step 2
+    fs.maybe_make_directory(fs.join(nuttx_root, 'apps', 'system', 'iotjs'))
+    for file in fs.listdir(fs.join(path.PROJECT_ROOT,
```
chokobole: please put space between ,

tools/precommit.py
chokobole: Okay, I agree

yichoi:
Fixed style issues as @chokobole commented

PR #540, opened 2016-11-17
Use clang-format (version 3.8) to check coding style

yichoi:
Most of the changes by clang-format seemed reasonable.

And if there are some changes that might not have been expected, I propose to accept the clang-format's modification to reduce our overhead on manually reviewing the coding style.

Or any suggestion on `.clang-format` file would be appreciated.


hs0225:
LGTM

PR #542, opened 2016-11-18
Rework iotjs_reqwrap_t and hide destructors of object wraps

yichoi:
Related issue: https://github.com/Samsung/iotjs/issues/439

On translating ReqWrap into iotjs_reqwrap_t, `delete reqwrap;` was changed into `iotjs_reqwrap_destroy(reqwrap); IOTJS_RELEASE(reqwrap);`.
Since this style would be hard to write and understand, I changed it into `iotjs_reqwrap_dispatched()`, following `ReqWrap::Dispatched()` of Node.js. This made it possible to hide more destructors of iotjs_*wrap_t.

### DIFF
--- a/src/module/iotjs_module_dns.c
+++ b/src/module/iotjs_module_dns.c
@@ -15,41 +15,50 @@
 
 #include "iotjs_def.h"
 
+#include "iotjs_module_dns.h"
+
 #include "iotjs_reqwrap.h"
 #include "uv.h"
 
 
-typedef struct {
-  iotjs_reqwrap_t reqwrap;
-  uv_getaddrinfo_t req;
-} IOTJS_VALIDATED_STRUCT(iotjs_getaddrinforeqwrap_t);
+#define THIS iotjs_getaddrinfo_reqwrap_t* getaddrinfo_reqwrap
 
-
-#define THIS iotjs_getaddrinforeqwrap_t* getaddrinforeqwrap
-
-void iotjs_getaddrinforeqwrap_initialize(THIS, const iotjs_jval_t* jcallback) {
-  IOTJS_VALIDATED_STRUCT_CONSTRUCTOR(iotjs_getaddrinforeqwrap_t,
-                                     getaddrinforeqwrap);
-  iotjs_reqwrap_initialize(&_this->reqwrap, jcallback, (uv_req_t*)&_this->req,
-                           getaddrinforeqwrap);
+iotjs_getaddrinfo_reqwrap_t* iotjs_getaddrinfo_reqwrap_create(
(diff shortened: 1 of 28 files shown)


src/module/iotjs_module_dns.c
```diff
@@ -15,30 +15,36 @@
...
                                      getaddrinforeqwrap);
-  iotjs_reqwrap_initialize(&_this->reqwrap, jcallback, (uv_req_t*)&_this->req,
-                           getaddrinforeqwrap);
+  iotjs_reqwrap_initialize(&_this->reqwrap, jcallback, (uv_req_t*)&_this->req);
+  return getaddrinforeqwrap;
 }
 
 
-void iotjs_getaddrinforeqwrap_destroy(THIS) {
+static void iotjs_getaddrinforeqwrap_destroy(THIS) {
```
glistening: What is `THIS` for? It seems not to be C, but C translated C++ code. Although I can figure out what is `THIS` by searching and thinking, for me, it needs time and hard to understand. Because I don't know the rule you translate C++ to C, and many of reader may not know it has history of traslated from C++. Could you explain what is `THIS` in comment or let it exposed directly, i.e. without macro.

nova0821:
Looks good. LGTM.

src/module/iotjs_module_dns.c
yichoi: Thanks for your review, @glistening. I agree that it will be better to declare the function without `THIS` macro. (It was added just to shorten the line.)

However, since this PR is not directly related to `THIS` macro, if you're OK with the rest of this PR, I think it will be better to close this one, and create a new one to remove `THIS` macros. Would it be OK with you?

src/module/iotjs_module_dns.c
```diff
@@ -94,8 +100,7 @@ static void AfterGetAddrInfo(uv_getaddrinfo_t* req, int status,
 
   iotjs_jargs_destroy(&args);
 
-  iotjs_getaddrinforeqwrap_destroy(req_wrap);
-  IOTJS_RELEASE(req_wrap);
+  iotjs_getaddrinforeqwrap_dispatched(req_wrap);
```
glistening: I feel difficulty in reading very long word like `iotjs_getaddrinforeqwrap` without separator. What about using iotjs_getaddrinfo_reqwrap? There are many cases. As you said above comment, although it is not main change of this patch, if you would like to upload the patch of enhancing readability, I will be happy. Since your patches for C++ to C has been very big (in terms of line or in terms of file changed) and related all the modules tightly. I could not review all the details of your patch  in appropriate time. Sorry about that. Thank you.

PR #547, opened 2016-11-22
Set fs.write max buffer size to 1023 for nuttx

glistening:
When I run testcase fs.writeFile on nuttx, I found it hangs.
fs.writeFiles fails to create file when the data source is about 1.5KB.
It makes zero-size file.
It is similar issue to #434. On nuttx, OS does not suppport writing
large buffer at once. This commit tries to divide writing request
to several write requests with max 1023 bytes.

Also, I run fs tests on nuttx. I update the related API document with the result.


yichoi:
LGTM

PR #549, opened 2016-11-22
Specify project language as C

glistening:
cmake try to validate C++ compiler. However, we don't need
the check anymore since we had migrated from C++ to C.
Also, I set travis language as C.


yichoi:
LGTM

PR #563, opened 2016-11-24
Remove add_custom_xxx in cmake/http-parser/jerry/libtuv.cmake

glistening:
Each add_custom_command is intended to execute touch external libraries.
I think nothing depends on the custom targets (i.e., targetJerry, targetHttpParser, targetLibtuv) so `touch` statement seems dead-code.


glistening:
There are two questions about touch command:

First, when does the custom command run ? I think it would not.
And I build x86_64-linux, arm-linux, arm-nuttx after I remove the code. I have no errors.

Two, why the touch statments need? Could anyone explain to me ? I know there is some case to modify the access time. But I cannot figure out for this case.

yichoi:
LGTM

PR #567, opened 2016-11-29
add /deps/nuttx at .gitignore

wangmir:
When I run the ./tools/precommit.py, nuttx sources were downloaded at /deps/nuttx.

But .gitignore file didn't consider about this, thus git recognized this source files as modified and un committed data.

I think, this single line should be added into .gitignore to prevent confusion.

IoT.js-DCO-1.0-Signed-off-by: Kim, Hyukjoong wangmir@gmail.com

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,9 @@ nbproject
 **.sublime-workspace
 .idea
 
+# deps files
+/deps/nuttx/*
+
 # Random Trash
 *.swp
 *.swo


glistening:
I had not been in trouble because I put the nuttx source out of iotjs source according to the guide for nuttx in https://github.com/Samsung/iotjs/blob/master/targets/nuttx-stm32f4/README.md. We prefer to place nuttx source out of iotjs source tree. We will make change in precommit.py.

Issue #409, opened 2016-09-22
Complete UART module


hs0225:
- [x] JS API
  - [x] initialize
  - [x] read
  - [x] write
  - [x] on
- [x] x86 linux dummy
- [x] arm-linux on Raspberry Pi
- [x] arm-nuttx on STM32F4-discovery
- [x] Create document

sinban04:
I'm going to implement this part, is it right that this part is not implemented yet?
When i see the docs, there is no document related to uart, so i think it's not implemented yet.

sae-bom:
I'm sorry that this part is already in progress. The code for UART on linux is internally almost done.

hs0225:
Done.

PR #570, opened 2016-12-01
Implement UART for linux (#409)

sae-bom:
* UART API on JS
  * constructor
  * open
  * write
  * on('data')
* x86 linux dummy
* arm-linux on Raspberry Pi

### DIFF
--- a/src/js/uart.js
+++ b/src/js/uart.js
@@ -1,0 +1,94 @@
+/* Copyright 2016 Samsung Electronics Co., Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+var EventEmitter = require('events').EventEmitter;
+var util = require('util');
+
+var uart = process.binding(process.binding.uart);
+
+//  VALIDATION ARRAYS
+var BAUDRATE = [0, 50, 75, 110, 134, 150, 200, 300, 600, 1200, 1800, 2400
+                , 4800, 9600, 19200, 38400, 57600, 115200, 230400];
+var DATABITS = [5, 6, 7, 8];
+
+var defaultSettings = {
+  autoOpen: true,
+  baudRate: 9600,
+  dataBits: 8,
+}
+
+function UART(path, options, callback) { //constructor
+  if (!(this instanceof UART)) {
+    return new UART(path, options, callback);
+  }
+
+  if (util.isFunction(options)) {
+    callback = options;
+    options = {};
+  }
+
+  if (util.isUndefined(options)) {
+    options = {};
+  }
+
+  var settings = {};
+  for (var propname in defaultSettings) {
+    settings[propname] = defaultSettings[propname];
+  }
+  for (var propname in options) {
+    settings[propname] = options[propname];
+  }
+
+  EventEmitter.call(this);
+
+  if (!path) {
+    throw new TypeError('No path specified');
+  }
+
+  if (BAUDRATE.indexOf(settings.baudRate) === -1) {
+    throw new TypeError("Invalid 'baudRate': " + settings.baudRate);
+  }
+
+  if (DATABITS.indexOf(settings.dataBits) === -1) {
+    throw new TypeError("Invalid 'databits': " + settings.dataBits);
+  }
+
+  this.path = path;
+  this.options = settings;
+
+  if (settings.autoOpen) {
+    this.open(callback);
+  }
+}
+
+util.inherits(UART, EventEmitter);
+
+UART.prototype.open = function(callback) {
+  uart.open(this, this.path, this.options, function (err) {
+    return process.nextTick(function() {
+      return callback(err);
+    });
+  }.bind(this));
+}
+
+UART.prototype.write = function(buffer, callback) {
+  uart.write(buffer, function(err) {
+    return process.nextTick(function() {
+      return callback(err);
+    });
+  });
+};
+
+module.exports = UART;
--- a/src/module/iotjs_module_uart.c
+++ b/src/module/iotjs_module_uart.c
@@ -1,0 +1,244 @@
+/* Copyright 2015-2016 Samsung Electronics Co., Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "iotjs_def.h"
+#include "iotjs_module_uart.h"
+#include "iotjs_objectwrap.h"
+
+
+#define THIS iotjs_uart_reqwrap_t* uart_reqwrap
+
+
+iotjs_uart_reqwrap_t* iotjs_uart_reqwrap_create(const iotjs_jval_t* jcallback,
+                                                UartOp op) {
+  iotjs_uart_reqwrap_t* uart_reqwrap = IOTJS_ALLOC(iotjs_uart_reqwrap_t);
+  IOTJS_VALIDATED_STRUCT_CONSTRUCTOR(iotjs_uart_reqwrap_t, uart_reqwrap);
+
+  iotjs_reqwrap_initialize(&_this->reqwrap, jcallback, (uv_req_t*)&_this->req);
+
+  _this->req_data.path = iotjs_string_create("");
+  _this->req_data.op = op;
+
+  return uart_reqwrap;
+}
+
+
+static void iotjs_uart_reqwrap_destroy(THIS) {
+  IOTJS_VALIDATED_STRUCT_DESTRUCTOR(iotjs_uart_reqwrap_t, uart_reqwrap);
+  iotjs_reqwrap_destroy(&_this->reqwrap);
+  iotjs_string_destroy(&_this->req_data.path);
+  IOTJS_RELEASE(uart_reqwrap);
+}
+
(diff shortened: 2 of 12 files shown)


src/js/uart.js
```diff
@@ -0,0 +1,119 @@
...
+  xon: false,
+  highWaterMark: 16 * 1024,
+}
+
+function UART(path, options, callback) { //constructor
+  if (!(this instanceof UART)) {
+    return new UART(path, options, callback);
+  }
+
+  if (typeof options === 'function') {
```
nova0821: Prefer to use ``util.isFunction()`` here. (For consistency of code style)

src/js/uart.js
```diff
@@ -0,0 +1,119 @@
...
+    settings[propname] = options[propname];
+  }
+
+  EventEmitter.call(this);
+
+  if (!path) {
+    throw new TypeError('No path specified');
+  }
+
+  if (typeof settings.baudRate !== 'number') {
```
nova0821: Prefer to use ``util.isNumber()``.

src/js/uart.js
```diff
@@ -0,0 +1,119 @@
...
+
+  FLOWCONTROLS.forEach(function(control) {
+    if (typeof settings[control] !== 'boolean') {
+      throw new TypeError("Invalid '" + control + "' is not boolean");
+    }
+  });
+
+  this.path = path;
+  this.options = settings;
+  this.opening = false;
```
nova0821: I can't fine use ``this.opening`` variable in whole js code.(Only assign a value in variable) Is it necessary? Or is this variable is for the future implementation?

src/js/uart.js
```diff
@@ -0,0 +1,119 @@
...
+  this.closing = false;
```
nova0821: Same as ``this.opening``.

src/js/uart.js
```diff
@@ -0,0 +1,119 @@
...
+
+  this.open(callback);
+}
+
+util.inherits(UART, EventEmitter);
+
+UART.prototype.open = function(callback) {
+  this.opening = true;
+  uart.open(this, this.path, this.options, function afterOpen(err) {
```
nova0821: Is it needed to have callback function name ``afterOpen``? I think it can be just ``function(err)`` here.

src/js/uart.js
```diff
@@ -0,0 +1,119 @@
...
+  if (STOPBITS.indexOf(settings.stopBits) === -1) {
+    throw new TypeError("Invalid 'stopbits': " + settings.stopbits);
+  }
+
+  if (PARITY.indexOf(settings.parity) === -1) {
+    throw new TypeError("Invalid 'parity': " + settings.parity);
+  }
+
+  FLOWCONTROLS.forEach(function(control) {
+    if (typeof settings[control] !== 'boolean') {
```
nova0821: Prefer to use ``util.isBoolean()``.

src/module/iotjs_module_uart.c
```diff
@@ -0,0 +1,238 @@
...
+  iotjs_jargs_destroy(&jargs);
+  iotjs_uart_reqwrap_dispatched(req_wrap);
+}
+
+
+#define UART_ASYNC(op)                                                 \
+  do {                                                                 \
+    uv_loop_t* loop = iotjs_environment_loop(iotjs_environment_get()); \
+    uv_work_t* req = iotjs_uart_reqwrap_req(req_wrap);                 \
+    uv_queue_work(loop, req, op##WorkerUart, AfterUARTWork);           \
```
nova0821: IMHO, is there any reason add ``Uart`` after ``Worker``?

test/run_pass/test_uart.js
```diff
@@ -0,0 +1,37 @@
...
+  }
+
+  uart.write("Hello? from Ubuntu\n", function(err) {
+  });
+
+  uart.on('data', function(data) {
+    console.log(data.toString());
+  });
+
+});
```
nova0821: I can't see call ``open`` method directly. Is it needed to ``open`` method as ``public``?

test/run_pass/test_uart.js
```diff
@@ -0,0 +1,37 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+var Uart = require('uart');
+
+function sleep(n) {
```
nova0821: Please check this function is needed. I can't find any line to use this function.

src/js/uart.js
```diff
@@ -0,0 +1,119 @@
...
+  if (typeof options === 'function') {
+    callback = options;
+    options = {};
+  }
+
+  if (util.isUndefined(options)) {
+    options = {};
+  }
+
+  var settings = defaultSettings;
```
hs0225: If several objects are created, `defalutSettings` are changed continuously. I think it's a good idea to make a copy of `defaultSettings`.

src/platform/iotjs_module_uart-linux-general.inl.h
```diff
@@ -0,0 +1,111 @@
...
+  saio.sa_restorer = NULL;
+  sigaction(SIGIO, &saio, NULL);
+
+  fcntl(fd, F_SETFL, FNDELAY);
+  fcntl(fd, F_SETOWN, getpid());
+  fcntl(fd, F_SETFL, O_ASYNC);
+
+  struct termios options;
+  tcgetattr(fd, &options);
+  options.c_cflag = B115200 | CS8 | CLOCAL | CREAD;
```
hs0225: Options are configurable. 
It is able to change options in javascript, but these are fixed in native.

src/module/iotjs_module_uart.c
sae-bom: It's because I found out that without adding Uart after Worker, it makes a conflict with Worker functions in i2c module

test/run_pass/test_uart.js
```diff
@@ -0,0 +1,39 @@
...
+var device = '/dev/ttyUSB0';
+var options = {
+  baudRate: 115200,
+  dataBits: 8,
+  autoOpen: true,
+}
+
+var serial = new uart(device, options, function(err){
+  assert.equal(err, null);
+  console.log('open done');
```
sae-bom: On line 23 autoOpen is set to true, so it would be opened before executing callback function.

src/js/uart.js
```diff
@@ -0,0 +1,95 @@
...
+
+var EventEmitter = require('events').EventEmitter;
+var util = require('util');
+var assert  = require('assert');
+
+var uart = process.binding(process.binding.uart);
+
+//  VALIDATION ARRAYS
+var BAUDRATE = [0, 50, 75, 110, 134, 150, 200, 300, 600, 1200, 1800, 2400
+                , 4800, 9600, 19200, 38400, 57600, 115200, 230400];
```
hs0225: `BAUDRATE` is unused variable. How about using this to check for valid baudrate like `DATABITS`?

src/js/uart.js
```diff
@@ -0,0 +1,95 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+var EventEmitter = require('events').EventEmitter;
+var util = require('util');
+var assert  = require('assert');
```
glistening: I cannot find any use of assert in uart.js. Could you remove this line? And it seems two spaces before equal sign in this line.

src/platform/iotjs_module_uart-linux-general.inl.h
```diff
@@ -0,0 +1,173 @@
...
+  options.c_iflag = IGNPAR;
+  options.c_oflag = 0;
+  options.c_lflag = 0;
+  tcflush(fd, TCIFLUSH);
+  tcsetattr(fd, TCSANOW, &options);
+}
+
+
+void signal_handler_IO(int status) {
+  sleep(2);
```
glistening: Could you explain why sleep is needed ?

src/platform/iotjs_module_uart-linux-general.inl.h
```diff
@@ -0,0 +1,173 @@
...
+    if (errno == EINTR) {
+      continue;
+    }
+
+    req_data->error = kUartErrWrite;
+    return;
+
+  } while (0);
+
+  sleep(2);
```
glistening: Ditto

src/platform/iotjs_module_uart-linux-general.inl.h
sae-bom: This signal_handler function is called by both of 'write' function and 'read' function. So I found out it makes a bug somewhere. I intended to bypass the bug with sleep function. But now, I'm planning to fix the bug with using uv_poll() instead of using signal handler directly.

src/platform/iotjs_module_uart-linux-general.inl.h
```diff
@@ -0,0 +1,168 @@
...
+#include <unistd.h>
+
+#include "module/iotjs_module_uart.h"
+
+void signal_handler_IO(int status);
+
+iotjs_jval_t jthis;
+int fd;
+uv_poll_t poll_handle;
+bool WRITING = false;
```
glistening: 1) I think we can put `static` not to pollute global name space.
2) Why do you use all capital variable name for `WRITING`? If there is a compelling reason, please change the name.

src/platform/iotjs_module_uart-linux-general.inl.h
```diff
@@ -0,0 +1,168 @@
...
+    case 5:
+      return CS5;
+  }
+  return -1;
+}
+
+void read_cb(uv_poll_t* req, int status, int events) {
+  while (WRITING) {
+    sleep(0);
+  }
```
glistening: You seems to use `WRITING` variable to avoid race condition between writing thread and reading thread? If so, I think we must use some synchronization primitives.

src/platform/iotjs_module_uart-linux-general.inl.h
```diff
@@ -0,0 +1,168 @@
...
+
+void OpenWorkerUart(uv_work_t* work_req) {
+  UART_WORKER_INIT_TEMPLATE;
+
+  jthis = req_data->jthis;
+  fd = open(iotjs_string_data(&req_data->path), O_RDWR | O_NOCTTY | O_NDELAY);
+
+  if (fd == -1) {
+    req_data->error = kUartErrOpen;
+  } else {
```
glistening: I got SIGSEGV when I ran run_pass/test_uart.js without attaching any uart device.
We need to stop proceeding when open returned -1. Otherwise, the call to uv_poll_start will lead to SIGSEGV.

glistening:
LGTM ( I approved this PR yesterday without comment "LGTM". )

PR #571, opened 2016-12-03
add comments for build problems

mulderp:
adding comments during the build might help to understand build problems.

(my build on Darwin breaks at 6) libuv )


mulderp:
I see there is a printing mechanism. Maybe numbering the steps would improve the build script.

mulderp:
the problem on darwin with building libuv is a missing define:

    source/nuttx/uv_platform.h:#define CLOCK_MONOTONIC 1

adding that manualy in source/darwin/uv_linux_clock.c works

Issue #58, opened 2015-05-29
Size of script code buffer


ruben-ayrapetyan:
Hello.

Code buffer size argument of `jerry_api_eval` in `JHANDLER_FUNCTION(Compile, handler)` and `JHANDLER_FUNCTION(CompileNativePtr, handler)` is calculated incorrectly : `sizeof (code)` equals to size of pointer, not buffer size.

Maybe, this could be fixed with the following:

```
diff --git a/src/iotjs_module_process.cpp b/src/iotjs_module_process.cpp
index 282bc09..bd73a3f 100644
--- a/src/iotjs_module_process.cpp
+++ b/src/iotjs_module_process.cpp
@@ -87,7 +87,7 @@ JHANDLER_FUNCTION(Compile, handler){

   char* code = handler.GetArg(0)->GetCString();
   JRawValueType ret_val;
-  jerry_api_eval(code,sizeof(code),true,false,&ret_val);
+  jerry_api_eval(code,strlen(code),true,false,&ret_val);
   JObject::ReleaseCString(code);

   JObject ret(&ret_val);
@@ -116,7 +116,7 @@ JHANDLER_FUNCTION(CompileNativePtr, handler){
   strcat(code,wrapper[1]);

   JRawValueType ret_val;
-  jerry_api_eval(code,sizeof(code),true,false,&ret_val);
+  jerry_api_eval(code,len,true,false,&ret_val);
   JObject::ReleaseCString(code);

   JObject ret(&ret_val);
```

PR #576, opened 2016-12-06
Update submodule tuv (#58) and Use uv API and convention

glistening:
- Update submodule tuv (#58)
- Use uv API and follow the convention
 : Use uv_chdir, rather than uv_cd
 : Use uv_sleep, rather than tuv_usleep
 : Don't use uv's private method directly in testdriver


hs0225:
LGTM

yichoi:
LGTM

PR #580, opened 2016-12-08
Update submodule libtuv and modify libtuv include path

glistening:
- Update submodule libtuv (Samsung/libtuv#59)
- Remove non-existing path from libtuv.cmake's include path


hs0225:
LGTM

PR #583, opened 2016-12-13
Re-implement pin

hs0225:
- Move pin information from pin module to target board module.
- update document about pin.


yichoi:
LGTM

PR #586, opened 2016-12-14
Implement UART.close()

sae-bom:
- Implement UART.close()
- Fix bugs in UART module which didn't free some values
- Update some documents for UART module

### DIFF
--- a/src/platform/iotjs_module_uart-linux-general.inl.h
+++ b/src/platform/iotjs_module_uart-linux-general.inl.h
@@ -20,12 +20,9 @@
 #include <errno.h>
 #include <fcntl.h>
 #include <sys/signal.h>
-#include <termios.h>
 #include <unistd.h>
 
 #include "module/iotjs_module_uart.h"
-
-void signal_handler_IO(int status);
 
 static iotjs_jval_t jthis;
 static int fd;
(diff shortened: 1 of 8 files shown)


src/platform/iotjs_module_uart-linux-general.inl.h
```diff
@@ -135,6 +135,14 @@ void OpenWorkerUart(uv_work_t* work_req) {
 }
 
 
+void UartClose() {
+  uv_close((uv_handle_t*)&poll_handle, NULL);
+  if (fd > 0) {
```
glistening: What do you mean by `fd > 0`? Valid file descriptor is non-negative number and zero is standard input. Therefore you are trying to close all seemingly valid file descriptor except for stdin. What is going about stdout and stderr? If you mean all valid file descriptor, it should be (fd >= 0).

sae-bom: If you see line 31,  `fd` is just declared with no initialization. So It is likely to be set to 0. `fd` can be 0 even though the file descriptor is invalid. If the code should be changed according to your suggestion, `fd` should be initialized to non-positive number on line 31.

Apart from your suggestion, I found a bug here. Line 139 should be executed after line 140's condition. I'll fix the bug immediately.

yichoi:
LGTM

PR #590, opened 2016-12-29
Provide patch file to activate additional features on Nuttx

sae-bom:
Stm32f4-discovery board has 6 uart ports but Nuttx code supports only 4 of them.
So, this commit will provide patch file which can support activating the other 2 ports.
This patch file will be also used by other modules than UART.


yichoi:
LGTM

PR #595, opened 2017-01-10
Improve build script Python 3.x compatibility

galpeter:
The current build script is heavily dependent on
the Python 2.x environment. On some distributions
the default Python version is now 3.x.

The modification allows the build script to run
on both 2.x and 3.x Python versions.
Tested on Python 2.7.13 and 3.6.

### DIFF
--- a/tools/build.py
+++ b/tools/build.py
@@ -14,6 +14,12 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from __future__ import print_function
+try:
+    basestring
+except:
+    # in Python 3.x there is no basestring just str
+    basestring = str
 
 import argparse
 import json
@@ -33,7 +39,7 @@ platform = Platform()
 # Initialize build option.
 def init_option():
     # Check config option.
-    arg_config = filter(lambda x: x.startswith('--config='), sys.argv)
+    arg_config = list(filter(lambda x: x.startswith('--config='), sys.argv))
 
     config_path = path.BUILD_CONFIG_PATH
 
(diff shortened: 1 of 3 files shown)


glistening:
Thank you for patch.

Unfortunately, when I run tools/build.py on ubuntu 14.04/x64 with python 3.4, I see following error:

> $ ./tools/build.py 
Traceback (most recent call last):
  File "./tools/build.py", line 749, in <module>
    option = init_option()
  File "./tools/build.py", line 53, in init_option
    config = json.loads(f.read().encode('ascii'))
  File "/usr/lib/python3.4/json/__init__.py", line 312, in loads
    s.__class__.__name__))
TypeError: the JSON object must be str, not 'bytes'

When I change

>    with open(config_path, 'r') as f:
        config = json.loads(f.read().encode('ascii'))

to

>    with open(config_path, 'rb') as f:
        config = json.loads(f.read().decode('ascii'))

I found tools/build.py works well on my system with both python 2.7.11 and python 3.4.0.

Could you check this additional changes on your system to confirm it also works well with python 3.6.0 ?

galpeter:
@glistening the change works on both 3.6 and 2.7 for me. However I propose an different solution: we could simply just use the `json.load` call which only requires a file object not a string or bytes. So it would look something like this:
```python
with open(config_path, 'r') as f:
    config = json.load(f)
```

What do you think?

glistening:
I checked your new solution works well on my system both using python 2.7.11 and python 3.4. I think it is good. But there had been an issue related to file encoding #404. Some systems shows error without specifying encoding. I am not sure this is same case and I don't know the internal working of `json.load`. If we see similar issue on some system, we may need to change to my solution than yours.

galpeter:
@glistening hmm it could be currently I'm not sure. So I'll update the PR with your changes.

Issue #596, opened 2017-01-11
One document to rule them all


loki04:
It would be lovely if there is only one place for the documentation of iotjs.

1.- Currently there is a minimal introduction of iotjs via gh-pages,
2.- there is a wiki at github which might be the most up-to-date documentation,
3.- and finally there is `doc` sub-directory under the git branches of iotjs - which is based on wiki with differences and outdated content.

How about if there will be only one place for all of those?

Let me just suggest a solution!

Create a separate repository which holds every docs and update the wiki and gh-pages from this. Check out this method: http://www.growingwiththeweb.com/2016/07/enabling-pull-requests-on-github-wikis.html

loki04:
Maybe it will be even better if iotjs just simple uses its `docs` folder and does not spread the information across different places. I am going to share a simple solution very soon.

LaszloLango:
#597 is landed.

PR #597, opened 2017-01-13
One doc to rule them all

loki04:
Restructure docs folder. (#596)

The final goal is to combine different guides and documentation scattered
various places to a common one.

This commit does the followings:
- create new folders to introduce some logic here
- copy the Wiki pages to the repository
- update the links and references in the MD files


docs/build/Build-Script.md
```diff
@@ -0,0 +1,244 @@
...
+
+If given path is not exist, build.py will create it.
+
+```
+./tools/build.py --builddir=./build
+```
+
+--
+#### `--clean`
+With given this option, build.py will claer all the build directory before start new build.
```
polaroi8d: clear

docs/devs/Inside-IoT.js.md
```diff
@@ -0,0 +1,292 @@
...
+```c
+{
+  // Create JavaScript object
+  // It increases reference count in JerryScript side.
+  iotjs_jval_t jobject = iotjs_jval_create();
+
+  // Use `jobject`
+  ...
+
+  // Before jobject goes out of scope, destroy it.
```
polaroi8d: object

docs/devs/Inside-IoT.js.md
```diff
@@ -0,0 +1,292 @@
...
+# libuv Binding
+
+IoT.js is using [libuv](http://libuv.org/) to perform various asynchronous I/O and threading.
+Because IoT.js adopts asynchronous programming model, libuv plays very important role in this project. Actually the [main loop](#event-loop) of IoT.js is libuv event loop waiting I/O event, picks the event, and dispatches it to corresponding event handler function.
+
+You can read [libuv design document](http://docs.libuv.org/en/v1.x/design.html) to get detailed information about asynchronous programming model based on libuv.
+
+## iotjs_handlewrap_t
+
+`iotjs_handlewrap_t` is to bind a Javascript object and a libuv handle(e.g. file descriptor) together.
```
polaroi8d: IMHO whitespace maybe required

docs/devs/Inside-IoT.js.md
```diff
@@ -0,0 +1,292 @@
...
+1. Javascript module calls builtin function to perform I/O applying arguments including callback.
+2. Builtin creates `iotjs_reqwrap_t` to wrap `uv_req_s` and Javascript callback function.
+3. Builtin calls libuv to perform the I/O.
+4. After I/O finished, libuv calls builtin after handler.
+5. Builtin after handler takes `iotjs_reqwrap_t` containing I/O result and Javascript callback function.
+6. Builtin after handler calls Javascript callback.
+7. Builtin after handler release `iotjs_reqwrap_t` by calling `iotjs_*reqwrap_dispatch()`
+
+`iotjs_reqwrap_t` does not inherits `iotjs_handlewrap_t` for wrapping the callback function object.
+Note that `HandleWrap` does not increase reference count of wrapping object. it does not guarantee guarantee liveness of the object even if the wrapper is alive.
```
polaroi8d: It

docs/devs/Inside-IoT.js.md
```diff
@@ -0,0 +1,292 @@
...
+                       path, flags, mode,
+                       AfterAsync);
+```
+
+Since all I/O are treated as non-blocking, calling for async I/O API returns immediately right after request was sent to libuv.
+And then next line of javascript program will be executed immediately without waiting the I/O.
+Thus in the above example 'requested' will be printed out right after file open request was made.
+
+If there were I/O requests, `uv_run()` in the main loop waits by polling the requests until at least one of the request processing were finished.
+When a result for a request was produced, internal part of libuv calls corresponding handler function(let it be after function) back.
```
polaroi8d: whitespace

docs/help/Developer-Tutorial.md
```diff
@@ -0,0 +1,161 @@
...
+```
+
+Then it gives:
+```
+Hello, world!
+```
+
+Pretty simple. But where did `console` come from? `console` is not defined in Global Object according to ECMAScript spec.
+
+The answer is `console` is a builtin module so it should have been `require`ed. However, `console` is a special case so we can use it directly without `require`. This is about Module System which We will cover later.
```
polaroi8d: we

loki04:
I did quick changes based on the comments, but the purpose of this change is to introduce new logic in the documentation instead to change the relevant content. This - the update of the text - can be done later, but first it should be agreed which way the docs of IoT.js will be presented.

yichoi:
It would be better you make this PR with one single commit

glistening:
Thank you for this PR. One doc plan is great! I hope it works well. 

Here are some questions about executing the plan.

1) According to the link ((http://www.growingwiththeweb.com/2016/07/enabling-pull-requests-on-github-wikis.html) in issue #596, we need separate repository for documents and make personal token access token known to travis. Is is secure? Although the link says travis secure variable, I am not sure about this.

2) There are some differences between iotjs docs and iotjs wiki. For example, a) md files in targets/nuttx-stm32f4 exist only in iotjs repository, it does not exist in wiki repository. So we uses full link to reference these md files in wiki files. b) Some filenames are different to corresponding ones. For example, wiki uses prefix "IoT.js-API:", instead of "IoT.js-API". Probably we still need manual works or need script. Could you share your plan about this concerns?

loki04:
In short there is no need for the Wiki in this form. The only extra is the sidebar in the Wiki compared to any MD file in the repository. In other hand the repository-based documentation has a not negligible advantage: there will be a review via pull request which is totally missing at the Wiki right now.

What I suggested in the #596 is more than what you can see in this PR. The link which describes the two way modification of Wiki is a robust solution to take control of each documentation changes, but it is not necessary. Let's discuss the details there if there is a need for such a solution. If you have concerns about the security you can still create an artificial account with a very limited permissions to do deal with the synchronization. This PR is not about to solve #596 , instead create something which can stand on its own feet, and can be the basic of a new repository if you are about to move the documentation to a separate repository.

About your 2nd question... the plan is simple:
1. move the content of Wiki into the repository - (this PR)
2. fix the links to point to the relevant repository files instead of Wiki pages - (this PR)
3. remove the whole wiki except a simple start page (Home.md) which redirects to the main repository file (Readme.md) - (can be done by the owner or a contributor of the Wiki)
4. update, fix the documentation in the repository - (can be done anyone via a simple PR)
5. decide if the current solution is enough or you would like to do the two way synchronization with the Wiki - (based on community decision at #596 )

nova0821:
@loki04 I have a question about your plan number 3. If we follow your plan, we can't use Wiki menu bar, right? (see below screenshot) 

wiki_menu_bar

I think this is very useful feature of Wiki page. (We can check all pages in here, and also can see a hierarchy of all documents.  Do you have any idea about this?

loki04:
@nova0821 Yes, this is the sidebar ( https://guides.github.com/features/wikis/ ). I can confirm that between action steps 3-5, you will not able to use the sidebars, since there will be no Wiki pages. If the community has the decision to adopt the synchronization method between a separate repository and the wiki ( #597 ) you will be able to use the sidebar again.
My opinion is that the lack of the review process and lack of possible PRs for the Wiki are big disadvantages of the current documentation (even if there is or isn't any sidebar).

glistening:
@loki04 I also wanted to remove wiki pages when I joined this project first. So I asked some member who joined this project before me. "Why do we keep wiki pages that is same to docs?", I got the answer that documents in github source repository are not searched in google. Besides the wiki documents could be. After doing several google searches related to iotjs, I agreed to keep our documents in both of the iotjs repo and wiki repo.

About the wiki review process, we have kept the policy of creating/updating documents in iotjs repo first, then applying the reviewed and merged documents into wiki. We don't need the duplicated review process for wiki.

In summary, 
1) we need to keep the wiki pages, and
2) we do not need wiki review process since we already review documents in iotjs repository.

loki04:
@glistening , you are wrong about the indexing method of Google. The search engines can index the master branch of any repository (except the owner of the repository prevents this). Here is the main info about it: https://github.com/robots.txt . I can confirm that the search engines are able to index the master branch. I have several public repositories indexed by Google, and checked several organizations' repositories which have been indexed successfully. So, I think the reason why Google is not able to index iot.js master branch is that the owner of the project (or the organization) prevents the indexing.

On the other hand if the wiki contained the exact content of the `docs` folder it would be wonderful and there wouldn't be such discussion what we have right now :) . As you can see there is no synchronous update on them. Sometimes the Wiki's content is outdated, one or the other has different content, and the naming schema is different. I think there is no doubt that there is a need for a unified, common, and up-to-date documentation. The current process to update Wiki and the docs folder in the repository are not perfect. If you are about to attract developers, maintainers, and users for iot.js the project has to have precise and eye-candy docs as well. I think the minimum is to combine the content of Wiki and docs folder (this is what this PR does).

Anyhow, this just *a pull request* which tries to combine different descriptions, guides, and notes from Wiki, and to do some reorganization of them. So, I suggest to do general discussion about Wiki, github's pages and docs on #596 issue instead. Let this PR as it is to merge the content.

glistening:
Thank you for information about the index method of Google. We need to find out the reason of not indexing of our master branch (including organization policy).

About your comment "no synchronous update on them", in fact, we continuously made efforts to synchronize `docs` folder and wiki pages. If there are differences between corresponding pages, please let us know the different pages (except for wiki only pages and docs only pages). It would be better to use diff tool. But as I mentioned previous comment, the file naming is slightly different. I will ask the reason of the slightly different naming to other colleagues.

About mering this PR, if we merge this PR, we must apply these changes also to wiki pages. If we merge only this one, we are out of synchronous state on the duration. It cannot be done by merely copying them. We must check whether there is broken link, and update sidebars. And if we made the decision of not keeping wiki pages based on your information and suggestion, the synchronizing work is not necessary.

I will gather other member's opinions. Thank you.

glistening:
I confirmed that iotjs master branch is searchable in Google. I tried to search the page `targets/nuttx-stm32f4/README.md`, (which exists only in master branch, not in wiki page), and I can see this page in search result.

glistening:
I discussed with other members. Although it turned out that documents in master branch are searchable, they prefer to keep wiki pages for user's perspective. (e.g., easy navigation, ...) Therefore we need a way to synchronize between iotjs repo and wiki on the condition of keeping wiki pages. Currently my plan is following:

(1) I will begin to apply this PR to wiki. (i.e., applying pages, and checking the validity of urls, and updating sidebar) 

(2) In  duration of (1), I will gather the exceptional cases that needs manual merging during this work. 

(3) We will merge this PR (#597)

(4) If possible, we will make a script for synchronizing automatically. (It will be another PR.)

(5) It would be better to trigger script of (4) automatically whenever docs are modified. It may be based on your first suggestion (using separate doc repository). It will be yet another PR.

Thank you.

yichoi:
LGTM. We will update WiKi pages along with this new restructuring

PR #603, opened 2017-01-20
Improve check_tidy script

galpeter:
List of improvements:
* Now works with both 3.x and 2.7+ Python versions.
* Separate classes for general style and clang-format checks.
* Allowed CMakeLists.txt for style checks.
* The diff for clang-format uses temp files.
* The diff for clang-format is now a unified diff.
* Uses clang-format if clang-format-3.8 is not found.
* If no clang-format found still run the other checks.
* Executor.run_cmd_output now accepts an list instead of a string.

### DIFF
--- a/tools/check_tidy.py
+++ b/tools/check_tidy.py
@@ -14,113 +14,180 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import sys
-import os
+from __future__ import print_function
 import fileinput
+import functools
+import os
+import subprocess
+import tempfile
+
+from distutils import spawn
 
 from check_license import CheckLicenser
-from common_py import path
 from common_py.system.filesystem import FileSystem as fs
 from common_py.system.executor import Executor as ex
 
-TERM_RED = "\033[1;31m"
-TERM_YELLOW = "\033[1;33m"
-TERM_GREEN = "\033[1;32m"
-TERM_BLUE = "\033[1;34m"
-TERM_EMPTY = "\033[0m"
 
+class StyleChecker(object):
 
-column_limit = 80
+    column_limit = 80
 
-count_err = 0
+    def __init__(self):
+        self.count_lines = 0
+        self.count_empty_lines = 0
+        self.errors = []
 
-interesting_exts = ['.c', '.h', '.js', '.py', '.sh', '.cmake']
-clang_format_exts = ['.c', '.h']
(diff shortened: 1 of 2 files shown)


yichoi:
LGTM

tools/check_tidy.py
```diff
@@ -14,113 +14,166 @@
...
-            if is_checked_by_clang(file):
-                formatted = ex.run_cmd_output(' '.join(['clang-format-3.8',
-                                                        '-style=file', file]),
-                                              quiet=True)
-                f = open(file + '.formatted', 'w')
-                f.write(formatted)
-                f.close()
-                ex.check_run_cmd('diff', [file, file+'.formatted'], quiet=True)
-                fs.remove(file + '.formatted')
+    exts_allowed = ['.c', '.h', '.js', '.py', '.sh', '.cmake']
```
LaszloLango: What about .txt files (e.g: CMakeLists.txt)?

yichoi:
sounds reasonable.

tools/check_tidy.py
galpeter: That's a good question. I'm using the same extensions which were already in place. Maybe someone more knowledgeable could answer that question.

galpeter:
Question about remove or update the merge.py was my next item :). So should I remove the merge.py changes from this file and create another PR which removes the merge.py file?

LaszloLango:
@galpeter, yes. It is unnecessary to change anything in 'merge.py', because it is obsolete. Just remove it in your next PR.

tools/check_tidy.py
LaszloLango: We should check CMakeLists.txt files if we check *.cmake files too. But allowing .txt files might be too general and could conflict with other things, likes test resources.

galpeter:
@LaszloLango PR is updated.
Changes:
* CMakeLists.txt now added for general style checking.
* Removed changes in the merge.py.

LaszloLango:
@galpeter, thanks for the update. LGTM. BTW the changes in merge.py are still there. :) But it does not matter, because we'll remove it anyway.

@yichoi, please check the changes since your last review.

galpeter:
@LaszloLango strange... I removed them. Just a sec I'll fix it.

galpeter:
@LaszloLango done. Should be ok now,

yichoi:
still LGTM

PR #604, opened 2017-01-23
Armhf added for Raspbian Jessie

pmarcinkiew:
Please see issue #601 

This commit contains temporary changes for successful build with
Raspbian Jessie used in Noobs 1.9.x.

The compilation was tested with IoT.js build scripts:
./tools/build.py  --target-board=rpi2 --target-arch=armhf \
--no-parallel-build

The no-parallel-build flags is mandatory to prevent RAM
depletion by many concurrent executions of gcc.


pmarcinkiew:
I have to wait for pull request for libtuv because there is dependency.

PR #605, opened 2017-01-23
Remove obsolete merge.py tool

galpeter:


yichoi:
LGTM. BTW, This patch is supposed to be added in #603

PR #607, opened 2017-01-25
Replace process.JSONParse with JSON.parse

galpeter:
process.JSONParse is now obsolete as JerryScript has full JSON support.
We can simply use JSON.parse instead.

### DIFF
--- a/src/js/iotjs.js
+++ b/src/js/iotjs.js
@@ -218,12 +218,6 @@
     fn(this.exports, Native.require, this);
   };
 
-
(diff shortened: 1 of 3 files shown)


glistening:
Could you also replace process.JSONParse in ./test/run_pass/require1/test_require.js?

    var json = process.JSONParse(str);

galpeter:
@glistening oh.. there was another place. Should that be also executed by the tests/travis? Anyway I'll update the PR.

galpeter:
PR updated

PR #612, opened 2017-02-06
Implement Buffer - writeUInt, readUInt

hs0225:
Implement Buffer API
 - new Buffer(array)
 - new Buffer(string, encoding: hex)
 - writeUInt8, writeUInt16LE, writeUInt32LE
 - readUInt8, readInt8, readUInt16LE

### DIFF
--- a/src/js/buffer.js
+++ b/src/js/buffer.js
@@ -18,11 +18,29 @@ var bufferBuiltin = process.binding(process.binding.buffer);
 var util = require('util');
 
 
+function checkInt(buffer, value, offset, ext, max, min) {
+  if (!(buffer instanceof Buffer))
+    throw new TypeError('buffer must be a Buffer instance');
+  if (value > max || value < min)
+    throw new TypeError('value is out of bounds');
+  if (offset + ext > buffer.length)
+    throw new RangeError('index out of range');
+}
+
+
+function checkOffset(offset, ext, length) {
+  if (offset + ext > length)
+    throw new RangeError('index out of range');
+}
+
+
 // Buffer constructor
 // [1] new Buffer(size)
 // [2] new Buffer(buffer)
 // [3] new Buffer(string)
-function Buffer(subject) {
+// [4] new Buffer(string, encoding)
+// [5] new Buffer(array)
+function Buffer(subject, encoding) {
   if (!util.isBuffer(this)) {
     return new Buffer(subject);
   }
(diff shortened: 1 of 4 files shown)


src/js/buffer.js
```diff
@@ -18,20 +18,38 @@ var bufferBuiltin = process.binding(process.binding.buffer);
...
   if (util.isNumber(subject)) {
     this.length = subject > 0 ? subject >>> 0 : 0;
   } else if (util.isString(subject)) {
-    this.length = Buffer.byteLength(subject);
-  } else if (util.isBuffer(subject)) {
+    this.length = Buffer.byteLength(subject, encoding);
+  } else if (util.isBuffer(subject) || util.isArray(subject)) {
     this.length = subject.length;
   } else {
     throw new TypeError('Bad arguments: Buffer(string|number|Buffer)');
```
LaszloLango: Error message should be updated too.

src/module/iotjs_module_buffer.c
```diff
@@ -119,6 +119,33 @@ static size_t bound_range(int index, size_t low, size_t upper) {
...
 
 
+static unsigned hex2bin(char c) {
+  if (c >= '0' && c <= '9')
+    return c - '0';
+  if (c >= 'A' && c <= 'F')
+    return 10 + (c - 'A');
+  if (c >= 'a' && c <= 'f')
+    return 10 + (c - 'a');
+  return (unsigned)(-1);
```
LaszloLango: I'd check the invalid input with a debug assert and return 0 in release.

src/module/iotjs_module_buffer.c
```diff
@@ -119,6 +119,33 @@ static size_t bound_range(int index, size_t low, size_t upper) {
...
+
+
+static size_t hex_decode(char* buf, size_t len, const char* src,
+                         const size_t srcLen) {
+  size_t i;
+
+  for (i = 0; i < len && i * 2 + 1 < srcLen; ++i) {
+    unsigned a = hex2bin(src[i * 2 + 0]);
+    unsigned b = hex2bin(src[i * 2 + 1]);
+    if (!~a || !~b)
```
LaszloLango: Not sure that is a safe way to check it. Might be worth to change the return value of `hex2bin` to signed int.

src/module/iotjs_module_buffer.c
```diff
@@ -119,6 +119,33 @@ static size_t bound_range(int index, size_t low, size_t upper) {
...
+      return i;
```
LaszloLango: This will hide the conversion errors, because it silently returns with a substring.

src/js/buffer.js
hs0225: I updated error message.

src/module/iotjs_module_buffer.c
hs0225: I added a debug assert. 
And `0` means success. I think, `-1` is suitable in this point.

src/module/iotjs_module_buffer.c
hs0225: I fixed the error handling. If input value is invalid hex string, `Buffer` constructor will throw a `TypeError`.

src/module/iotjs_module_buffer.c
hs0225: I don't change the return value of `hex2bin`, so no problem.

PR #626, opened 2017-02-14
Add support for Tizen on ARTIK10

kolipka:
+ Documentation how to build and test


cmake/config/arm-tizen.cmake
```diff
@@ -0,0 +1,22 @@
+# Copyright 2017 Samsung Electronics Co., Ltd. and other contributors
```
glistening: 2017-present

kolipka: fixed

yichoi:
LGTM

PR #627, opened 2017-02-15
Add linting for Javascript files

kolipka:
lint should be run before each commit.


.eslintrc.js
```diff
@@ -0,0 +1,36 @@
+/* global module */
+/*eslint quotes: [2, "double"]*/
```
yichoi: please add space. It should be ``/* eslint qutes: [2, "double"] */``

yichoi:
travis failed because of tidy error for js files.
I think js file should be fixed first along with js coding convention before this patch landed

nova0821:
I think this should be included in travis test. 

Also, We have the rules for JS code in https://github.com/Samsung/iotjs/wiki/Coding-Style-Guideline#coding-style-guideline-for-javascript.

Does it follow these rules?

kolipka:
@yichoi it was tidy check. I added ESLint config file to ignored, added empty line at the end of files and keep 80 chars line limit.
I would like to add this first, and if we agree on rules at overall, I will make next PR fixing all errors and adding this test to travis.

About all rules that are enabled you can read here: http://eslint.org/docs/rules/
in .eslintrc.js you can check rules that I've changed

@nova0821 I think it should follow our rules, but I will double check.

tools/lint.sh
```diff
@@ -0,0 +1,34 @@
...
+TEST='which npm'
+if [ "$?" -ne  0 ] ; then
+  echo "NPM is required in order to run ESLint!\nPlease install it."
+  exit 2
+fi
+
+
+if ! [ -x "$IOTPATH/node_modules/.bin/eslint" ] ; then
+  echo "ESLint not installed!\nYou can install it using:"
+  echo "$IOTPATH$$npm install"
```
hs0225: To run the command `npm install`, please create `package.json` file.

yichoi:
Ok. please check above comment of mine.

kolipka:
Travis failed again - this time because 80 chars limit was exceeded after eslint did automated fixes. 
Are we sure we want to keep such limit? I would suggest to extend it at least to 120 chars per line.

yichoi:
@kolipka I will not extend to 120 per line since I respect what the committer or contributor have been here even it is not his/her favorites. Please keep following the [coding rules](https://github.com/Samsung/iotjs/wiki/Coding-Style-Guideline#coding-style-guideline-for-javascript) @nova0821 mentioned before.

src/js/ble_hci_socket_gap.js
```diff
@@ -41,7 +41,7 @@ var util = require('util');
 
 var Hci = require('ble_hci_socket_hci');
 
-var isLinux = (process.platform === 'linux');
+// TODO: var isLinux = (process.platform === 'linux');
```
LaszloLango: What is the TODO here? Please be more specific.

kolipka: isLinux is unused - I could delete it, but below are commented out lines that use it. Unfortunatelly whole BLE looks like WIP and in my personal opinion should not be commited. I want to avoid doing harm removing something useful for its author, that is why I just commented it out, with TODO marker.

kolipka:
I started conversation in https://github.com/Samsung/iotjs/issues/631 about js code style

kolipka:
This PR is missleading. After #631 and #639 will be settled, we can fix errors and add it to Travis.

PR #633, opened 2017-02-16
Explicit cross compilation for atrik10/tizen. Platfom module building fixed.

tadziopazur:
This is the first step to get the iotjs fully functional on artik. One issue fixed here is the problem of building the iotjs itself (mainly build script fixes).
The other major fix here is the inclusion of platform modules (src/platform/arm-linux/iotjs_module_<module name>-arm-linux.c) in absence of board-specific module.
For arm-linux this is the only way, as there are no non-generic (board-specific) modules.

The change was verified both for host build (x86_64) and artik10/tizen.
0. Get the artik10 with linux installed
1. Install a cross toolchain (arm-linux-gnueabi) and matching sysroot
2. Ensure the toolchain is on the path (PATH=$PATH:path-to-cross-gcc>)
3. Build iotjs:
 $ tools/build.py --target-arch=arm --target-board=artik10 --target-os=tizen

Toolchain: gcc-linaro-4.9.4-2017.01-x86_64_arm-linux-gnueabi.tar.xz
Sysroot: sysroot-eglibc-linaro-2017.01-arm-linux-gnueabi.tar.xz

The commit message:

Build flags extended for linux, darwin x86, x86_64 and arm
  -mthumb added as default flag for arm
  artik10 board added
  tizen target os added
Build fix: arm-linux platform provides armv7l arch, while
  cmake/config.cmake only allows "arm"
Linux platform modules (gpio, i2c, ...) include generic system platform modules if board-specific are not provided
Build fix: explicit builddir (-B) nad home (-H) cmake invocation

Explicit cross compilation support:
  --cross-compile and --sysroot added
  --nuttx_home replaced with --sysroot
  honor CMAKE_C_COMPILER in compiler setup (cmake/config/arm-tizen.cmake)


nova0821:
@tadziopazur Check travis test result.

yichoi:

tadziopazur:
Rebased to master. Fixed travis complaints.
Fixed precommit.py outcome

artik10 not yet added to precommit.py, as it requires installing another toolchain. Subsequent patch

yichoi:
Please rebase it to the master. again :)

cmake/config.cmake
```diff
@@ -14,6 +14,24 @@
 
 cmake_minimum_required(VERSION 2.8)
 
+macro(setIfEmpty varName varValue)
```
galpeter: I don't see this macro used at all.

tools/build.py
```diff
@@ -93,11 +93,17 @@ def init_options():
...
-        choices=['none', 'stm32f4dis', 'rpi2'], default='none',
+        choices=['none', 'artik10', 'stm32f4dis', 'rpi2'], default='none',
         help='Specify the targeted board (if needed): '
              '%(choices)s (default: %(default)s)')
     parser.add_argument('--nuttx-home', default=None,
-        help='Specify the NuttX base directory (required for NuttiX build)')
+        help='Specify the NuttX base directory (required for NuttX build)')
+
+    parser.add_argument('--cross-compile', dest='cross_compile',
+        action='store', help='Specify the croll compilation toolkit prefix. ')
```
galpeter: Did you wanted to write `croll` in the help message?

tools/build.py
```diff
@@ -190,19 +196,22 @@ def adjust_options(options):
...
+                options.sysroot = fs.abspath(options.sysroot)
+            if not fs.exists(options.sysroot):
+                ex.fail('Nuttx sysroot %s does not exist' % options.sysroot)
     if options.target_arch == 'x86':
         options.target_arch = 'i686'
     if options.target_arch == 'x64':
         options.target_arch = 'x86_64'
 
-    if options.target_board == 'rpi2':
+    if options.target_board == 'rpi2' or options.target_board == 'artik10':
```
galpeter: IMHO a bit better would be this:
```python
if options.target_board in ['rpi2', 'artik10']:
   ...
```

tools/build.py
```diff
@@ -190,19 +196,22 @@ def adjust_options(options):
...
     if options.target_os == 'nuttx':
         options.buildlib = True
-        if not options.nuttx_home:
-            ex.fail('--nuttx-home needed for nuttx target')
-
-        options.nuttx_home = fs.abspath(options.nuttx_home)
-        if not fs.exists(options.nuttx_home):
-            ex.fail('--nuttx-home %s does not exists' % options.nuttx_home)
-
+        if not options.nuttx_home and not options.sysroot:
```
galpeter: If we want to keep the `--nuttx-home` option, but simplify this part we could specify the argument's target variable to be same as the `--sysroot` argument so then we would only have the `options.sysroot`.

tools/build.py
tadziopazur: Thanks, fixed.

tools/build.py
tadziopazur: Done.

tools/build.py
tadziopazur: I'll follow the suggestion.

cmake/config.cmake
tadziopazur: Removed.

tools/build.py
```diff
@@ -293,11 +299,10 @@ def build_cmake_args(options, for_jerry=False):
 
     # external include dir
     include_dirs = []
-    if options.target_os == 'nuttx':
-        include_dirs.append('%s/include' % options.nuttx_home)
+    if options.target_os == 'nuttx' and options.sysroot:
```
galpeter: Just noticed: at this point the `options.sysroot` will always have a path value the `adjust_options` will do an error if there is any problem. So we can remove the check from here.

tadziopazur: Yes, this is taken care of by the adjust_options function.
Yet this function (build_cmake_args) is unrelated to adjust_options, and shouldn't depend on the latter having been called before.

I could think of putting an assert equivalent here, but that would not be in line with the rest of the file.

yichoi:
@galpeter any other comments ?

yichoi:
LGTM

PR #635, opened 2017-02-16
Initial implementation of libcoap module

pmarcinkiew:
1. The libcoap submodule was introduced.
2. The build scripts were extend to compile static libcoap library.
3. IoT.js skeleton native module for coap was introduced.
4. IoT.js skeleton JS wrapper for coap was introduced.
5. Test with server listening for connections was implemented.

CoAP was tested and compiled at Linux PC

Compilation for IoT.js with coap is possible, if you will add it as included module in build.config.

You need full PC libcoap build to compile CoAP client example:
```
$ apt-get update && apt-get install -y cmake build-essential python \
git vim autoconf automake pkg-config libtool asciidoc
 $ git clone https://github.com/obgm/libcoap.git
 $ cd libcoap
 $ ./autogen.sh
 $ ./configure
 $ make
```
Test of IoT.js:
```
 $ iotjs/build/x86_64-linux/debug/iotjs/iotjs iotjs/test/run_pass/test_coap.js
```
It listens for client connection:
```
 $ libcoap/examples/coap-client coap://127.0.0.1:6666
v:1 t:CON c:GET i:3cbf {} [ ]
This is a test server made with libcoap (see https://libcoap.net)
Copyright (C) 2017 Samsung Electronics
```

IoT.js-DCO-1.0-Signed-off-by: Piotr Marcinkiewicz <p.marcinkiew@samsung.com>

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -7,3 +7,6 @@
 [submodule "deps/libtuv"]
 	path = deps/libtuv
 	url = https://github.com/Samsung/libtuv.git
+[submodule "deps/libcoap"]
+	path = deps/libcoap
+	url = https://github.com/obgm/libcoap.git
--- a/cmake/config.cmake
+++ b/cmake/config.cmake
@@ -36,6 +36,7 @@ set(CFLAGS_COMMON "${CFLAGS_COMMON} -Wall -Wextra -Werror")
 set(CFLAGS_COMMON "${CFLAGS_COMMON} -Wno-unused-parameter")
 set(CFLAGS_COMMON "${CFLAGS_COMMON} -Wno-unused-variable")
 set(CFLAGS_COMMON "${CFLAGS_COMMON} -Wno-missing-field-initializers")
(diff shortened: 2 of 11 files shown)


pmarcinkiew:
https://en.wikipedia.org/wiki/Constrained_Application_Protocol

pmarcinkiew:
I would like to implement this programming interface:
https://github.com/mcollina/node-coap

yichoi:
I am not sure this is common module for Linux/NuttX.  this will cause binary size increased. How about add it later?

kolipka:
I think we could make it a module that is turned off by default. It could be on for large boards, like Artik. 
@pmarcinkiew Could you check how much binary size increased when you add this module?

pmarcinkiew:
I disabled coap module in default build so it will no incrase binary size for NuttX. It works very well at Linux and Artik so in my opinion we can leave it as optional.

yichoi:
I understand however I prefer to issue clean up for 1.0 Release. Let me think.

pmarcinkiew:
Let's postpone CoAP implementation until IoT.js will be more modular.

Issue #479, opened 2016-10-26
iotjs_string_t should be able to handle strings with '\0' inside


yichoi:
It has been only partially supported.

glistening:
`iotjs_string_t` is simply a pair of buffer and size. It provides very simple functions (i.e., create, destroy, empty, getting data/size, append), which internally use `memcpy`, `realloc`, and don't try to interpret the internal buffer one by one as character. So `iotjs_string_t` is able to handle strings with '\0' inside. I believe we can close this issue.

PR #636, opened 2017-02-17
Remove capacity attribute of iotjs_string_t.

LaszloLango:
This patch simplifies the strings in iotjs a bit and save 4 byte for each string.

### DIFF
--- a/src/iotjs_string.c
+++ b/src/iotjs_string.c
@@ -21,42 +21,43 @@
 #include <string.h>
 
 
-static char iotjs_empty_string[] = { '\0' };
+iotjs_string_t iotjs_string_create() {
+  iotjs_string_t str;
+  IOTJS_VALIDATED_STRUCT_CONSTRUCTOR(iotjs_string_t, &str);
 
+  _this->size = 0;
+  _this->data = NULL;
 
-iotjs_string_t iotjs_string_create(const char* data) {
-  return iotjs_string_create_with_size(data, strlen(data));
+  return str;
 }
 
 
-iotjs_string_t iotjs_string_create_with_size(const char* data, unsigned size) {
+iotjs_string_t iotjs_string_create_with_size(const char* data, uint32_t size) {
   iotjs_string_t str;
   IOTJS_VALIDATED_STRUCT_CONSTRUCTOR(iotjs_string_t, &str);
 
   IOTJS_ASSERT(data != NULL);
-  IOTJS_ASSERT(strnlen(data, size) == size); // forbidden due to ambiguity
 
-  _this->size = _this->cap = size;
+  _this->size = size;
 
   if (size > 0) {
-    _this->data = iotjs_buffer_allocate(size + 1);
+    _this->data = iotjs_buffer_allocate(size);
     memcpy(_this->data, data, size);
-    _this->data[size] = '\0';
   } else {
-    _this->data = iotjs_empty_string;
+    _this->data = NULL;
   }
 
   return str;
 }
 
 
-iotjs_string_t iotjs_string_create_with_buffer(char* buffer, unsigned size) {
+iotjs_string_t iotjs_string_create_with_buffer(char* buffer, uint32_t size) {
   iotjs_string_t str;
   IOTJS_VALIDATED_STRUCT_CONSTRUCTOR(iotjs_string_t, &str);
 
   IOTJS_ASSERT(buffer != NULL);
 
-  _this->size = _this->cap = size;
+  _this->size = size;
   _this->data = buffer;
 
   return str;
@@ -66,25 +67,10 @@ iotjs_string_t iotjs_string_create_with_buffer(char* buffer, unsigned size) {
 void iotjs_string_destroy(iotjs_string_t* str) {
   IOTJS_VALIDATED_STRUCT_DESTRUCTOR(iotjs_string_t, str);
 
-  if (_this->data != iotjs_empty_string) {
+  if (_this->data != NULL) {
     iotjs_buffer_release(_this->data);
+    _this->size = 0;
   }
-}
-
-
-void iotjs_string_reserve(iotjs_string_t* str, unsigned capacity) {
-  IOTJS_VALIDATED_STRUCT_METHOD(iotjs_string_t, str);
-
-  if (_this->cap >= capacity)
-    return;
-
-  if (_this->cap == 0) {
-    IOTJS_ASSERT(_this->data == iotjs_empty_string);
-    _this->data = iotjs_buffer_allocate(capacity + 1);
-  } else {
-    _this->data = iotjs_buffer_reallocate(_this->data, capacity + 1);
-  }
-  _this->cap = capacity;
 }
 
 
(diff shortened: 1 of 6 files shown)


src/iotjs_string.c
```diff
@@ -98,57 +78,45 @@ bool iotjs_string_is_empty(const iotjs_string_t* str) {
...
-    IOTJS_ASSERT(_this->data == iotjs_empty_string);
-    _this->cap = _this->size = size;
-    _this->data = iotjs_buffer_allocate(_this->cap + 1);
-    strncpy(_this->data, data, _this->size);
-  } else if (_this->cap >= _this->size + size) {
-    // Have enough capacity to append data.
-    strncpy(_this->data + _this->size, data, size);
-    _this->size += size;
+  if (_this->data != NULL) {
+    _this->data = iotjs_buffer_reallocate(_this->data, _this->size + size);
```
pmarcinkiew: Are you aware that memory allocation can be as slow as garbadge colector? This code will incrase number of allocator calls. Let's assume that you string with 100 letters will be constructed from 10 words with 9 letters and space. Your code uses 10 allocations but old code used to call malloc 5 times. It is very significant difference.

Please consider keeping capacity. Why don't we store size and capacity in 2 x 16 bits. We will support strings with maximum size 65535 but do we need longer strings in IoT solution? 

I created PR for this change here. Please review it:
https://github.com/Samsung/iotjs/pull/638

yichoi:
Please take into account #479 either

LaszloLango:
@pmarcinkiew IoT.js must handle unicode strings, so the they can contain '\0' characters. Only the size must define the end of the string. There is no need for extra byte. We should remove all of the `strlen` calls, because of this in the near future. As @yichoi said, we must take into account #479 and fix unicode string handling.

pmarcinkiew:
I modified old string to use memcpy instead of strcpy and tests are passing.

I created draft test test_fs_readfileSync_null_string.js but I can't find a way to verify that both input and output files are identical nor the way to check what is inside data buffer.

The console.log prints just characters to first null but data.length is 8 just like size of test file.

Do you have any idea how to improve my test?

pmarcinkiewicz:
I agree that strlen is not good idea for Unicode strings so let's remove strings defined with '\0' terminator. I will drop my pull request. LGTM

src/module/iotjs_module_process.c
```diff
@@ -39,7 +39,6 @@ static iotjs_jval_t WrapEval(const char* source, size_t length, bool* throws) {
   int len1 = strlen(wrapper[1]);
 
   iotjs_string_t code = iotjs_string_create("");
-  iotjs_string_reserve(&code, len0 + length + len1);
```
zherczeg: It wouldn't be a bad idea if we could create a string with a specific size, and use some iotjs_string_rewrite function which can overwrite part of a string. So we would still have a single memory allocation in such cases.

src/module/iotjs_module_buffer.c
```diff
@@ -459,7 +459,7 @@ JHANDLER_FUNCTION(ToHexString) {
   const char* data = iotjs_bufferwrap_buffer(buffer_wrap);
 
   iotjs_string_t str = iotjs_string_create("");
-  iotjs_string_reserve(&str, length * 2);
+
   for (int i = 0; i < length; i++) {
     iotjs_string_append(&str, &"0123456789abcdef"[data[i] >> 4 & 0xF], 1);
```
zherczeg: Could do that here as well.

src/module/iotjs_module_process.c
LaszloLango: I had the same idea, but I was not sure it must be part of this patch or not. I'll do it in this PR then.

kolipka:
@LaszloLango If string must handle unicode chars, we need to remove iotjs_string_create(const char* data) method, since it uses strlen for size checking, and iotjs_string_append should not accept negative size for the same reason. It will work only with null terminated utf8.

I agree with @zherczeg. If we remove capacity, we should be able to create empty string with required size. It would require also function to replace part of the string..... to emulate capacity+append tandem :)

Last but not least: iotjs_string should be designed again it we want to support strings containing \0 (no strlen allowed anywhere) and we should stop using int because its size is platform dependant. 

IMHO our iotjs string should store utf8 string only for simplicity. It should provide functions: from/to ascii/utf16/utf32. Current API is missleading and inconsistant

LaszloLango:
@kolipka, thanks for your suggestions. Looks like it lead us too far from the original purpose of this PR. :)
I agree with you and @zherczeg. I'll rework the current string implementation and update this PR. (IMHO it is unnecessary to separate the work into multiple PRs)

kolipka:
@LaszloLango If you add to PR description:
"... at the cost of runtime performance penalty for string concatenation via iotjs_string_append" I'm OK ;)

yichoi:
@zherczeg any further comments ?

LaszloLango:
@yichoi, I've been working on the update of the PR. Sorry for the delay.

yichoi:
no problem

LaszloLango:
@zherczeg, @yichoi, I've updated the PR. Please check.

src/iotjs_string.c
```diff
@@ -98,57 +84,41 @@ bool iotjs_string_is_empty(const iotjs_string_t* str) {
...
-    // Reallocate buffer and copy data.
-    _this->data = iotjs_buffer_reallocate(_this->data, _this->cap + 1);
-    strncpy(_this->data + _this->size, data, size);
-    _this->size += size;
+    IOTJS_ASSERT(_this->size == 0);
+    _this->data = iotjs_buffer_allocate(size);
   }
-  _this->data[_this->size] = 0;
+
+  strncpy(_this->data + _this->size, data, size);
```
glistening: I expected memcpy. If it is your intension, please could you add comment why using strncpy instead of memcpy?

src/module/iotjs_module_buffer.c
```diff
@@ -458,12 +458,16 @@ JHANDLER_FUNCTION(ToHexString) {
...
 
-  iotjs_string_t str = iotjs_string_create("");
-  iotjs_string_reserve(&str, length * 2);
+  char* buffer = iotjs_buffer_allocate(length * 2);
+  iotjs_string_t str = iotjs_string_create_with_buffer(buffer, length * 2);
+
   for (int i = 0; i < length; i++) {
-    iotjs_string_append(&str, &"0123456789abcdef"[data[i] >> 4 & 0xF], 1);
-    iotjs_string_append(&str, &"0123456789abcdef"[data[i] >> 0 & 0xF], 1);
+    memcpy(buffer, &"0123456789abcdef"[data[i] >> 4 & 0xF], 1);
```
glistening: Could you kindly explain why do you use memcpy instead of iotjs_string_append?

src/module/iotjs_module_process.c
```diff
@@ -38,17 +38,15 @@ static iotjs_jval_t WrapEval(const char* source, size_t length, bool* throws) {
...
   int len1 = strlen(wrapper[1]);
 
-  iotjs_string_t code = iotjs_string_create("");
-  iotjs_string_reserve(&code, len0 + length + len1);
-  iotjs_string_append(&code, wrapper[0], len0);
-  iotjs_string_append(&code, source, length);
-  iotjs_string_append(&code, wrapper[1], len1);
+  uint32_t buffer_length = len0 + len1 + length;
+  char* buffer = iotjs_buffer_allocate(buffer_length);
+  memcpy(buffer, wrapper[0], len0);
```
glistening: Ditto.

src/module/iotjs_module_buffer.c
LaszloLango: @kolipka had some worries about the performance of realloc calls, so it is better to avoid using iotjs_string_append. In this case we know the size we need.

src/iotjs_string.c
LaszloLango: The old code used `strncpy` too. I think it is safer to use `strncpy`, because it adds zero padding if the size is bigger.

glistening: As @yichoi commented, we must consider string that contains '\0' inside. As far as I know, strncpy stops copying before size when it encounter '\0'.

```
$ cat h.cc

#include <stdio.h>
#include <string.h>

int main(void)
{
  char dst[64] = { 0, };
  char src[7] = { 'H', 'e', 'l', 0, 'l', 'o', 0 };

  strncpy(dst, src, 7);
  for (int i = 0; i < 7; i++)
    printf("dst[%d] = %d\n", i, dst[i]);

  return 0;
}

$ g++ -o h h.cc
$ ./h
dst[0] = 72
dst[1] = 101
dst[2] = 108
dst[3] = 0
dst[4] = 0
dst[5] = 0
dst[6] = 0
```

LaszloLango: @glistening, you are right. I'll update the PR.

LaszloLango:
@glistening, I've update the PR. I could not add a JS test to string append. Any idea? Is it possible to test it in the current JS test environment?

src/module/iotjs_module_buffer.c
```diff
@@ -458,12 +458,16 @@ JHANDLER_FUNCTION(ToHexString) {
...
+    buffer++;
+    memcpy(buffer, &"0123456789abcdef"[data[i] >> 0 & 0xF], 1);
```
zherczeg: For me this int->hex conversion is still ugly, but perhaps it could be done in another patch,

PR #638, opened 2017-02-17
IoT.js string size optimization 32 -> 16

pmarcinkiew:
1. Capacity and size fields of IoT.js string were changed from unsigned int to unsigned short int.
2. Two bytes was saved in structure size.
3. Additional asserts were introduced to check for max size.

Verification:
 $ tools/precommit.py

All automated tests at 64-bits Linux PC are passing.


yichoi:
Please take into account #479 either

pmarcinkiew:
I used memcpy instead of strncpy and created draft test with read of file, which contains '\0'. My test can create exact copy of this file in \tmp folder, but I don't know how to validate that is it identical inside JavaScript.

LaszloLango:
There are some cases where you must handle bigger strings than 65535,  e.g.: SunSpider benchmark. IMHO this should be an optional restriction.

LaszloLango:
About the test: I'd add only one simple line in the test file (after the copyright header ofc) and paste the same text to the test file and compere with the result of the read function with `===` operator. To test the write function, read again the output file.

kolipka:
@pmarcinkiew 
I have general questions: 
1. why we allocate size / cap +1?
2. why we need cap? We use it only once in our code: in iotjs_module_process: WrapEval. There it can be done using iotjs_string_create_with_size and copying to proper location of string. Maybe new API would be required to do this safely, eg. iotjs_string_replace_at_position(iotjs_string_t* str, offset, data, size)

LaszloLango:
@kolipka, see #636
In short we don't need caption and +1.

pmarcinkiew:
A null terminated strings are obsolete and capacity is not needed.

Issue #631, opened 2017-02-16
Javascript coding convention


kolipka:
I would like to start conversation about JS coding convention. On wiki https://github.com/Samsung/iotjs/wiki/Coding-Style-Guideline#coding-style-guideline-for-javascript there is information about few selected rules from google coding guidelines. This is missleading:
1. Do we use whole google guidelines and selected rules are just exceptions?
  Propably not since selected rules are in line with google.
2. Do we stick only to selected rules?
  If yes, then there is nothing about 80 chars per line limit, but our check_tidy.py enforce this limit.

In my personal opinion we should stick to well know, strict and widely followed in JS community:
https://www.npmjs.com/package/eslint-config-airbnb - most widely used
https://www.npmjs.com/package/eslint-config-standard - second in popularity
https://www.npmjs.com/package/eslint-config-idiomatic - not quite popular, but very well documented in many languages (https://github.com/rwaldron/idiomatic.js/)

This will prevent many errors and enforce good quality code in our repository.

nova0821:
I am not first author of guideline, I understand like below:

1. Developer must follow the guideline which is mentioned in guideline.
2. If there are some ambiguous cases, developer can search in google's rule and recommend to follow it.

@kolipka Can you write some rules which is not general in JS community? or need more additional rules? Give us more detail example not just URL.

kolipka:
OK, this is not a simple request :)

There is a lot of rules. Google does specify only some of them (~55 from more than 250) while others explicite sets them all. Problems I see in google:
allows this: 
 - http://eslint.org/docs/rules/no-cond-assign (if (a=1) - is OK - source of hard to find errors)
 - http://eslint.org/docs/rules/eqeqeq (does not enforce === and !==)
 - var shadowing
and a lot more related eg.: to code style. 

I think we should have more restrictive rules to have better quality of JS code. Since we will use ESLint, it will be enforced automatically.

nova0821:
@kolipka Could you update our rule document(in https://github.com/Samsung/iotjs/wiki/Coding-Style-Guideline) and make PR? (You can find the document file in ``/docs/help``.)

If you want to update JS coding style rule, you need to make more specific document because we already have our rules in there. I think it will be better to discuss with your proposed rules.

After our JS coding style rule is updated, #627 should be updated with new rule.

kolipka:
@nova0821 I would love to edit Wiki, but have no access to do that. Actually I would write there two sentences:
[...]

PR #639, opened 2017-02-20
Changes to Javascript code style

kolipka:
+ ESLint tool for verification

Google standard extended to AirBnB styleguide with one exception about line length.
It has a lot more rules, is more restrictive so unify code way better. Makes code  easier to read and prevent bugs.

Next step is to fix detected bugs and connect style check to travis-ci

### DIFF
--- a/docs/help/Coding-Style-Guideline.md
+++ b/docs/help/Coding-Style-Guideline.md
@@ -8,10 +8,7 @@ Coding Style Guideline
   * Naming
   * Comments
 * [Coding Style Guideline for Javascript](#coding-style-guideline-for-javascript)
-  * Javascript Language Rules
-  * Javascript Style Rules
-    * Naming
-    * Formatting
+  * [How to verify](#how-to-verify)
 * [Coding Style Guideline for Python](#coding-style-guideline-for-python)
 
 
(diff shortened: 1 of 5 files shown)


yichoi:
please clarify and specify the rules not change the rules.

docs/help/Coding-Style-Guideline.md
```diff
@@ -160,47 +157,20 @@ Use either // or /* */ style comments. However, // style is much prefered.
 
 # Coding Style Guideline for Javascript
 
-This coding standard is based on [google javascript coding standard](https://google.github.io/styleguide/javascriptguide.xml)
+The Javascript code follows [AirBnB guideline](https://github.com/airbnb/javascript)
```
yichoi: please clarify and specify the rules not change the rules.

kolipka: All the rules are listed in attached link. With examples, rationale behind it, in many languages. I will not make it better, and copy/pasting it here will just add maintenace cost. I explained more here:
https://github.com/Samsung/iotjs/issues/631#issuecomment-281317920

nova0821: @kolipka It looks URL(https://github.com/airbnb/javascript) which you link has rules for ES6.

kolipka: You are right, however first link in this document points to ES5 rules extract. I will update my commit.

kolipka: OK, I extended description with links to ESLint documentation and rules switches. Please re-check is it sufficient

nova0821:
@kolipka I tried to run ``eslint`` with ``--fix`` option, it still has 2409 problems. Also, it generate new errors like line line length is over than 80 characters. I guess we need to fix them manually if we follow this eslint config, right?

Also, we have many codes from ``node.js`` and it looks them also generate errors with this eslint config. Do we need to fix codes from ``node.js``?

I'm worried that it will be huge task and also there will be mistakes if we need to fix them manually.
How do you think?

kolipka:
@nova0821 If you approve this PR, I will do it by myself and put PR tomorrow EoD :D

To be serious - yes, there is a lot of simple errors, and few bigger. I will do it, but I suggest to put PR with separate commit per js file (or PR per file). It will be easy to review I hope.

nova0821:
@kolipka Need an answer - We have many codes from ``node.js`` and it looks them also generate errors with this eslint config. Do we need to fix codes from ``node.js``? ``ble`` module also from node.js module - ``bleno``. We need to think about it seriously, because we need to maintenance them. It's hard to maintenance if we change them.

Also, we need to test all of them are working fine after you fix all errors from eslint. Are you sure it will be done in tomorrow? I think it's not simple task.

kolipka:
@nova0821 In short: yes, we keep consistant style in our codebase. We exclude only deps from deps folder :)

When I put code in my repository I take full reponsibility about this code, despite it is taken from someone else. With moment it is included in my codebase, it is my code. I fix it to follow my quality standards, because not even single user cares that my product is broken because I copy/pasted low quality code from somewhere - it works or not. 
If I don't want this burden, I use it as a dependency. IMHO this is where bleno should end up (via npm). We don't like it, we contribute to their repository, or keep own fork. We don't care what is inside until it works as we require.

bleno use jshint - is it compatible with google guidelines? If no, we should leave it and disable quality checking? It is designed to work in node.js, not in Jerryscript - what we should do if it use too much memory or cause any problems?

Copy/paste 3 libraries all with differend style to single code base and it will be a horror to maintain anyway.

Are you sure ble works now? From errors I saw during linting, it works by accident if it works. We have no tests to prove it works or not now.

yichoi:
I would like to discuss this issue after 1.0 release

PR #640, opened 2017-02-20
Eliminate the macro generated jump methods

galpeter:
The JHANDLER_FUNCTION macro always created a
simple jump method for the given c method.
These extra jump methods were responsible to
convert the JerryScript arguments into IoT.js
handlers.

These extra methods can be eliminated by creating
a generic dispatch method which can call the target
native method. This is achieved by attaching the target
method as a native handler for the JS method.

### DIFF
--- a/src/iotjs_binding.c
+++ b/src/iotjs_binding.c
@@ -267,11 +267,11 @@ static jerry_value_t iotjs_jval_as_raw(const iotjs_jval_t* jval) {
 
 
 void iotjs_jval_set_method(const iotjs_jval_t* jobj, const char* name,
-                           JHandlerType handler) {
+                           iotjs_native_handler_t handler) {
   const IOTJS_VALIDATED_STRUCT_METHOD(iotjs_jval_t, jobj);
   IOTJS_ASSERT(iotjs_jval_is_object(jobj));
 
-  iotjs_jval_t jfunc = iotjs_jval_create_function(handler);
+  iotjs_jval_t jfunc = iotjs_jval_create_function_with_dispatch(handler);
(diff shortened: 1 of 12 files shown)


src/module/iotjs_module_adc.c
```diff
@@ -234,10 +234,10 @@ JHANDLER_FUNCTION(Unexport) {
 iotjs_jval_t InitAdc() {
   iotjs_jval_t jadc = iotjs_jval_create_object();
 
-  iotjs_jval_set_method(&jadc, "export", Export);
```
yichoi: ``iotjs_jval_set_method`` comes from the C++ to C notations what I did before.

- iotjs: namespace
- jval: return value
- set_method: method or function name.

Do you have any special reason to name it ``iotjs_register_native_handler``

galpeter: No there is none, I'm open for suggestions. Didn't know about the convention.

yichoi: If you don't have any other special reason to name it, I prefer to leave as it is.

galpeter: It's ok for me. Will do the update asap.

galpeter: Just one thing: you mentioned the `jval` in the name as the return value. However the return type of the `iotjs_jval_set_method` is `void`.

yichoi: ah, my fault. ``jval`` means the value which is transferred JS side.

galpeter:
I've updated the pr

yichoi:
LGTM

PR #642, opened 2017-02-21
build.py modified to ignore commented JS during dependency analysis

pmarcinkiew:
1. Build uses regexp to look for require in JS files. It is used for
   automated dependency analysis.
2. Additional regexp was added to remove basic JS comments from
   analysed file.

Verification:
1. Modify existing module to use require for non existing module.
2. Build fails due to missing module.
3. Comment require statement using // and /* in one line.
4. Build is successful.


yichoi:
LGTM

PR #643, opened 2017-02-22
Add missing iotjs_module_uart-i686-linux.c

bzsolt:


bzsolt:
This is a quick fix, but we have to consider to reorganize this part, because these files are mostly identical. (for each module the `arm-linux`, `i686-linux` and `x86_64-linux` variants)

hs0225:
@bzsolt You're right. We're going to reorganize platform folder and remove identical code. 
LGTM

PR #646, opened 2017-02-23
Add address- and undefined behavior sanitizer to Travis CI

bzsolt:
Additional modifications:
  - Introduce a new option (--buildoptions) to precommit.py,
    which allows to customize testing with extra buildoptions.
  - Merge two similar Travis job into one.
  - Allow to run tests on Linux, when the host is x86_64 and the target is i686.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,7 @@ before_install:
   - if [[ "$INSTALL_ARM_DEPS" == "yes" ]]; then tools/apt-get-install-arm.sh; fi
   - if [[ "$INSTALL_NUTTX_DEPS" == "yes" ]]; then tools/apt-get-install-nuttx.sh; fi
   - if [[ "$INSTALL_TIZEN_DEPS" == "yes" ]]; then . tools/apt-get-install-tizen.sh; fi
+  - if [[ "$INSTALL_TRAVIS_I686_DEPS" == "yes" ]]; then tools/apt-get-install-travis-i686.sh; fi
   - tools/apt-get-install-deps.sh
 
 install:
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -7,15 +7,39 @@ sudo: required
...
+      env: OPTS="--test=host --buildtype=debug --buildoptions=--target-arch=i686,--compile-flag=-fsanitize=address,--compile-flag=-fno-omit-frame-pointer,--compile-flag=-fno-common,--jerry-cmake-param=-DJERRY_LIBC=OFF,--jerry-cmake-param=-DFEATURE_SYSTEM_ALLOCATOR=ON,--no-snapshot,--no-check-valgrind" INSTALL_TRAVIS_I686_DEPS=yes ASAN_OPTIONS=detect_stack_use_after_return=1:check_initialization_order=true:strict_init_order=true
+    - compiler: gcc-4.9
+      addons:
+        apt:
+          sources:
+            - ubuntu-toolchain-r-test
+          packages:
+            - gcc-4.9
+            - gcc-4.9-multilib
+      env: OPTS="--test=host --buildtype=debug --buildoptions=--target-arch=i686,--compile-flag=-fsanitize=undefined,--compile-flag=-fno-omit-frame-pointer,--compile-flag=-fno-common,--jerry-cmake-param=-DJERRY_LIBC=OFF,--jerry-cmake-param=-DFEATURE_SYSTEM_ALLOCATOR=ON,--no-snapshot,--no-check-valgrind" INSTALL_TRAVIS_I686_DEPS=yes UBSAN_OPTIONS=print_stacktrace=1
```
LaszloLango: This job is green, so should not be in allowed failures (IMHO)

bzsolt: There are two error after the linking phase, but it seems these errors are suppressed, so that's the reason why I put it to allow_failure section.

.travis.yml
```diff
@@ -7,15 +7,39 @@ sudo: required
...
+  include:
+    - compiler: gcc-4.9
+      addons:
+        apt:
+          sources:
+            - ubuntu-toolchain-r-test
+          packages:
+            - gcc-4.9
+            - gcc-4.9-multilib
+      env: OPTS="--test=host --buildtype=debug --buildoptions=--target-arch=i686,--compile-flag=-fsanitize=address,--compile-flag=-fno-omit-frame-pointer,--compile-flag=-fno-common,--jerry-cmake-param=-DJERRY_LIBC=OFF,--jerry-cmake-param=-DFEATURE_SYSTEM_ALLOCATOR=ON,--no-snapshot,--no-check-valgrind" INSTALL_TRAVIS_I686_DEPS=yes ASAN_OPTIONS=detect_stack_use_after_return=1:check_initialization_order=true:strict_init_order=true
```
tadziopazur: Why do we even need to specify compiler black magic (--compile-flag=)?

.travis.yml
```diff
@@ -7,15 +7,39 @@ sudo: required
...
 
 env:
-  - OPTS="--test=host --buildtype=release"
-  - OPTS="--test=host --buildtype=debug"
+  - OPTS="--test=host"
   - OPTS="--test=rpi2" INSTALL_ARM_DEPS=yes
   - OPTS="--test=nuttx" INSTALL_NUTTX_DEPS=yes
   - OPTS="--test=misc"
+
+matrix:
```
tadziopazur: My take is that this change brings us towards being able to detect build breaks on a single x86_64 machine.
So far so good.

Still, AFAIU the matrix feature is broken on travis (https://github.com/travis-ci/travis-ci/issues/1519), so why pretend we do it here? precommit.py handles some of the option setup, and this change provides another source of compiler flags, cmake defines and such.
Why can't we have it defined what is built/tested in a single place?

.travis.yml
```diff
@@ -7,15 +7,39 @@ sudo: required
...
+  include:
+    - compiler: gcc-4.9
```
tadziopazur: Does this mean it only builds/works with gcc-4.9 or that we just happen to test with this compiler only?

.travis.yml
bzsolt: I don't think the matrix feature is broken on Travis already, and I've to specify the compiler version for these jobs. So do you suggest to place these jobs explicitly into the `precommit.py` like the existing ones?

.travis.yml
bzsolt: The trusty environment on Travis uses gcc-4.8.4, this version is suitable for `-fsanitize=address`, but not for `-fsanitize=undefined`, thus I have to tell to use a certain version of the compiler.

.travis.yml
bzsolt: `--compile-flag=-fsanitize=address`
I think it's obvious.
`--compile-flag=-fno-omit-frame-pointer`
This flag is recommended for sanitizing, to get more verbose stack-traces.
`--compile-flag=-fno-common`
It can help to detect more bugs. (disable to place global variables into a common section)

.travis.yml
kolipka: @bzsolt actually gcc-4.8 is used for all compilation, except arm-linux-gnueabi, where gcc 4.7 is default on trusty :) In this case JerryScript does not compile. Compiler version is really important. Having linux test on 4.9, arm builds on 4.8 could lead to unexpected, hard to find problems.

bzsolt: This is a simple i686 job on a x64 machine, I don't see any relation to the arm builds.

yichoi:

yichoi:
LGTM

PR #649, opened 2017-02-27
Update function-names in js2c.py after cleanup

bzsolt:


galpeter:
I should have checked this before. I also did the same fix :) so informally lgtm from me.

yichoi:
LGTM

PR #653, opened 2017-02-28
Improve buffer documentation.

zherczeg:
This is a suggested format for all API documentation in IoT.js. Please click on the `view` button to see the formatted MD file. I think this proposal is close to the Node JS documentation:

https://nodejs.org/api/buffer.html

What do you think?

After the format is finalized, all documentation (especially for new code) should follow this style.


yichoi:
Mostly looks good. @nova0821 and @hs0225 are preparing the updates of documents meanwhile the style of documents will be updated too. Please share your idea after their commits raised and then we can decide final style.

yichoi:
LGTM

PR #654, opened 2017-02-28
ADC implementation for Linux

pmarcinkiew:
1. Asynchronous ADC read was implemented for Linux devices.
2. ADC test for Linux was introduced.
3. ADC was tested at Artik 10.
4. Synchronous read fails.

Verification.
1. Enable ADC modules in IoT.js.
2. Run test_adc_linux.js at root in Artik 10.

Expected result:

```
sh-3.2# ./iotjs test_adc_linux.js
ADC initialized
test1 start(read async test)
1456
1456
1456
1454
1457
test1 complete
test2 start(read sync test)
read failed
uncaughtException: AssertionError: {}
```
A measurements were obtained from analogue input 0 with voltage 1.3V


src/platform/iotjs_module_adc-linux-general.inl.h
```diff
@@ -1,4 +1,4 @@
-/* Copyright 2016-present Samsung Electronics Co., Ltd. and other contributors
+/* Copyright 2017-present Samsung Electronics Co., Ltd. and other contributors
```
galpeter: Why the change from 2016? This file was created last year.

src/platform/iotjs_module_adc-linux-general.inl.h
```diff
@@ -16,28 +16,95 @@
...
+  } else {
+    req_data->result = kAdcErrRead;
+  }
+}
+
+
+
 void iotjs_adc_export_worker(uv_work_t* work_req) {
-  IOTJS_ASSERT(!"Not implemented");
+  InitializeadcWorker(work_req);
```
galpeter: Do we need the extra method at all? Can't we just move the implementation here?

src/platform/iotjs_module_adc-linux-general.inl.h
```diff
@@ -16,28 +16,95 @@
...
 }
 
 
 void iotjs_adc_read_worker(uv_work_t* work_req) {
-  IOTJS_ASSERT(!"Not implemented");
+  ReadadcWorker(work_req);
```
galpeter: Same as the method above.

src/platform/iotjs_module_adc-linux-general.inl.h
```diff
@@ -16,28 +16,95 @@
...
+  snprintf(path, sizeof(path), ADC_PIN_INTERFACE, pin);
+  FILE* file = fopen(path, "r");
+  if (NULL == f) {
+    return -1;
+  }
+  int32_t value;
+  if (1 == fscanf(f, "%d", &value)) {
+    fclose(file);
+    return value;
+  } else {
```
galpeter: Do we need the else at all? Also there line 99 will never be executed this way.

src/platform/iotjs_module_adc-linux-general.inl.h
```diff
@@ -16,28 +16,95 @@
...
+  ReadadcWorker(work_req);
 }
 
 
 int32_t iotjs_adc_read_sync(int32_t pin) {
-  IOTJS_ASSERT(!"Not implemented");
+  char path[ADC_PATH_BUFFER_SIZE];
+  snprintf(path, sizeof(path), ADC_PIN_INTERFACE, pin);
+  FILE* file = fopen(path, "r");
+  if (NULL == f) {
```
galpeter: Should the `f` be `file`?

src/platform/iotjs_module_adc-linux-general.inl.h
```diff
@@ -16,28 +16,95 @@
...
+    return -1;
+  }
+  int32_t value;
+  if (1 == fscanf(f, "%d", &value)) {
```
galpeter: The `f` should be `file` imho.

src/platform/iotjs_module_adc-linux-general.inl.h
```diff
@@ -16,28 +16,95 @@
...
+    fclose(file);
```
galpeter: As we are closing the file in both cases, maybe a different sequence could be used. Something like this? (pseudo-code-ish):
```
read_count = fscanf(...)
fclose(..)
if (read_count == 1) { return  value }
return -1
```
What do you think?

src/platform/iotjs_module_adc-linux-general.inl.h
pmarcinkiew: Thank you for all these helpful comments.

I would like to keep this ADC implementation interface to be identical like for Nuttx but still keep it similar to GPIO for Linux for further refactoring.

It is clear that ADC, GPIO and PWM for Linux should be implemented 100% in JavaScript using FS API. I love to do it just now, but first I have to test all key APIs at Artik board. It is better to create working API, make tests and next refactor it to make it shine.

yichoi:
Please check the coding style with ``/too/check-tidy.py``

src/platform/iotjs_module_adc-linux-general.inl.h
```diff
@@ -16,28 +16,95 @@
...
   return -1;
 }
 
 
 void iotjs_adc_unexport_worker(uv_work_t* work_req) {
-  IOTJS_ASSERT(!"Not implemented");
+  iotjs_adc_reqwrap_t* req_wrap = iotjs_adc_reqwrap_from_request(work_req);
+  iotjs_adc_reqdata_t* req_data = iotjs_adc_reqwrap_data(req_wrap);
+  DDDLOG("ADC UnexportadcWorker()");
+  req_data->result = kAdcErrOk;
```
hs0225: Is there no work for unexport ADC?

hs0225:
Please integrate `test_adc_linux.js` with `test_adc.js`. You can use it like below.
```javascript
if (process.platform === 'linux') {}
else if (process.platform === 'nuttx') {}
```

src/platform/iotjs_module_adc-linux-general.inl.h
pmarcinkiew: Linux drivers for ADC, GPIO and PWM use sysfs:
https://en.wikipedia.org/wiki/Sysfs
Artik 
https://developer.artik.io/documentation/developer-guide/gpio/kernel-gpio.html

Kernel interface

GPIO:
http://git.kernel.org/cgit/linux/kernel/git/torvalds/linux.git/tree/Documentation/gpio/sysfs.txt
PWM:
http://git.kernel.org/cgit/linux/kernel/git/torvalds/linux.git/tree/Documentation/pwm.txt
http://developer.toradex.com/knowledge-base/pwm-(linux)
ADC:
http://developer.toradex.com/knowledge-base/adc-(linux)
https://github.com/torvalds/linux/tree/master/drivers/iio/adc


GPIO, PWM, ADC is 100% file system I/O - open/close, write/read files. Nuttix uses other approach but at least for Linux these modules should be pure JavaScript using:
https://github.com/Samsung/iotjs/blob/master/docs/api/IoT.js-API-File-System.md

This 100% JS solution will be portable for all Linux devices - PC, RPi2 and Artik 5xx, 7xx, 10xx board and also for OpenWRT. We don't care about CPU architecture in JS so it will be much easier to maintain.

src/platform/iotjs_module_adc-linux-general.inl.h
```diff
@@ -16,28 +16,83 @@
...
+  char buffer[ADC_VALUE_BUFFER_SIZE];
+  if (DeviceOpenReadClose(value_path, buffer, sizeof(buffer))) {
+    req_data->result = kAdcErrOk;
+    req_data->value = atoi(buffer);
+  } else {
+    req_data->result = kAdcErrRead;
+  }
+}
 
 int32_t iotjs_adc_read_sync(int32_t pin) {
```
hs0225: I know `readSync` isn't work. Please add annotation and log for this.

yichoi:
LGTM. @hs0225, please follow up your new API convention with this patch

test/run_pass/test_adc.js
```diff
@@ -16,9 +16,17 @@
 
 var adc = require('adc');
 var assert = require('assert');
-var stm32fPin = require('pin').STM32F4DIS;
+var testPin = null;
```
galpeter: The `testPin` variable is not used at all. Should this be used when creating an adc object?

PR #655, opened 2017-03-01
Implement testrunner in Python

rtakacs:
A new test runner was implemented in Python that could replace the JavaScript based one. All information about the tests are placed into a JSON file. The new test system works similarly to the currently used one, the only exception is that every test is run separately.

The patch has effects on the binary size that can be seen in the following table:

[RPI2 release build]

Sections           | master | testrunner_refactor
-------------------|--------|--------------------
.gnu.hash          |  10632 | 10616
.dynsym            |  21952 | 21888
.dynstr            |  31070 | 31016
.gnu.version       |   2744 | 2736
.text              | 179988 | 179628
.rodata            |  74168 | 73880
.bss               | 265568 | 265560
**Total**              | **590159** | **589361**

### DIFF
--- a/tools/common_py/path.py
+++ b/tools/common_py/path.py
@@ -59,3 +59,6 @@ CHECKTEST_PATH = fs.join(TOOLS_ROOT, 'check_test.js')
 BUILD_CONFIG_PATH = fs.join(PROJECT_ROOT, 'build.config')
 BUILD_MODULE_CONFIG_PATH = fs.join(PROJECT_ROOT, 'build.module')
 BUILD_TARGET_CONFIG_PATH = fs.join(PROJECT_ROOT, 'build.target')
+
+# IoT.js build information.
+BUILD_INFO_PATH = fs.join(TOOLS_ROOT, 'iotjs_build_info.js')
--- a/tools/iotjs_build_info.js
+++ b/tools/iotjs_build_info.js
@@ -1,0 +1,31 @@
+/* Copyright 2017-present Samsung Electronics Co., Ltd. and other contributors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/* Just for the testrunner to get runtime information about the build. */
+var native_modules = Object.keys(process.native_sources)
+var binding_modules = Object.keys(process.binding)
+var builtins = native_modules.concat(binding_modules)
+
+if (process.env.IOTJS_ENV.indexOf("experimental") > -1)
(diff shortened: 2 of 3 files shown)


zherczeg:
Good improvement!

kolipka:
While I agree that test driver should not be part of IoT.JS binary, using python is not a good idea. To run tests on embedded device, I would need python there first? I would love to be able to write:
```bash
iotjs tests/suite.js
```
to execute all tests and get results :)

anyway: lines removed to lines added ratio is awesome :) I like it :)

zherczeg:
There are two key problems with a js testrunner
- JS variables are spilling making other tests flakey
- You cannot test certain exceptions / aborts. The current test system neglects these cases but we need it in the future.

So we have not much choices.

kolipka:
@zherczeg I am not sure I understand:
>  JS variables are spilling making other tests flakey

Following Mocha style, or at least Turbo, we would have nicely separated tests. Would that solve it?

> You cannot test certain exceptions / aborts. The current test system neglects these cases but we need it in the future.

This is about process.exit? This is problematic without child_process indeed. There is python or bash. I can't see good solution :(

My only concern is that It would be nice to run test suite, even in limited scope, on target device without much effort.

nova0821:
@rtakacs @zherczeg Please see https://github.com/Samsung/iotjs/pull/450.

Our testdriver was implemented in phyton, but there is no way to run testdriver without pyton(like nuttx).
That's why we changed testdriver python to javascript.

I also think changing testdriver python to js is not a good idea.

yichoi:
@rtakacs Thank you!

yichoi:
@rtakacs Thanks you!. Impressive. As @nova0821 said, JS environment is preferred.

rtakacs:
@yichoi @nova0821 the primary motivation for the python based solution is that we can validate the output (compare it with a text output) and the modules are restarted after each test, so an incorrect state does not hurt follow-up tests. However if you prefer the JS based solution we can keep it of course.

glistening:
@rtakacs @yichoi give an idea of unifying the test framework. We can use js-remote-test for x86/desktop also, then, we can remove local test driver totally (for either js driver and python driver). Most sharable code in test runner will be shared. I will contact you for details on IRC.

rtakacs:
@glistening @yichoi Nice idea! In this case we should not maintain two testrunner :)

rtakacs:
I think the only missing thing is the JS coverage support. It needs further investigation how to use that with a Python based testrunner.

glistening:
@rtakacs Oh, a good catch. We already have an internal patch used for system i/o coverage although we need test. @daeyeon made the patch. Let's talk by IRC for details.

yichoi:
Overall, we are going to move local test environment to remote test environment.
Please make concrete discussion then close this.

LaszloLango:
IMHO it is worth to merge this before we change to JSRemoteTest, because it will take much time to prepare it to be suitable for this task.

glistening:
@LaszloLango I agree. Let's push PR incrementally.
@rtakacs Please don't remove JS test driver in this PR. Just introduce python driver. And for for integrating with code coverage tool, make another PR.

rtakacs:
@glistening: I've updated the PR based on your comment.

rtakacs:
Need to mention that there is an `iotjs_build_info.js` execution in the testrunner's `__init__` method. This helps to determine which modules are compiled into the binary (to skip test files that want to use a module that is not supported).

yichoi:
LGTM

PR #657, opened 2017-03-01
[Bugfix] PWM/linux: Fix command buffer for enable/disable

tadziopazur:
Sprintf buffer was too short, resulting in only terminating byte fitting


hs0225:
LGTM

PR #662, opened 2017-03-09
Add check_tidy.py option to edit the detected clang format errors.

hs0225:
- Add  `autoedit` option that edits to files instead of displaying a diff.(only clang-format)

### DIFF
--- a/tools/check_tidy.py
+++ b/tools/check_tidy.py
@@ -15,6 +15,8 @@
 # limitations under the License.
 
 from __future__ import print_function
+
+import argparse
 import fileinput
 import functools
 import os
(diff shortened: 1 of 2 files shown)


tools/check_tidy.py
```diff
@@ -28,6 +30,16 @@
 from common_py.system.executor import Executor as ex
 
 
+def parse_option():
+    parser = argparse.ArgumentParser()
+    parser.add_argument('--autoedit', action='store_true', default=False,
+        help='edits to files instead of displaying a diff' +
```
galpeter: There is no need for the `+` sign. The concatenation will also happen without it (strings are automatically concatenated).

galpeter: Maybe a bit better help message would be: `Automatically edit/fix the detected clang format problems. No diff(s) will be created/displayed.`

hs0225: @galpeter Done. Thanks for your review.

yichoi:
LGTM

PR #663, opened 2017-03-09
Allow js2c to be invoked from cmdline

galpeter:
By allowing the js2c to be invoked from the cmdline
the testing of the functionality becomes easier. Also
this gives us opportunities to improve the build system.


yichoi:
LGTM

PR #664, opened 2017-03-09
Remove duplicate tizen entires from build.config

galpeter:
After a previous commit the tizen entires in the
build config file were duplicated. As the config
file is a json there is no reason for this duplication.


yichoi:
LGTM

PR #667, opened 2017-03-14
Enable unused variable check

bzsolt:

### DIFF
--- a/src/iotjs_binding.c
+++ b/src/iotjs_binding.c
@@ -167,7 +167,7 @@ void iotjs_jval_destroy(iotjs_jval_t* jval) {
 
 
 static void iotjs_jval_destroy_norelease(iotjs_jval_t* jval) {
-  IOTJS_VALIDATED_STRUCT_DESTRUCTOR(iotjs_jval_t, jval);
+  IOTJS_VALIDATABLE_STRUCT_DESTRUCTOR_VALIDATE(iotjs_jval_t, jval);
 }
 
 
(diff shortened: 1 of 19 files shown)


src/iotjs_binding.c
```diff
@@ -167,7 +167,7 @@ void iotjs_jval_destroy(iotjs_jval_t* jval) {
 
 
 static void iotjs_jval_destroy_norelease(iotjs_jval_t* jval) {
-  IOTJS_VALIDATED_STRUCT_DESTRUCTOR(iotjs_jval_t, jval);
+  IOTJS_VALIDATED_STRUCT_DESTRUCTOR_VALIDATE(iotjs_jval_t, jval);
```
yichoi: Roughly, naming style in this patch is somewhat awkward.

bzsolt: Just suffixed the current ones, but I'm open for suggestions.

yichoi: I could not pick the best prefix term since I am not native English speaker :)

glistening: Suggested name ``IOTJS_VALIDATED_STRUCT_DESTRUCTOR_VALIDATE gives more information. But it is too long and some word are duplicated.

Basically, I don't like current prefix  ``IOTJS_VALIDATED_STRUCT``:
1) Too long. 
2) Also this structure does not validated, but validatable. But it seems validatable is not popular word. Sanity checkable would be better. But it is too long.

I cannot suggest better candidate that satisfy aboves.
What about landing unused variable warning only? And rename the name in another PR?

yichoi: Let's go with ``IOTJS_VALIDATABLE_STRUCT_DESTRUCTOR_VALIDATE``.
If we find better idea, then rename it in another PR.

bzsolt: Should I apply this form to the `IOTJS_VALIDATED_STRUCT_METHOD` too?

yichoi: yes, please

bzsolt:
I've updated the PR.

yichoi:
LGTM

PR #669, opened 2017-03-14
Refactor testrunner to use JSON as test descriptor.

rtakacs:


tools/test_runner.js
```diff
@@ -80,21 +79,22 @@ Runner.prototype.checkSkipModule = function() {
 
 
 Runner.prototype.run = function() {
-  if (this.attr.skip && (this.attr.skip.indexOf('all') >= 0 ||
-      this.attr.skip.indexOf(this.driver.os) >= 0)) {
+  if (skip = this.test['skip']) {
```
chokobole: I hope to declare `skip`  variable for the function scope, or this `skip`, which someone might use this name on later, let him/her be very hard to seek to find where it is used.

kolipka: 1. skip is undeclared, this is against coding rules
2. assignment in if is a bad practice, please don't do this :)

yichoi:
The benefit of changing to JSON is for maintenance. Actually these JS files are not code but more like data. @kolipka, do you have any requirement to describe function to this file?

kolipka:
@yichoi Currently no. In the future maybe - if we will support running tests on multiple platforms, we would possibly want to skip certain tests, not the whole file. For this purpose I would prefer to tests in test file decide does it need execution (based on some env values), rather than skipping whole file.

This is just a question. Every JSON is valid JS, so switching back cost nothing if needed. There is just one thing to fix in code, and whole change is fine for me.

zherczeg:
The motivation for this is sharing this data with other environments like python. This is needed for automated testing IoT.js on devices.

rtakacs:
@kolipka, @chokobole: Thanks for your notices. As @zherczeg wrote, JSON like syntax is more general, many programming languages can deal with it easily. We have implemented an automatized test environment in Python (for stm32f4 board) and we would like the test information to be in a common format.

rtakacs:
I've updated the PR.

kolipka:
@rtakacs Thanks. If we want to use it in python, advantage of json is clear :)

yichoi:
LGTM

PR #678, opened 2017-03-17
Simplify HTTP parser methods

galpeter:
Removed callback macros and generated methods from HTTP parser module.
The macros were used to generate a simple jump function for the
actual method.


chokobole:
LGTM

PR #682, opened 2017-03-20
Fix .gitignore: ensure that build.config is under revision control

akosthekiss:
Nuttx-related build tasks may create .config files, which as auto-
generated files should be ignored. However, there are other files
with this extension, which are important. Most notably, the master
build.config file is an example. However, even though it is
committed in the repository, it is also on the ignore list. This
patch fixes the .gitignore file to list only what really should be
ignored.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,3 +33,4 @@ TAGS
 
 # config files
 *.config
+!build.config


chokobole:
Then how about marking with exclamation mark like `!build.config`? Is there a problem?

akosthekiss:
@chokobole that can work too. I've updated the PR and adjusted the commit message.

chokobole:
LGTM

yichoi:
LGTM

PR #686, opened 2017-03-21
Remove remnants of C++ (documentation & dependency install scripts)

akosthekiss:
The project is (almost completely) clean C, so this patch removes
references to C++ compilers from documentation and install scripts.

"Almost completely clean C" refers to the nuttx-stm32f4 target,
which still contains a cxx source file and a Makefile that seems to
mandate a C++ compiler. The cleanup of the target directory is left
for follow-up work.


yichoi:
LGTM

PR #687, opened 2017-03-21
Simplify nuttx precommit steps

akosthekiss:
Don't build nuttx twice, especially since the first try is
guaranteed to fail, because the libraries in the `copylib` make
target are not present. So, just perform a `make context` first
that configures nuttx properly by creating some necessary headers
and symlinks, and do the full build `make all` only once.


yichoi:
LGTM

PR #689, opened 2017-03-22
STM32F4Discovery config and nuttx makefile fix

tadziopazur:
Makefile meant for nuttx apps/system/iotjs changed to accept external IOTJS_ROOT_DIR
config file for nuttx provided (targets/stm32f4discovery/nuttx/config_mpu)


targets/nuttx-stm32f4/nuttx/config_mpu
```diff
@@ -0,0 +1,1673 @@
+#
+# Automatically generated file; DO NOT EDIT.
```
akosthekiss: The `targets/nuttx-stm32f4/nuttx` directory already contains various config files following the naming pattern of `.config.*`. It would be nice to keep things consistent, IMHO, which would mean a `.config.mpu` name for this file.

hs0225:
Please update on documents about this PR.

yichoi:
@tadziopazur Do you have any further updates?

targets/nuttx-stm32f4/nuttx/config_mpu
tadziopazur: Done.
I'd say I am not sure starting file names with the dot '.' is a good idea, as they do not show when listing directories with ls (standard practice on unix/OSX).

tadziopazur:
I am not sure multiple config files are an issue: u-boot and linux kernel have hundreds of those, no problem. I prefer a solid copy to guide, because applying the guide is both time-consuming and error-prone.

How many config files we need is another question.
If this approach is not a problem, I'll rebase the config for nuttx 7.19.

targets/nuttx-stm32f4/nuttx/config_mpu
glistening: @tadziopazur I agree with you. The preceding dot makes the file invisible in unix/osx. Could you remove dot in config file? You can change it in this PR or make another PR. In this change, please don't forget update the [guide](https://github.com/Samsung/iotjs/tree/master/targets/nuttx-stm32f4#3-configure-nuttx).

glistening:
I think multiple config file may be OK if the config name is clear and the config contains more description what this config it is.

For this PR's new config, I heard that the only change is mpu option. If so, it would be enough to add mpu configuration in default config. In this case, please update the document in [configuring nuttx](https://github.com/Samsung/iotjs/tree/master/targets/nuttx-stm32f4#3-configure-nuttx).

targets/nuttx-stm32f4/nuttx/config_mpu
akosthekiss: Well, my key point was consistency. If dot-files are disliked, then the existing config files (their names) should be adapted. (I do know `ls`. I even know `ls -a`.)

glistening: @akosthekiss I know your point. You talked about consistency, not dot prefix. My comment was about the dot prefix that @tadziopazur talked. My opinion is that there is no reason to keep the config files invisible by using dot-prefix.

hs0225:
@tadziopazur I have some question. What is the difference `.config.default` and `.config.mpu`?
When do you need this config file?
If you just want to set the mpu options, I would like to create config file based on the `.config.default` file or add this configuration on `.config.default`. I don't know why `.config.mpu` need setting like i2c, pwm and so on.

glistening:
@tadziopazur This pr is not update for several days. If you are busy, we'll follow up it.

yichoi:
@tadziopazur, no further updates?

yichoi:

PR #694, opened 2017-03-24
Extend magic string list with string constants in the native code.

LaszloLango:
There are a lot of strings that are not added to the magic string list. Currently the magic strings are generated by `tools/js2c.py` based only on the JavaScript code. Those properties that are added in native code won't be in the list. For example in `test_httpserver.js` (strings used more than 10 times):

| string | freq. |
|---|---|
| }); | 11 |
| isUndefined | 11 |
| Server | 11 |
| Stream | 11 |
| t | 11 |
| args | 12 |
| ended | 12 |
| HTTPParser | 12 |
| id | 12 |
| inherits | 12 |
| isNumber | 12 |
| Native | 12 |
| read | 12 |
| binding | 13 |
| cb | 13 |
| finish | 13 |
| isBuffer | 13 |
| once | 13 |
| onread | 13 |
| port | 13 |
| _readableState | 13 |
| e | 14 |
| headers | 14 |
| _socketState | 14 |
| _writableState | 14 |
| checkArgFunction | 15 |
| destroy | 15 |
| _handle | 15 |
| handleTimeout | 15 |
| setTimeout | 15 |
| chunk | 16 |
| owner | 16 |
| req | 16 |
| data | 17 |
| error | 17 |
| fs | 17 |
| offset | 17 |
| assert | 18 |
| isString | 18 |
| res | 18 |
| timeout | 18 |
| on | 19 |
| options | 19 |
| parser | 20 |
| checkArgString | 21 |
| end | 22 |
| isFunction | 23 |
| write | 23 |
| fsBuiltin | 25 |
| Buffer | 26 |
| module | 26 |
| self | 26 |
| fd | 27 |
| err | 29 |
| state | 30 |
| close | 31 |
| emit | 34 |
| path | 34 |
| require | 35 |
| stream | 36 |
| _builtin | 41 |
| exports | 41 |
| socket | 51 |
| _buffer | 65 |
| callback | 70 |
| _onNextTick | 76 |
| util | 86 |

Some of these strings are came from the test file itself, but there are some that came from native implementations of the modules, such as `_buffer, _builtin`, etc.

I'd like to suggest that put every string constant into a separate header and add those strings to magic string list. This decrease the memory consumption. I ran a few test with Freya tool of Valgrind and found out that the PR can save approximately a few hundred bytes in peak memory usage and a few kilobyte in the total memory usage. E.g. on `test_httpserver.js`:

**before the patch**
```
    [2] Peak: 6.1Kb (6256b) Allocs: 210 Reallocs: 0 Total: 8.0Kb (8252b) 
    ecma_new_ecma_string_from_utf8 (iotjs/deps/jerry/jerry-core/ecma/base/ecma-helpers-string.c:190)
```

**after the patch**
```
    [2] Peak: 5.8Kb (6010b) Allocs: 134 Reallocs: 0 Total: 6.9Kb (7116b) 
    ecma_new_ecma_string_from_utf8 (iotjs/deps/jerry/jerry-core/ecma/base/ecma-helpers-string.c:190)
```


glistening:
I think it is good. If we choose the magic string selectively, not mandatory. And could you resolve the conflicting files?

yichoi:
LGTM

PR #695, opened 2017-03-27
Add a description of --config

daeyeon:
Problem: There is no brief description of --config.
Change: This commit adds the description.

### DIFF
--- a/tools/build.py
+++ b/tools/build.py
@@ -64,8 +64,7 @@ def init_options():
                     argv.append('--%s=%s' % (opt_key, val))
 
     # Apply command line argument to argv.
-    argv = argv + ([arg for arg in sys.argv[1:]
-                    if not arg.startswith('--config=')])
+    argv = argv + sys.argv[1:]
 
     # Prepare argument parser.
     parser = argparse.ArgumentParser()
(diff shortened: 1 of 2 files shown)


tools/build.py
```diff
@@ -64,8 +64,7 @@ def init_options():
                     argv.append('--%s=%s' % (opt_key, val))
 
     # Apply command line argument to argv.
-    argv = argv + ([arg for arg in sys.argv[1:]
-                    if not arg.startswith('--config=')])
+    argv = argv + sys.argv[1:]
```
glistening: Could you check that it is safe to remove the filtering ? The original author of the line is @chokobole.

tools/build.py
```diff
@@ -82,6 +81,10 @@ def init_options():
     parser.add_argument('--clean', action='store_true', default=False,
         help='Clean build directory before build (default: %(default)s)')
 
+    parser.add_argument('--config', default=path.BUILD_CONFIG_PATH,
+        help='Specify the config file (default: %(default)s)',
+        dest='config_path')
+
```
glistening: It seems almost all arguments in ``tools/build.py`` are matched the members in ``build.config``. It would be consistent to add "config" in "build_option" in ``build.config`. Please put the item in alphabetical order although some items break the rules.

PR #697, opened 2017-03-28
Set Coverity-Scan

daeyeon:
This commit gets Coverity-Scan run whenever master branch is updated.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,16 +17,31 @@ install:
 script: "tools/precommit.py $OPTS"
 
 env:
-  - OPTS="--test=host"
-  - OPTS="--test=rpi2" INSTALL_ARM_DEPS=yes
-  - OPTS="--test=nuttx" INSTALL_NUTTX_DEPS=yes
-  - OPTS="--test=artik10" INSTALL_TIZEN_DEPS=yes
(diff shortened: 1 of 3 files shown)


robertsipka:
I'm not sure that this secure (COVERITY_SCAN_TOKEN) encrypted variable should work. It was encrypted and encoded using the private Project token from Coverity Scan project settings. As I see [here](https://scan.coverity.com/projects/daeyeon-iotjs), it points to your forked repository.
Encryption and decryption keys are tied to the repository. If you fork a project and add it to Travis CI, it will not access to the encrypted variables.

daeyeon:
Thank you for your concern. This is actually a different commit from the one you mentioned with the encrypted value for my repository. This uses the different one which is pointing to [Samsung/iotjs](https://scan.coverity.com/projects/samsung-iotjs) instead.

robertsipka:
Thanks, I can't found that, I can't able to  access this page.  This would be a private measurement site? 
If not, could you please also place the project's Coverity Scan badge in the README file?

daeyeon:
Agree with you. I just pushed another commit including the badge. The page would be viewable for all users.

yichoi:
LGTM

PR #698, opened 2017-03-28
Remove leftover iotjs_module_pin implementations

akosthekiss:
Some platform directories still contain `iotjs_module_pin-*.c`
files that implement a "platform-specific" `iotjs_pin_initialize`
function (but actually platform-independently assert out). More
importantly, the `iotjs_pin_initialize` function seems not to be
called anywhere in the code base. This commit removes this leftover
dead code.


yichoi:
LGTM

PR #700, opened 2017-03-28
Force string conversation when processing the magic strings

galpeter:
Previously the magic strings were only converted to Python strings
just before writing them out into the header. This was modified
to convert the magic strings after reading them from the
snapshot file.

### DIFF
--- a/tools/js2c.py
+++ b/tools/js2c.py
@@ -47,6 +47,13 @@ def remove_whitespaces(code):
     return re.sub('\n+', '\n', re.sub('\n +', '\n', code))
 
 
+def force_str(string):
+    if not isinstance(string, str):
+        return string.decode('utf-8')
+    else:
+        return string
+
+
 def parse_literals(code):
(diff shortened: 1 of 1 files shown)


tools/js2c.py
```diff
@@ -266,10 +273,7 @@ def js2c(buildtype, no_snapshot, js_modules, js_dumper, verbose=False):
 
         sorted_strings = sorted(magic_string_set, key=lambda x: (len(x), x))
         for idx, magic_string in enumerate(sorted_strings):
-            if not isinstance(magic_string, str):
-                magic = magic_string.decode('utf-8')
-            else:
-                magic = magic_string
+            magic = force_str(magic_string)
```
glistening: Could you explain why we need to apply ``force_str`` here? I think ``magic_string`` already has forced strings.

galpeter: Ohhh yes, that is true. We don't need to call the force_str here it's just a leftover from the first version. Will remove it asap.

PR #704, opened 2017-03-29
Fix the close function in gpio, adc and uart

CedricDum:
When the close function is called, that returned an error :
uncaughtException: ReferenceError: value is not defined

The value parameter does not exist in the close function and the close callback is called with only one parameter.

IoT.js-DCO-1.0-Signed-off-by: cedricDum duminy.cedric@gmail.com

### DIFF
--- a/src/js/gpio.js
+++ b/src/js/gpio.js
@@ -161,7 +161,7 @@ GpioPin.prototype.close = function(callback) {
   }
 
   this._binding.close(function(err) {
-    util.isFunction(callback) && callback.call(self, err, value);
+    util.isFunction(callback) && callback.call(self, err);
   });
 
   this._binding = null;
--- a/docs/api/IoT.js-API-GPIO.md
+++ b/docs/api/IoT.js-API-GPIO.md
@@ -166,7 +166,7 @@ console.log('value:', gpio10.readSync());
 
 
 ### `gpiopin.close([callback])` <a name="gpiopin-close"></a>
-* `callback <Function(err: Error | null, value: Boolean)>`
+* `callback <Function(err: Error | null)>`
 
 Closes a GPIO pin asynchronously.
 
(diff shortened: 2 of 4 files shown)


src/js/gpio.js
```diff
@@ -160,7 +160,7 @@ GpioPin.prototype.close = function(callback) {
     throw new Error('GPIO pin is not opened');
   }
 
-  this._binding.close(function(err) {
+  this._binding.close(function(err, value) {
```
galpeter: I'm not sure that this is the correct fix. In case of close I think there is no reason to get back a value (also what does the value contains?) (however I could be wrong). 

imho: the correct fix is to remove the `value` parameter one line below.

src/module/iotjs_module_gpio.c
```diff
@@ -143,6 +143,7 @@ void iotjs_gpio_after_worker(uv_work_t* work_req, int status) {
           iotjs_jargs_append_error(&jargs, "GPIO Close Error");
         } else {
           iotjs_jargs_append_null(&jargs);
+          iotjs_jargs_append_bool(&jargs, req_data->value);
```
galpeter: AFAIK: In case of close the `value` does not hold any meaningful information, as it is not set by any of the platform/board code part.

yichoi:
Please check your DCO. It would be better to make one single commit

glistening:
@CedricDum Thank you for your patch. Could you make your commits into one commit? Currently this PR contains total 3 commits (two commits and one merge commit). Please use git rebase to merge them into one,  and make a force push into your forked repository. Then, the squashed commit will be automatically applied to this PR. Also, could you fix the typo in PR title? Commit title is already fixed, however, the PR title still has the typo _Fixe_.

glistening:
@CedricDum I checked the similar error in other modules and found ``adc.close`` and ``uart.close`` have the same bugs. Could you kindly fix the same error? The API document seems correct. We just need to change `uart.js` and `adc.js` one line respectively.

CedricDum:
@glistening Of course. It's done.

yichoi:
LGTM

PR #711, opened 2017-03-30
Fix wrong document format and code in system I/O module

hs0225:
- change `iotjs_xxx_instance_from_jval` to static function.

### DIFF
--- a/src/js/gpio.js
+++ b/src/js/gpio.js
@@ -38,7 +38,6 @@ Gpio.prototype.DIRECTION = gpio.DIRECTION;
 Gpio.prototype.MODE = gpio.MODE;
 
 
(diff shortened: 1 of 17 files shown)


src/js/gpio.js
```diff
@@ -100,7 +100,7 @@ function GpioPin(configuration, callback) {
   })(this));
 }
 
-// gpio.write(value[, callback])
+// gpiopin.write(value[, callback])
```
glistening: I don't think these comments give more information. What about getting rid of these comments?

hs0225: I agree with your opinion. I will remove them.

yichoi:
LGTM

PR #722, opened 2017-04-05
Allow to use JerryScript-debugger

bzsolt:


LaszloLango:
Some documentation would be great for this feature. How to build with debugger and how to use. Are there any limitations when IoT.js is built with snapshot. etc.

bzsolt:
Good point, I've added a doc file. Thank you.

docs/devs/Use-Jerry-Debugger.md
```diff
@@ -0,0 +1,25 @@
+## Jerry-debugger
+
+Detailed description about the debugger is available
+[here](https://github.com/jerryscript-project/jerryscript/blob/master/docs/07.DEBUGGER.md).
+
+### Enable debugger support in IoT.js
+
+To enable the debugger support under IoT.js, the `--jerry-debugger` option
+should be passed to the `tools/build.py`. The debugger doesn't work with
+jerry-libc therefore this option will implicitly turn on the system libc.
```
zherczeg: I don't think IoT.js works with jerry libc.

bzsolt: Yes, you're right, I would just like to emphasize that this option not only enables the debugger interface, but also ensures that the system libc is used by JerryScript. I've modified this part.

docs/devs/Use-Jerry-Debugger.md
```diff
@@ -0,0 +1,24 @@
...
+you can do so with the `--jerry-debugger-port=<PORT>` option.
+
+### Usage
+
+To start the debugger-server: `<iotjs binary> test.js --start-debug-server`
+
+Two clients are included, a [python](https://github.com/jerryscript-project/jerryscript/blob/master/jerry-debugger/jerry-client-ws.py)
+and an [HTML](https://github.com/jerryscript-project/jerryscript/blob/master/jerry-debugger/jerry-client-ws.html) variant, they can be found under `deps/jerry/jerry-debugger/`.
+
+When snapshot support is enabled, you won't be able to examine js-modules
```
LaszloLango: `*Note*: When snapshot ...`

yichoi:
Please rebase it to the master.

bzsolt:
Done, I've rebased it.

Issue #710, opened 2017-03-30
Refactor platform folder structure


hs0225:
There are a lot of unnecessary files that must exist because of the build test in the current `src/platform` folder. We need to arrange these. And the targets folder also need the arrangement.
I suggest changing the folder structure and name as below.
 - replace `.inl.h` with `.c`.
 - does not implement dummy file and not test it in Travis.
 - rename `targets` folder to `config` and put all config file in that folder.
 - put all documents in docs folder. (move the documents in targets folder to the docs folder)
 - board folder naming convention is `os-board name`(ex. nuttx-stm32f4dis)

```
├── config
│   └── nuttx-stm32f4dis
│       └── app
├── docs
│   ├── api
│   ├── build
│   ├── devs
│   ├── help
│   └── targets
└── src
    ├── js
    ├── modules
    │   └── iotjs_module_gpio.c
    └── platform
        ├── linux
        │   └── iotjs_module_gpio-linux.c
        └── nuttx
            ├── iotjs_module_gpio-nuttx.c
            └── nuttx-stm32f4dis
                └── iotjs_module_gpio-nuttx.c
```

tadziopazur:
Could you please hold off for a moment?
I think next week I'll have some time to understand and, if I have something useful to say - discuss this item.

akosthekiss:
@hs0225 In the platform subtree, I'm not sure that `platform/os/os-board` has much more expressive power than `platform/os/board`. Is there any specific reason to repeat the OS part on the path? And for the source file names, why is it needed to add a suffix to the platform-specific parts of the module implementations, or - if platform-specific suffix is a must - why don't we add `-os-board` for board-specific files (e.g., `iotjs_module_gpio-nuttx-stm32f4dis.c`), or why don't we just add a generic suffix everywhere (e.g., `-platform` or `-impl`)?

Alternatives/suggestions:
```
    src
    ├── js
    ├── modules
    │   └── iotjs_module_gpio.c
    └── platform
        ├── linux
        │   └── iotjs_module_gpio-linux.c or
                iotjs_module_gpio.c or
                iotjs_module_gpio-impl.c
        └── nuttx
            ├── iotjs_module_gpio-nuttx.c or
                iotjs_module_gpio.c
                iotjs_module_gpio-impl.c
            └── stm32f4dis
                └── iotjs_module_gpio-nuttx.c or 
                    iotjs_module_gpio-nuttx-stm32f4dis.c or 
[...]

PR #730, opened 2017-04-07
Arrange folder structure and name

hs0225:
related issue #710
 - need to specify the implemented module for each platform in `build.config`.
only exclude module has been applied. We may also need to apply the include module later.


cmake/iotjs.cmake
```diff
@@ -25,27 +25,21 @@ function(find_value RESULT VALUE VALUE_TRUE VALUE_FALSE)
...
-       ${IOTJS_SOURCE_DIR}/platform/iotjs_*-linux.c)
-  list(APPEND IOTJS_PLATFORM_SRC ${IOTJS_PLATFORM_LINUX_SRC})
-  # in case of Tizen force the platform to Linux
-  string(REGEX REPLACE "tizen$" "linux"
-         PLATFORM_DESCRIPTOR "${PLATFORM_DESCRIPTOR}")
-endif()
+string(TOLOWER ${CMAKE_SYSTEM_NAME} IOTJS_SYSTEM_OS)
+set(PLATFORM_OS_SRC
+    ${IOTJS_SOURCE_DIR}/platform/${IOTJS_SYSTEM_OS})
+set(OS_SRC ${PLATFORM_OS_SRC}/iotjs_[^module]*${IOTJS_SYSTEM_OS}.c)
```
akosthekiss: I'm surprised to see this glob working as expected. I thought that `[^module]*` would reject anything that has 'm', 'o', 'd', 'u', 'l', 'e' characters in it (i.e., that it did not look for the whole string but for any of the individual letters. (I see that something like this was there in the original code - that's why I'm surprised.)

The other thing is: however we formulate the glob, don't we want to enforce a dash before the OS suffix (i.e., `-${IOTJS_SYSTEM_OS}`)?

cmake/iotjs.cmake
```diff
@@ -25,27 +25,21 @@ function(find_value RESULT VALUE VALUE_TRUE VALUE_FALSE)
...
+file(GLOB IOTJS_OS_SRC ${OS_SRC})
+list(APPEND IOTJS_PLATFORM_SRC ${IOTJS_OS_SRC})
```
akosthekiss: Is it really necessary to do the globbing in so many steps? Wouldn't
```cmake
file(GLOB IOTJS_PLATFORM_SRC ${IOTJS_PLATFORM_DIR}/iotjs_[^module]*-${IOTJS_SYSTEM_OS}.c)
```
also achieve the same? (The example already applies some of the suggested changes.)

cmake/iotjs.cmake
```diff
@@ -25,27 +25,21 @@ function(find_value RESULT VALUE VALUE_TRUE VALUE_FALSE)
...
-  if("${BOARD_DESCRIPTOR}" STREQUAL "stm32f4dis")
-    set(BOARD_DESCRIPTOR stm32)
-  endif()
-
-  set(BOARD_SRC ${IOTJS_SOURCE_DIR}/platform/${PLATFORM_DESCRIPTOR})
-  set(BOARD_SRC ${BOARD_SRC}/iotjs_[^module]*${BOARD_DESCRIPTOR}.c)
+  set(PLATFORM_BOARD_SRC ${IOTJS_SOURCE_DIR}/platform/${IOTJS_SYSTEM_OS})
+  set(PLATFORM_BOARD_SRC ${PLATFORM_BOARD_SRC}/${IOTJS_SYSTEM_OS})
+  set(PLATFORM_BOARD_SRC ${PLATFORM_BOARD_SRC}-${TARGET_BOARD})
+  set(BOARD_SRC ${PLATFORM_BOARD_SRC}/iotjs_[^module]*${IOTJS_SYSTEM_OS}.c)
```
akosthekiss: This is too many steps, again (and/or misleading variable names), IMHO.
```cmake
set(IOTJS_BOARD_DIR ${IOTJS_PLATFORM_DIR}/${IOTJS_SYSTEM_OS}-${TARGET_BOARD})
file(GLOB IOTJS_BOARD_SRC ${IOTJS_BOARD_DIR}/iotjs_[^module]*-${IOTJS_SYSTEM_OS}.c)
```

BTW, I'm not sure that the repetition in `nuttx/nuttx-stm32f4dis` is necessary, perhaps a directory structure of `nuttx/stm32f4dis` could be just as expressive. I'll raise this remark on the issue as well but mention it here, too, as it affects the required glob pattern.

cmake/iotjs.cmake
```diff
@@ -134,38 +145,28 @@ set(PLATFORM_SRC
...
-  set(PLATFORM_MODULE_BASE ${PLATFORM_SRC}_${module}-${PLATFORM_DESCRIPTOR})
-  set(PLATFORM_MODULE_SRC ${PLATFORM_MODULE_BASE}.c)
-  if(VERBOSE)
-    message("Checking platform file: ${PLATFORM_MODULE_SRC}")
-  endif()
-  if(EXISTS "${PLATFORM_MODULE_SRC}")
-    list(APPEND IOTJS_MODULE_SRC ${PLATFORM_MODULE_SRC})
-    list(APPEND IOTJS_PLATFORM_SUPPORT ${MODULE})
+  # first, check if there is the module in <os>/<os>-<board>
+  # ../platform/<os>/<os>-board>iotjs_module_<module> ...
```
akosthekiss: Cannot interpret this line of comment. (Perhaps a leftover from the original variant of the cmake file?)

cmake/iotjs.cmake
```diff
@@ -112,11 +106,28 @@ add_custom_command(
           ${IOTJS_SOURCE_DIR}/js/*.js
 )
 
+macro(add_platform_dependency_module
+      RESULT BASE_SRC MODULE SUPPORT_LIST)
+  set(PLATFORM_MODULE_BASE
+      ${BASE_SRC}/iotjs_module_${module}-${IOTJS_SYSTEM_OS})
```
akosthekiss: all-caps should be used for `${MODULE}` to keep consistency

cmake/iotjs.cmake
```diff
@@ -112,11 +106,28 @@ add_custom_command(
           ${IOTJS_SOURCE_DIR}/js/*.js
 )
 
+macro(add_platform_dependency_module
+      RESULT BASE_SRC MODULE SUPPORT_LIST)
```
akosthekiss: Just like above, `BASE_DIR` might be a better name

hs0225:
@akosthekiss I updated PR. Please review again.

cmake/iotjs.cmake
```diff
@@ -134,42 +131,39 @@ set(PLATFORM_SRC
...
-  endif()
-  if(EXISTS "${PLATFORM_MODULE_SRC}")
-    list(APPEND IOTJS_MODULE_SRC ${PLATFORM_MODULE_SRC})
-    list(APPEND IOTJS_PLATFORM_SUPPORT ${MODULE})
+  # first, check if there is the module in <os>/<board>
+  set(ADD_MODULE_RESULT FALSE)
+  if(NOT "${TARGET_BOARD}" STREQUAL "None")
+    set(PLATFORM_MODULE_SRC ${PLATFORM_BOARD_DIR}/iotjs_module_${module})
+    set(PLATFORM_MODULE_SRC
+        ${PLATFORM_MODULE_SRC}-${IOTJS_SYSTEM_OS}-${TARGET_BOARD}.c)
```
akosthekiss: Why the two-step assignment?
```cmake
set(PLATFORM_MODULE_SRC ${PLATFORM_BOARD_DIR}/iotjs_module_${module}-${IOTJS_SYSTEM_OS}-${TARGET_BOARD}.c)
```

cmake/iotjs.cmake
```diff
@@ -25,28 +25,25 @@ function(find_value RESULT VALUE VALUE_TRUE VALUE_FALSE)
...
 
 # Board Configuration (not module)
-if(DEFINED TARGET_BOARD)
-  string(TOLOWER "${TARGET_BOARD}" BOARD_DESCRIPTOR)
-
-  if("${BOARD_DESCRIPTOR}" STREQUAL "stm32f4dis")
-    set(BOARD_DESCRIPTOR stm32)
+if(NOT "${TARGET_BOARD}" STREQUAL "None")
+  set(PLATFORM_BOARD_DIR
+      ${IOTJS_SOURCE_DIR}/platform/${IOTJS_SYSTEM_OS}/${TARGET_BOARD})
```
hs0225: I fixed it.

cmake/iotjs.cmake
hs0225: I fixed it.

cmake/iotjs.cmake
```diff
@@ -25,28 +25,25 @@ function(find_value RESULT VALUE VALUE_TRUE VALUE_FALSE)
...
+  file(GLOB IOTJS_BOARD_SRC ${PLATFORM_BOARD_DIR}/iotjs_*.c)
+  file(GLOB PLATFORM_MODULE_SRC ${PLATFORM_BOARD_DIR}/iotjs_module*.c)
```
hs0225: Done.

cmake/iotjs.cmake
```diff
@@ -25,28 +25,25 @@ function(find_value RESULT VALUE VALUE_TRUE VALUE_FALSE)
...
-  list(APPEND IOTJS_PLATFORM_SRC ${IOTJS_PLATFORM_LINUX_SRC})
-  # in case of Tizen force the platform to Linux
-  string(REGEX REPLACE "tizen$" "linux"
-         PLATFORM_DESCRIPTOR "${PLATFORM_DESCRIPTOR}")
+# System Configuration (not module)
+string(TOLOWER ${CMAKE_SYSTEM_NAME} IOTJS_SYSTEM_OS)
+set(PLATFORM_OS_DIR
+    ${IOTJS_SOURCE_DIR}/platform/${IOTJS_SYSTEM_OS})
+file(GLOB IOTJS_PLATFORM_SRC ${PLATFORM_OS_DIR}/iotjs_*.c)
+file(GLOB PLATFORM_MODULE_SRC ${PLATFORM_OS_DIR}/iotjs_module*.c)
```
akosthekiss: `module_*` should be used, as done below

cmake/iotjs.cmake
akosthekiss: ah, that's so bad :(

tools/precommit.py
```diff
@@ -117,28 +119,44 @@ def build(buildtype, args=[]):
...
-    include_module = ['--iotjs-include-module=' + include_module]
-else:
-    include_module = []
+os_dependency_module = \
+  get_os_dependency_exclude_module(config['module']['exclude'])
+
+for os_name in os_dependency_module:
+    if os_dependency_module[os_name]:
+        os_dependency_module[os_name] = \
+        ['--iotjs-include-module=' + ','.join(os_dependency_module[os_name])]
```
akosthekiss: I've never really understood this part of code, not even in its original form. Whatever we give in the config file in the `['module']['exclude']` section, will be passed on to the build script as a module to be included with `--iotjs-include-module`. What is the rationale behind that? It is a bit counter-intuitive, so at least some comments explaining it would be helpful.

hs0225: Excluded modules are not tested for builds. To test build all implemented modules in Travis, this code includes the excluded modules in the build test by using `--iotjs-include-module` command. In other words, Travis will test all implemented modules. I commented on this.

cmake/iotjs.cmake
hs0225: I fixed it.

tools/precommit.py
```diff
@@ -117,28 +119,46 @@ def build(buildtype, args=[]):
...
-if len(config['module']['exclude']) > 0 :
-    include_module = ','.join(config['module']['exclude'])
-    include_module = ['--iotjs-include-module=' + include_module]
-else:
-    include_module = []
+os_dependency_module = \
+  get_os_dependency_exclude_module(config['module']['exclude'])
+
+# Excluded modules are also included in the build test.
+# Travis will test all implemented modules.
```
akosthekiss: Honestly, the deeper I dig into the semantics of the "exclude" list in build.config, build.py, and precommit.py, the more I get confused. (E.g.,`--iotjs-exclude-module` is never used anywhere. The "exclude" option in build.config is only used by build.py to check whether they are not required by included modules. However, by precommit.py, they are turned into included modules for build tests.) It really sounds like a misnomer to me, or two concepts mixed into one (not sure yet which one is the case).

I acknowledge that the core of this logic is inherited, so I'm not insisting to fix it in this patch (if you don't want to), as it might require a bit bigger or differently scoped refactoring. But please do let me know whether you plan to tackle it (now or later). If you'd like to leave it for someone else, I'll open a discussion issue for this topic.

tools/module_analyzer.py
```diff
@@ -163,6 +170,7 @@ def _load_options(argv):
 
     arg_config = list(filter(lambda x: x.startswith('--config='), argv))
     config_path = path.BUILD_CONFIG_PATH
+    target_os = list(filter(lambda x: x.startswith('--target_os='), argv))
```
hs0225: I look for target os in a different way. I no longer use a list.

tools/precommit.py
```diff
@@ -117,28 +119,46 @@ def build(buildtype, args=[]):
...
+
 option = parse_option()
 config = get_config()
-if len(config['module']['exclude']) > 0 :
-    include_module = ','.join(config['module']['exclude'])
-    include_module = ['--iotjs-include-module=' + include_module]
-else:
-    include_module = []
+os_dependency_module = \
+  get_os_dependency_exclude_module(config['module']['exclude'])
```
hs0225: Done.

tools/precommit.py
hs0225: I agree that `exclude` in `build.config` is misleading. (But, I think `--iotjs-exclude-module` command is necessary.) I don't have any specific plans for this issue. I think we need to discuss this topic.

yichoi:

hs0225:
@yichoi I completed rebasing this PR.

yichoi:
LGTM

PR #734, opened 2017-04-07
Fix build environment for nuttx (and derivatives)

tadziopazur:
So far we did not control dependencies of the build, such as:
 - C library
 - include files
and the include files could have been resolved to either:
 - target environment (if accessed locally)
 - bundled with the compiler
 - host default paths (/usr/include)
This is a problem in all builds but the host build (when the latter are the same).

Fix build dependencies:
 - pass sysroot to cmake (as CFLAGS) if set
 - exclude non-target dependencies on nuttx/tinyara (-nostd{inc,lib})

### DIFF
--- a/tools/build.py
+++ b/tools/build.py
@@ -200,16 +200,6 @@ def init_options():
 
 
 def adjust_options(options):
-    # First fix some option inconsistencies
-    if options.target_os in ['nuttx', 'tizenrt']:
-        options.buildlib = True
-        if not options.sysroot:
-            ex.fail('--sysroot needed for nuttx target')
-
-        options.sysroot = fs.abspath(options.sysroot)
-        if not fs.exists(options.sysroot):
-            ex.fail('Nuttx sysroot %s does not exist' % options.sysroot)
-
     if options.target_arch == 'x86':
         options.target_arch = 'i686'
     if options.target_arch == 'x64':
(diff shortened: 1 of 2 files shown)


tools/build.py
```diff
@@ -230,6 +220,27 @@ def adjust_options(options):
...
 
+    # Verify options consistency (sysroot vs. cross build)
+    if options.target_tuple != options.host_tuple:
+        # This is cross build, sysroot should/must be supplied
+        if options.target_os in ['nuttx', 'tizenrt']:
+            require_sysroot = True
+            options.buildlib = True
+        if not options.sysroot:
+            if require_sysroot:
+                ex.fail('--sysroot required for nuttx/tinyara')
```
akosthekiss: Is that tizenrt or tinyara? (The user is expected to specify the OS as tizenrt, so better give notifications in the same format.)

BTW, a second thought: even though right now only nuttx and tizenrt requires sysroot, should the build system evolve, others may require it, too. Then, this message will have to be updated again and again. Having an error string like `'--sysroot required for %s' % options.target_os` would be more future-proof (and would also solve the problem I've pointed out).

tools/build.py
```diff
@@ -230,6 +220,27 @@ def adjust_options(options):
...
+            else:
+                ex.warn('--sysroot recommended for cross builds target')
+        options.sysroot = fs.abspath(options.sysroot)
+        if not fs.exists(options.sysroot):
+            ex.fail('sysroot %s does not exist' % options.sysroot)
+
+    arch_for_iotjs = 'arm' if options.target_arch[0:3] == 'arm' else \
```
akosthekiss: `options.target_arch.startswith('arm')` might be better. No need to juggle with substring indices.

tools/build.py
```diff
@@ -285,6 +296,20 @@ def build_cmake_args(options, for_jerry=False):
...
+    # while prebuild GCC mostly comes with glibc/newlib.
+    # TODO: consider factoring this out to an option
+    if options.target_os in ['nuttx', 'tizenrt']:
+        compile_flags.extend(['-nostdinc', '-nostdlib'])
+
+    # With sysroot set, exclude default (host) directories from build.
+    if options.sysroot:
+        compile_flags.extend(['--sysroot', options.sysroot])
+        compile_flags.extend(['-isystem =/include',
+                              '-isystem =/include/tinyara'])
```
akosthekiss: Shouldn't this be OS-specific? For generic NuttX, this include path is invalid/useless.

tools/build.py
tadziopazur: Done

tools/build.py
tadziopazur: Done

tadziopazur:
Travis checks do fail, for this change depends on libtuv and jerry being updated to support tinzenrt
https://github.com/jerryscript-project/jerryscript/pull/1720,
https://github.com/Samsung/libtuv/pull/74

yichoi:

PR #735, opened 2017-04-07
Fixed http_server.writeHead function

tosyu:
Hello,

I have fixed writeHead function in the http_server module. The problem was that `Object.keys` returns an array so the `for..in` loop was iterating over array properties, not array values (which were numerical). And this wass set as response header which have results like `{0: undefined, 1: undefined}`


yichoi:
LGTM

tosyu:
Thank you

PR #737, opened 2017-04-07
Console.log crash problem fix

KonradNowicki:
console.log("test", null, undefined);

When there are some null parameters console.log is crushing.


yichoi:

PR #742, opened 2017-04-10
Fix build configuration for TizenRT

haesik:
Current Configuration doesn't build jerry-libm for TizenRT


cmake/jerry.cmake
```diff
@@ -56,12 +56,12 @@ endif()
 
 
 # use system libc/libm on Unix like targets
-if("${TARGET_OS}" MATCHES "LINUX|TIZEN|DARWIN")
+if("${TARGET_OS}" STREQUAL "LINUX|TIZEN|DARWIN")
```
akosthekiss: I'm surprised that this works. According to my best understanding, STREQUAL of cmake looks for equivalence of strings. How can TARGET_OS be ever equal to "LINUX|TIZEN|DARWIN" (with all three OS names and the pipes inbetween)?

yichoi:
LGTM

PR #744, opened 2017-04-10
Provisions to reliably build in cross environment

tadziopazur:
So far the build proceeded by chance -- pieces were taken from either:
 - target environment (if sysroot was passed)
 - compiler bundled environment (hosted cross builds)
 - host system (yes, even for cross builds it happens)

Not sure this is the last patch addressing the problem. There are issues still:
 - any cross build should be constrained to target or compiler environment
 - perhaps this should be extracted as a build flag (bar compiler from accessing bundled sysroot), something like --freestanding? This would be useful to build with compiler that has incompatible libc (i.e.: newlib bundled with popular linaro arm-none-eabi-gcc 4.9.3 against nuttx bundled uClibc)

Fix build dependencies:
 - pass sysroot to cmake (as CFLAGS) if set
 - exclude non-target dependencies on nuttx/tinyara (-nostd{inc,lib})

### DIFF
--- a/tools/build.py
+++ b/tools/build.py
@@ -200,16 +200,6 @@ def init_options():
 
 
 def adjust_options(options):
-    # First fix some option inconsistencies
-    if options.target_os in ['nuttx', 'tizenrt']:
-        options.buildlib = True
-        if not options.sysroot:
-            ex.fail('--sysroot needed for nuttx target')
-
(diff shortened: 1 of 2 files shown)


tadziopazur:
Bugger! forgot I have a PR in flight covering this.

PR #753, opened 2017-04-13
Refactor testsuite to not use parent directory hard-link

rtakacs:
Since [NuttX does not use file system that supports parent directory hard-link](https://bitbucket.org/nuttx/nuttx/issues/39/relative-path-with-parent-directory-symbol), I've removed that hard-link from relative paths in tests.

Testrunner is just a bit modified. Now, there is no need to change directory (to the current testset folder), tests are executed always from the root test directory.

### DIFF
--- a/test/run_pass/test_fs.js
+++ b/test/run_pass/test_fs.js
@@ -19,7 +19,7 @@ var fs = require('fs');
 var assert = require('assert');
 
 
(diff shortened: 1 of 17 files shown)


yichoi:
LGTM

hs0225:
LGTM

PR #762, opened 2017-04-18
Remove BUILDTYPE cmake argument from http-parser

galpeter:
After the latest http-parser update the BUILDTYPE
cmake argument is now obsolete and can be removed.

The CMAKE_BUILD_TYPE argument is used instead.


yichoi:
LGTM

PR #763, opened 2017-04-18
Fix JerryScript api header reference in dev docs

galpeter:
jerry-api.h is deprecated the docs should refer to
the newer jerryscript.h.


yichoi:
LGTM

PR #772, opened 2017-04-20
Fix require message to include missing module name

kolipka:


yichoi:
LGTM

PR #773, opened 2017-04-20
[Code Coverage] Covering buffer + removing unneccessary functions

DanielBallaSZTE:
Buffer.js got a 100% coverage.
In addition there was an unnecessary if statement which checked whether the object calling it is a Buffer or not. This is not needed, since the function itself can only be called from a Buffer object, therefore this test would never run.
In iotjs_module_buffer.c IOTJS_ASSERT made the script crash, thus got commented.

### DIFF
--- a/src/modules/iotjs_module_buffer.c
+++ b/src/modules/iotjs_module_buffer.c
@@ -131,7 +131,6 @@ static unsigned hex2bin(char c) {
   if (c >= 'a' && c <= 'f')
     return 10 + (c - 'a');
 
-  IOTJS_ASSERT(!"Bad argument");
   return (unsigned)(-1);
 }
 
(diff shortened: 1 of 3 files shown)


src/modules/iotjs_module_buffer.c
```diff
@@ -131,7 +131,7 @@ static unsigned hex2bin(char c) {
   if (c >= 'a' && c <= 'f')
     return 10 + (c - 'a');
 
-  IOTJS_ASSERT(!"Bad argument");
+  // IOTJS_ASSERT(!"Bad argument");
```
LaszloLango: Remove this line if it's not needed anymore.

DanielBallaSZTE:
@LaszloLango The requested change is done, the line got deleted.

yichoi:
LGTM

PR #775, opened 2017-04-21
Switch to MinSizeRel build type if we are doing Release build

galpeter:
By swithing to MinSizeRel the -Os option is automatically
passed to the compilation of the source files.


galpeter:
On RPI2 there is a ~11 kbytes size reduction.

yichoi:
LGTM

PR #776, opened 2017-04-21
Use array for headers in http_common.js

galpeter:
Previously for storing headers an object was used as an array and
had a helper method do add elements into the headers object.

As the JS engine has full array support the logic can be
replaced with an array.


yichoi:
LGTM

PR #778, opened 2017-04-25
Fix dns module test.

knightburton:
- Add a missing branch test.
 - Modify the error code check in the invalid hostname test cases.
      The returned code depends on the used system,
      in this case the -3007 and the -3008 error codes are valid.


yichoi:
LGTM

PR #780, opened 2017-04-25
Merge tryPath and statPath functions in module.js

galpeter:
The statPath function was only used by the tryPath function.
By merging the two functions the .rodata size decreases
a bit.


yichoi:
LGTM

PR #781, opened 2017-04-25
Replace concatdir usage with array concat in module.js

galpeter:
As the JS engine has full es5.1 support there is no need
to manually concatenate arrays.


yichoi:
LGTM

PR #783, opened 2017-04-25
Detect supported modules when running the tests

galpeter:
The patch will allow the test framework to automatically
detect which modules are built-in the IoT.js. Based on this
information it can check if the input test file contains
one of the supported module name and if not the test can
be skipped.

### DIFF
--- a/tools/test_runner.js
+++ b/tools/test_runner.js
@@ -15,6 +15,8 @@
 
 var testdriver = require('testdriver');
 var console_wrapper = require('common_js/module/console');
+var builtin_modules =
+  Object.keys(process.native_sources).concat(Object.keys(process.binding));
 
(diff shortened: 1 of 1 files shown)


tools/test_runner.js
```diff
@@ -15,6 +15,8 @@
 
 var testdriver = require('testdriver');
 var console_wrapper = require('common_js/module/console');
+var builtin_modules = ['process'].concat(Object.keys(process.native_sources),
```
hs0225: `process` is already included `process.native_sources`.

galpeter: True, will fix asap.

galpeter:
@hs0225 I've update the pr.

yichoi:
LGTM

PR #787, opened 2017-04-26
Activate sanitizer-jobs on Travis CI

bzsolt:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -51,7 +51,7 @@ matrix:
           packages:
             - gcc-4.9
             - gcc-4.9-multilib
-      env: OPTS="--test=host-linux --buildtype=debug --buildoptions=--target-arch=i686,--compile-flag=-fsanitize=address,--compile-flag=-fno-omit-frame-pointer,--compile-flag=-fno-common,--jerry-cmake-param=-DJERRY_LIBC=OFF,--jerry-cmake-param=-DFEATURE_SYSTEM_ALLOCATOR=ON,--no-snapshot,--no-check-valgrind" INSTALL_TRAVIS_I686_DEPS=yes ASAN_OPTIONS=detect_stack_use_after_return=1:check_initialization_order=true:strict_init_order=true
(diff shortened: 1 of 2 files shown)


bzsolt:
I would like to activate the sanitizer jobs on the CI, but despite the extra `-O2` compile flag, the `test_net_3.js` still has timeout.
I've looked for a way what allow to specify greater timeout interval just in this sanitizer case (e.g. via an environment variable) but each test has (or not) an own timeout value in `test/testsets.json`.
The current solution (increase the timeout up to 100 sec) looks like an arbitrary step, but I couldn't find any proper way to do this. Anyone has a better idea?

.travis.yml
```diff
@@ -61,6 +61,4 @@ matrix:
             - gcc-4.9
             - gcc-4.9-multilib
       env: OPTS="--test=host-linux --buildtype=debug --buildoptions=--target-arch=i686,--compile-flag=-fsanitize=undefined,--compile-flag=-fno-omit-frame-pointer,--compile-flag=-fno-common,--jerry-cmake-param=-DJERRY_LIBC=OFF,--jerry-cmake-param=-DFEATURE_SYSTEM_ALLOCATOR=ON,--no-snapshot,--no-check-valgrind" INSTALL_TRAVIS_I686_DEPS=yes UBSAN_OPTIONS=print_stacktrace=1
-  allow_failures:
```
daeyeon: The fix itself works for me. BTW, there is another task named with `compiler: gcc-4.9`. Once remove this, then the task will be also active. Is it your intention? Can you kindly let me know what difference is b/t each tasks?

bzsolt: Yes, it's intentional.
The first task is an address-sanitizer (`--compile-flag=-fsanitize=address`), the second one is an undefined-behaviour sanitizer (`--compile-flag=-fsanitize=undefined`).

daeyeon: Thanks for clarifying that.

daeyeon:
LGTM

yichoi:
LGTM

PR #789, opened 2017-04-27
Fix argument types of 'iotjs_timerwrap_start'

LaszloLango:
Time and repeat time arguments cannot be negtive numbers.
'uv_timer_start' also expects uint64_t arguments.


yichoi:
LGTM

PR #794, opened 2017-04-27
Inline emitError into error prototype method

galpeter:


yichoi:
LGTM

PR #797, opened 2017-04-28
Use array for headers in http_client/server

daeyeon:
The type of `headers` in `http_common` has recently been replaced
to array from object. This commit changes a few parts found in
`http_client/server` which handle it as object.

### DIFF
--- a/src/js/http_client.js
+++ b/src/js/http_client.js
@@ -73,7 +73,7 @@ function setupConnection(req, socket) {
   req.connection = socket;
   parser.socket = socket;
   parser.incoming = null;
-  parser._headers = {};
+  parser._headers = [];
   req.parser = parser;
 
   socket.parser = parser;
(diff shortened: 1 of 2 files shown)


yichoi:
LGTM

glistening:
@daeyeon Could you add test case for this issue?

glistening:
@daeyeon Perhaps, `http_outgoing.js` also needs same modifications for `_headers`.

daeyeon:
@glistening Thank you for the review. `http_outgoing.js` and `http_incoming.js` have a property named the same, `_headers`. But, it seems not to be used on the HTTPParser context. IMO, we can consider using array on those files in another commit later.

glistening:
@daeyeon You mean the we can keep ` _headers` in `http_outgoing.js` as object without trouble? If so, LGTM. But personally I can't understand the inconsistency.

daeyeon:
I made #798 for more discussion. Thanks.

PR #799, opened 2017-04-28
[Code Coverage] Improve coverage of fs.js

DanielBallaSZTE:

### DIFF
--- a/test/run_pass/test_fs_exists.js
+++ b/test/run_pass/test_fs_exists.js
@@ -25,6 +25,22 @@ var assert = require('assert');
 }
 
 {
+  var filePath = "";
+
+  assert.doesNotThrow(function(){
+    fs.exists(filePath);
+  });
+}
+
+{
+  var filePath = "resources/tobeornottobe.txt";
+
+  assert.doesNotThrow(function(){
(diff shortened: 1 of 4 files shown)


test/run_pass/test_fs_exists.js
```diff
@@ -25,6 +25,18 @@ var assert = require('assert');
 }
 
 {
+  var filePath = "";
+
+  fs.exists(filePath);
```
daeyeon: The test case isn't completed with the comparison between expectation and result using assert. This problem can be found in other files as well as here.

DanielBallaSZTE:
@daeyeon I've changed those two cases to check with assert, I didn't think it would need assertion, since the else case of both ifs are undefined.

daeyeon:
@DanielBallaSZTE The others mentioned for same reason were not addressed. I added my suggestion in detail.

DanielBallaSZTE:
@daeyeon Thanks for the suggestion, I'm going to keep those in mind next time. I hope everything is fixed now.

daeyeon:
LGTM

yichoi:
LGTM

PR #800, opened 2017-04-28
Fix sign conversions and add the warning check to the CMake

LaszloLango:

### DIFF
--- a/src/iotjs_util.c
+++ b/src/iotjs_util.c
@@ -27,8 +27,9 @@ iotjs_string_t iotjs_file_read(const char* path) {
   IOTJS_ASSERT(file != NULL);
 
   fseek(file, 0, SEEK_END);
-  long len = ftell(file);
-  IOTJS_ASSERT(len >= 0);
+  long ftell_ret = ftell(file);
+  IOTJS_ASSERT(ftell_ret >= 0);
+  size_t len = (size_t)ftell_ret;
   fseek(file, 0, SEEK_SET);
 
   char* buffer = iotjs_buffer_allocate(len + 1);
@@ -45,7 +46,7 @@ iotjs_string_t iotjs_file_read(const char* path) {
 #else
   size_t read = fread(buffer, 1, len, file);
 #endif
-  IOTJS_ASSERT(read == (size_t)len);
+  IOTJS_ASSERT(read == len);
 
   *(buffer + len) = 0;
 
@@ -57,14 +58,14 @@ iotjs_string_t iotjs_file_read(const char* path) {
 }
 
 
-char* iotjs_buffer_allocate(unsigned size) {
+char* iotjs_buffer_allocate(size_t size) {
   char* buffer = (char*)(calloc(size, sizeof(char)));
   IOTJS_ASSERT(buffer != NULL);
   return buffer;
 }
 
 
-char* iotjs_buffer_reallocate(char* buffer, unsigned size) {
+char* iotjs_buffer_reallocate(char* buffer, size_t size) {
   IOTJS_ASSERT(buffer != NULL);
   return (char*)(realloc(buffer, size));
 }
--- a/src/modules/iotjs_module_buffer.c
+++ b/src/modules/iotjs_module_buffer.c
@@ -16,6 +16,7 @@
 #include "iotjs_def.h"
 #include "iotjs_module_buffer.h"
 
+#include <math.h>
 #include <stdlib.h>
 #include <string.h>
 
@@ -112,26 +113,26 @@ size_t iotjs_bufferwrap_length(iotjs_bufferwrap_t* bufferwrap) {
 }
 
 
-static size_t bound_range(int index, size_t low, size_t upper) {
-  if (index < (int)low) {
+static size_t bound_range(size_t index, size_t low, size_t upper) {
+  if (index == SIZE_MAX) {
     return low;
   }
-  if (index > (int)upper) {
+  if (index > upper) {
     return upper;
   }
   return index;
 }
 
 
-static unsigned hex2bin(char c) {
+static int8_t hex2bin(char c) {
   if (c >= '0' && c <= '9')
-    return c - '0';
+    return (int8_t)(c - '0');
   if (c >= 'A' && c <= 'F')
-    return 10 + (c - 'A');
+    return (int8_t)(10 + (c - 'A'));
   if (c >= 'a' && c <= 'f')
-    return 10 + (c - 'a');
+    return (int8_t)(10 + (c - 'a'));
 
-  return (unsigned)(-1);
+  return (int8_t)(-1);
 }
 
 
@@ -140,9 +141,9 @@ static size_t hex_decode(char* buf, size_t len, const char* src,
   size_t i;
 
   for (i = 0; i < len && i * 2 + 1 < srcLen; ++i) {
-    unsigned a = hex2bin(src[i * 2 + 0]);
-    unsigned b = hex2bin(src[i * 2 + 1]);
-    if (!~a || !~b)
+    int8_t a = hex2bin(src[i * 2 + 0]);
+    int8_t b = hex2bin(src[i * 2 + 1]);
+    if (a == -1 || b == -1)
       return i;
     buf[i] = (a << 4) | b;
   }
@@ -226,7 +227,7 @@ JHANDLER_FUNCTION(Buffer) {
 
   const iotjs_jval_t* jbuiltin = JHANDLER_GET_THIS(object);
   const iotjs_jval_t* jbuffer = JHANDLER_GET_ARG(0, object);
-  int length = JHANDLER_GET_ARG(1, number);
+  size_t length = JHANDLER_GET_ARG(1, number);
 
   iotjs_jval_set_property_jval(jbuiltin, IOTJS_MAGIC_STRING__BUFFER, jbuffer);
 
@@ -252,6 +253,17 @@ JHANDLER_FUNCTION(Compare) {
 }
 
 
+#define DECLARE_SIZE_T_FROM_DOUBLE(n, d) \
+  size_t n;                              \
+  do {                                   \
+    if (d < 0 || isnan(d) || isinf(d)) { \
+      n = SIZE_MAX;                      \
+    } else {                             \
+      n = (size_t)d;                     \
+    }                                    \
+  } while (0)
+
+
 JHANDLER_FUNCTION(Copy) {
   JHANDLER_CHECK_THIS(object);
   JHANDLER_CHECK_ARGS(4, object, number, number, number);
@@ -264,15 +276,16 @@ JHANDLER_FUNCTION(Copy) {
   iotjs_bufferwrap_t* dst_buffer_wrap =
       iotjs_bufferwrap_from_jbuffer(jdst_buffer);
 
-  int dst_start = JHANDLER_GET_ARG(1, number);
-  int src_start = JHANDLER_GET_ARG(2, number);
-  int src_end = JHANDLER_GET_ARG(3, number);
-
   size_t dst_length = iotjs_bufferwrap_length(dst_buffer_wrap);
   size_t src_length = iotjs_bufferwrap_length(src_buffer_wrap);
 
+  DECLARE_SIZE_T_FROM_DOUBLE(dst_start, JHANDLER_GET_ARG(1, number));
   dst_start = bound_range(dst_start, 0, dst_length);
+
+  DECLARE_SIZE_T_FROM_DOUBLE(src_start, JHANDLER_GET_ARG(2, number));
   src_start = bound_range(src_start, 0, src_length);
+
+  DECLARE_SIZE_T_FROM_DOUBLE(src_end, JHANDLER_GET_ARG(3, number));
   src_end = bound_range(src_end, 0, src_length);
 
   if (src_end < src_start) {
@@ -291,15 +304,16 @@ JHANDLER_FUNCTION(Write) {
   JHANDLER_CHECK_ARGS(3, string, number, number);
 
   iotjs_string_t src = JHANDLER_GET_ARG(0, string);
-  int offset = JHANDLER_GET_ARG(1, number);
-  int length = JHANDLER_GET_ARG(2, number);
 
   const iotjs_jval_t* jbuiltin = JHANDLER_GET_THIS(object);
 
   iotjs_bufferwrap_t* buffer_wrap = iotjs_bufferwrap_from_jbuiltin(jbuiltin);
 
   size_t buffer_length = iotjs_bufferwrap_length(buffer_wrap);
+  DECLARE_SIZE_T_FROM_DOUBLE(offset, JHANDLER_GET_ARG(1, number));
   offset = bound_range(offset, 0, buffer_length);
+
+  DECLARE_SIZE_T_FROM_DOUBLE(length, JHANDLER_GET_ARG(2, number));
   length = bound_range(length, 0, buffer_length - offset);
   length = bound_range(length, 0, iotjs_string_size(&src));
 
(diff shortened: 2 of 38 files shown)


src/iotjs_exception.c
```diff
@@ -22,7 +22,7 @@
 #include "uv.h"
 
 
-iotjs_jval_t iotjs_create_uv_exception(int errorno, const char* syscall) {
+iotjs_jval_t iotjs_create_uv_exception(ssize_t errorno, const char* syscall) {
```
chokobole: Why do we have to change type to `ssize_t` here? I think here `int` is enough because the function `uv_strerror` using `errorno` accepts `int`, and this amount of range will be fixed(Currently 76) at some point. Even though it is run on how many bit machine. the number itself won't increase.
```c
const char* uv_strerror(int err) {
```

src/modules/iotjs_module_buffer.c
```diff
@@ -112,26 +112,26 @@ size_t iotjs_bufferwrap_length(iotjs_bufferwrap_t* bufferwrap) {
 }
 
 
-static size_t bound_range(int index, size_t low, size_t upper) {
-  if (index < (int)low) {
+static size_t bound_range(int32_t index, size_t low, size_t upper) {
```
chokobole: In semantic, index seems to be proper to have `size_t`

src/modules/iotjs_module_buffer.c
```diff
@@ -112,26 +112,26 @@ size_t iotjs_bufferwrap_length(iotjs_bufferwrap_t* bufferwrap) {
...
+  if (index > (int32_t)upper) {
     return upper;
   }
-  return index;
+  return (size_t)index;
 }
 
 
-static unsigned hex2bin(char c) {
+static uint8_t hex2bin(char c) {
```
chokobole: For me, as we return -1, it is natural to have `int8_t`.

src/modules/iotjs_module_buffer.c
```diff
@@ -140,9 +140,9 @@ static size_t hex_decode(char* buf, size_t len, const char* src,
   size_t i;
 
   for (i = 0; i < len && i * 2 + 1 < srcLen; ++i) {
-    unsigned a = hex2bin(src[i * 2 + 0]);
-    unsigned b = hex2bin(src[i * 2 + 1]);
-    if (!~a || !~b)
+    uint8_t a = hex2bin(src[i * 2 + 0]);
+    uint8_t b = hex2bin(src[i * 2 + 1]);
```
chokobole: ditto

src/modules/iotjs_module_buffer.c
```diff
@@ -317,20 +316,20 @@ JHANDLER_FUNCTION(WriteUInt8) {
...
 
   iotjs_bufferwrap_t* buffer_wrap = iotjs_bufferwrap_from_jbuiltin(jbuiltin);
 
   size_t buffer_length = iotjs_bufferwrap_length(buffer_wrap);
-  offset = bound_range(offset, 0, buffer_length);
-  length = bound_range(length, 0, buffer_length - offset);
-  length = bound_range(length, 0, 1);
+  size_t offset =
+      bound_range((int)JHANDLER_GET_ARG(1, number), 0, buffer_length);
+  length = (int)bound_range(length, 0, buffer_length - offset);
```
chokobole: I think `length` should be `size_t`

src/modules/iotjs_module_buffer.c
```diff
@@ -411,12 +409,12 @@ JHANDLER_FUNCTION(Slice) {
...
 
-  iotjs_jval_t jnew_buffer = iotjs_bufferwrap_create_buffer(length);
+  iotjs_jval_t jnew_buffer = iotjs_bufferwrap_create_buffer((size_t)length);
   iotjs_bufferwrap_t* new_buffer_wrap =
       iotjs_bufferwrap_from_jbuffer(&jnew_buffer);
   iotjs_bufferwrap_copy_internal(new_buffer_wrap,
-                                 iotjs_bufferwrap_buffer(buffer_wrap), start,
-                                 end, 0);
+                                 iotjs_bufferwrap_buffer(buffer_wrap),
+                                 (size_t)start, (size_t)end, 0);
```
chokobole: I think `start` and `end` should be declared as `size_t`!

src/modules/iotjs_module_i2c.c
```diff
@@ -223,7 +223,7 @@ static void GetI2cArray(const iotjs_jval_t* jarray,
   req_data->buf_len = iotjs_jval_as_number(&jlength);
   req_data->buf_data = iotjs_buffer_allocate(req_data->buf_len);
 
-  for (int i = 0; i < req_data->buf_len; i++) {
+  for (uint32_t i = 0; i < req_data->buf_len; i++) {
```
chokobole: As `buf_len` is declared as `uint8_t`, so here `uint8_t` seems better, I think.

src/iotjs_util.c
```diff
@@ -43,28 +43,29 @@ iotjs_string_t iotjs_file_read(const char* path) {
...
 #endif
   IOTJS_ASSERT(read == (size_t)len);
 
   *(buffer + len) = 0;
 
   fclose(file);
 
-  iotjs_string_t contents = iotjs_string_create_with_buffer(buffer, len);
+  iotjs_string_t contents =
+      iotjs_string_create_with_buffer(buffer, (uint32_t)len);
```
chokobole: I think `len` should have `size_type`. Basically every time using some variable if we have to cast with some specific type, then I think this variable should have declared with the type to cast at the first time.

src/modules/iotjs_module_buffer.c
LaszloLango: That was my first thought too. The problem with `size_t index` is on the call sites of `bound_range`. It is usually used without any validation after `JHANDLER_GET_ARG(1, number)` calls. So the first option is what I did in the PR and second one is to lot of copy-paste code to validate the number value before we call `bound_range`. I think this one is simpler and better maintainable.

src/iotjs_exception.c
LaszloLango: right, good catch

src/modules/iotjs_module_buffer.c
LaszloLango: They could be negative numbers

src/iotjs_util.c
LaszloLango: The return value of `ftell` is long and there is an assert for checking the returned value is a positive number. It is only safe cast after that. We can declare a new variable for the return value if you think it's better.
E.g:
```c
long ftell_ret = ftell(file);
IOTJS_ASSERT(ftell_ret >= 0);
size_t len = (size_t)ftell_ret;

...

iotjs_string_t contents =
    iotjs_string_create_with_buffer(buffer, len);
```

BTW, I was thinking on that we might check the return value of `ftell` in release mode too.

src/modules/iotjs_module_buffer.c
chokobole: First, very sorry to late reply, It was long holidays in korea. Then how about trying cast that parameter to `size_t` and and passing over?

src/modules/iotjs_module_buffer.c
chokobole: This time I think we need to copy and paste, because it can happen to lose ranges when copying return value of `bound_range` to `start` or `end`.

src/iotjs_util.c
chokobole: As you told, if the value was returned from some int returning function, then it is okay to leave it and cast it when it is needed. But here the thing i worried was that can it make different result because of the difference between `size_t` and `uint32_t` for example on the 64 bit platform?

src/modules/iotjs_module_buffer.c
LaszloLango: Sorry, I'm a bit confused. Where can we lose ranges and what are you suggesting to avoid it?

yichoi:

src/modules/iotjs_module_buffer.c
chokobole: Like here, because both of `iotjs_bufferwrap_length` and `bound_range` returns `size_t`.
```c
 start += (int)iotjs_bufferwrap_length(buffer_wrap);
start = (int)bound_range(start, 0, iotjs_bufferwrap_length(buffer_wrap));
```
So my suggestion is like below. How do you think?
```c
int start = JHANDLER_GET_ARG(0, number);
size_t start_idx = start;

if (start < 0) {
  start_idx +=  iotjs_bufferwrap_length(buffer_wrap);
}
start_idx = bound_range(start_idx , 0, iotjs_bufferwrap_length(buffer_wrap));
```

LaszloLango: `bound_range` is always called with `0` (as a low range). If we change the index from int to size_t, than the index cannot be lower than `0`. But we still have to check that before we call `bound_range`.  So we basically moving out the low range check from `bound_range` and copy everywhere in this file. That means a lot of code duplication which could lead errors later if we must change the code. IMHO it is not a good idea.

chokobole: Considering `bound_range` semantics, yes I think you're right. Then we can trick using `SIZE_MAX`. the first argument of `bound_range` is used as an index, so it can't be `SIZE_MAX`.(Because the length is less than `SIZE_MAX`). So inside the `bound_range`, we can implement like below.
```c
static size_t bound_range(size_t index, size_t low, size_t upper) {
  if (index == SIZE_MAX) {
    return low;
  }  
  if (index > upper) {
    return upper;
  }
  return index;
}
```
Then with few preprocessing, for example here` 
```c
int start = JHANDLER_GET_ARG(0, number);
size_t start_idx;

if (start < 0 && -start > len) {
  start_idx = SIZE_MAX;
} else {
  start_idx = start;
}

start_idx = bound_range(start_idx, 0, iotjs_bufferwrap_length(buffer_wrap));
```
What do you think ??

LaszloLango: Sounds better, I'll try it.

chokobole: 1 more
```c
if (start < 0) {
  if (-start > len) {
    start_idx = SIZE_MAX;
  } else {
    start_idx = len + start;
  }
} else {
  start_idx = start;
}
```

glistening: @chokobole @LaszloLango 
Our documentation of [`Buffer`](https://github.com/Samsung/iotjs/blob/master/docs/api/IoT.js-API-Buffer.md#bufcopytargetbuffertargetstartsourcestartsourceend) says `targetStart`, `sourceStart`, and `sourceEnd` as `Number`. According to document, the type of `start` should be `double`. But I think the documentation is wrong since we will not support such a big range. So first, we have to correct the API document `Number` to `Integer`.  Then, second, we need to define what we will do if the user gives out of Integer range. node.js seems to cast the double number into int64_t if it is not SMI. If I understand correctly, and we agree to follow the style,  finally we can use `int64_t` as type of `start`.

chokobole: Yes, as spec says `integer`'s range will be integers of all integers from -(253 - 1) inclusive to 253 - 1 inclusive, called `safe integer`. So I think it is also okay to accept `start` or `end` as `int64_t` as @glistening says.

glistening:
@LaszloLango This patch includes changes in pwm, i2c, and so on. So I want to get review from @hs0225. Please wait. And could you change the Buffer module API document ? 

```
buf.copy(targetBuffer[,targetStart[,sourceStart[,sourceEnd]]])

    targetBuffer <Buffer> - modified buffer
    targetStart <Number> Default: 0
    sourceStart <Number> Default: 0
    sourceEnd <Number> Default: buf.length
    Returns: <integer> - the number of bytes copied.
```

to

```
buf.copy(targetBuffer[,targetStart[,sourceStart[,sourceEnd]]])

    targetBuffer <Buffer> - modified buffer
    targetStart <Integer> Default: 0
    sourceStart <Integer> Default: 0
    sourceEnd <Integer> Default: buf.length
    Returns: <integer> - the number of bytes copied.
```

hs0225:
@glistening They have no problem.

LaszloLango:
I've updated the PR.

chokobole:
LGTM

src/modules/iotjs_module_buffer.c
```diff
@@ -252,6 +252,17 @@ JHANDLER_FUNCTION(Compare) {
 }
 
 
+#define CONVERT_DOUBLE_TO_SIZE(n, d) \
+  size_t n;                          \
+  do {                               \
+    if (d < 0) {                     \
```
glistening: @LaszloLango Thank you for nice patch. I have probably last question. When d is +inf or -inf, and NaN, how will it work? If you know, let me know. Thank you again.

LaszloLango: It's undefined, so we should handle it. http://stackoverflow.com/questions/3986795/what-is-the-result-of-casting-float-inf-inf-and-nan-to-integer-in-c
Any suggestion? What about using `0`, `INT64_MIN` and `INT64_MAX` for NaN, -inf and inf respectively?

glistening: @LaszloLango Basically I agree with your suggestion. However, I've checked how node.js deal with this situation.

First, I searched Buffer API documentation in Node. It say nothing about NaN, +Inf/-Inf except for `lastIndexOf`, 'IndexOf'. Second, I've done some tests with Node.js. It seemingly work differently for each APIs.

```
> new Buffer(Infinity)
RangeError: Invalid typed array length
    at new Uint8Array (native)
    at createBuffer (buffer.js:66:15)
    at allocate (buffer.js:183:12)
    at Function.Buffer.allocUnsafe (buffer.js:151:10)
    at new Buffer (buffer.js:93:19)
    at repl:1:1
    at REPLServer.defaultEval (repl.js:269:27)
    at bound (domain.js:287:14)
    at REPLServer.runBound [as eval] (domain.js:300:12)
    at REPLServer.<anonymous> (repl.js:439:12)
> new Buffer(-Infinity)
<Buffer >
> new Buffer(NaN)
<Buffer >
> new Buffer(0)
<Buffer >
> var b = new Buffer(8)
undefined
> console.log(b)
<Buffer 78 ee bc c2 83 7f 00 00>
undefined
> var c = new Buffer(8)
undefined
> console.log(c)
<Buffer 80 a5 0c 03 00 00 00 00>
undefined
> b.copy(c, 0, 0)
8
> console.log(c)
<Buffer 78 ee bc c2 83 7f 00 00>
undefined
> b.copy(c, 0, NaN)
8
> c = new Buffer(8)
<Buffer 80 a5 0c 03 00 00 00 00>
> b.copy(c, 0, NaN)
8
> console.log(c)
<Buffer 78 ee bc c2 83 7f 00 00>
undefined
> c = new Buffer(8)
<Buffer 0a 00 00 00 00 00 00 00>
> b.copy(c, 0, -Infinity)
8
> console.log(c)
<Buffer 78 ee bc c2 83 7f 00 00>
undefined
> c = new Buffer(8)
<Buffer 08 a6 0c 03 00 00 00 00>
> b.copy(c, 0, 0, 2)
2
> console.log(c)
<Buffer 78 ee 0c 03 00 00 00 00>
undefined
> b.copy(c, 0, 0, Infinity)
0
> console.log(c)
<Buffer 78 ee 0c 03 00 00 00 00>
undefined
> b.copy(c, 0, Infinity)
8
> console.log(c)
<Buffer 78 ee bc c2 83 7f 00 00>
```

I suggest `0`, `0`, `SIZE_MAX` for NaN, -inf and inf respectively. I've changed your suggestion slightly because `CONVERT_DOUBLE_TO_SIZE` returns `size_t`.

src/modules/iotjs_module_buffer.c
```diff
@@ -252,6 +252,17 @@ JHANDLER_FUNCTION(Compare) {
 }
 
 
+#define CONVERT_DOUBLE_TO_SIZE(n, d) \
+  size_t n;                          \
```
glistening: @LaszloLango IMHO, what about moving the declaration out of this macro? Let the user to define their variable before `CONVERT_DOUBLE_TO_SIZE'. 

From
```
CONVERT_TO_DOUBLE_TO_SIZE(dst_start, 0, dst_length);
```

to

```
size_t dst_start;
CONVERT_TO_DOUBLE_TO_SIZE(dst_start, 0, dst_length);
```

Because I was confused where is the declaration of `dst_start`, `src_start`, and so on. I am not used to this style. Moreover, automatically declared variable in `CONVERT_DOUBLE_TO_SIZE` remains after `CONVERT_DOUBLE_TO_SIZE`.

Or alternatively, what about changing the name like `DECLARE_SIZE_T_FROM_DOUBLE` because it does not simply convert, but also declare variable.

LaszloLango:
@glistening, PR is updated. Please check.

PR #801, opened 2017-04-28
Remove unimplemented functions in fs.js

DanielBallaSZTE:
fstat is not implemented in the C source file, causing crashes. fstatSync is also causing crashes, since it returns the missing builtin function fstat.
I suggest removing them from the code.

### DIFF
--- a/src/js/fs.js
+++ b/src/js/fs.js
@@ -101,17 +101,6 @@ fs.statSync = function(path) {
 };
 
 
-fs.fstat = function(fd, callback) {
-  fsBuiltin.fstat(checkArgNumber(fd, 'fd'),
-                  checkArgFunction(callback, 'callback'));
-};
-
-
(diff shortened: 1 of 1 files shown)


glistening:
@DanielBallaSZTE Thank you for informing us. I think fstat is almost implemented. libtuv fstat support, javascript implementation, and the document is already ready. So I would like to choose to complete the remaining work. (e.g., adding binding iotjs_module_fs.c and test case in test/run_pass).

DanielBallaSZTE:
@glistening So do you want me to close the pull request, as the stated features are going to be implemented later on?

glistening:
@DanielBallaSZTE Oh, Thank you, please close this PR. And we can't start to implement `fs.stat` immediately for other tasks, what about implementing `fs.stat` by yourself if you would like to do ?

PR #805, opened 2017-05-02
Followup patch for the Jerry-debugger

bzsolt:
After jerryscript#1734 we don't need to add
the debugger-related defines at hostjerry build.

This patch also get rid of some remnants of a previous version of #722.


yichoi:
LGTM

PR #806, opened 2017-05-02
Additional error events for socket connect failures

vprunet:
I noticed that some synchronous socket connection errors were ignored. 
This happens for example with ENOMEM errors (STM32 on nuttx, when the maximum number of allowed sockets is reached).

The patch gets the error code and generates a JavaScript error event on the socket.
I have put the `socket.emit` call within a `process.nextTick` to allow for late declarations of event handlers. Unfortunately, on nuttx this does not work as I would expect: the event is fired immediately. Did I misuse the API or should I file a bug for that?


glistening:
@vprunet Thank you for your patch. It seems the DCO is missing in commit message. We check DCO in commit message, not in PR description.

vprunet:
@glistening Thanks. I've put the DCO in the commit message and rebased the commit.

src/js/net.js
```diff
@@ -279,7 +279,11 @@ function connect(socket, ip, port) {
     }
   };
 
-  socket._handle.connect(ip, port, afterConnect);
+  var err = socket._handle.connect(ip, port, afterConnect);
+  if (err) {
+    emitError(socket, new Error('connect failed - status: ' +
+      TCP.errname(err)));
+  }
```
chokobole: I think the function `emitError` is the function to emit error instantly and I think the meaning `it is done at the next tick` is not meant. If you want to try to do that, you have to do yourself like below I think.
```
if (err) {
  process.nextTick(function() {
    socket.emit('error',  new Error('connect failed - status: ' + TCP.errname(err)));
  }
}
```

vprunet: @chokobole ``EmitError`` is a function defined locally in the same file and used only twice. I've patched this function to call ``process.nextTick`` (same code as your suggestion, but in a function). Both uses of the function are consistent with a deferred event propagation.

chokobole: @vprunet You mean 2 local uses should be deferred? Then I am worried the uses of the `emit*` outside this module like `src/js/stream_writable.js` is a bit different from here. The rests uses propagate event instantly, not deferred.

vprunet:
Actually, ``process.nextTick`` is not required because the events are fired within the ``dns.lookup``callback or even later. 
I will open another issue to discuss a problem I found with dns.lookup which explains why some events are triggered earlier than expected, and potentially not handled.

yichoi:
@vprunet Please check your commit. Your 2 commits are supposed to be one single commit

yichoi:
LGTM

PR #808, opened 2017-05-03
[Code Coverage] Improving net.js coverage

DanielBallaSZTE:
Improve the net.js coverage, some parts of the code can not be covered.
Also, fix a typo in net.js at line 123.

### DIFF
--- a/src/js/net.js
+++ b/src/js/net.js
@@ -115,12 +115,15 @@ Socket.prototype.connect = function() {
 
   self._host = host;
   dns.lookup(host, dnsopts, function(err, ip, family) {
+    if (self._socketState.destroyed) {
+      return;
+    }
     self.emit('lookup', err, ip, family);
 
     if (err) {
       process.nextTick(function() {
         self.emit('error', err);
-        self._destroy();
+        self.destroy();
       });
     } else {
       resetSocketTimeout(self);
@@ -493,6 +496,7 @@ Server.prototype.listen = function() {
   var err = self._handle.bind(host, port);
   if (err) {
     self._handle.close();
+    self._handle = null;
     return err;
   }
 
@@ -505,6 +509,7 @@ Server.prototype.listen = function() {
 
   if (err) {
     self._handle.close();
+    self._handle = null;
     return err;
   }
 
--- a/test/run_pass/test_net_11.js
+++ b/test/run_pass/test_net_11.js
@@ -1,0 +1,98 @@
+/* Copyright 2015-present Samsung Electronics Co., Ltd. and other contributors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS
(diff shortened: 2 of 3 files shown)


DanielBallaSZTE:
`Assertion failed: (!uv__io_active(&stream->io_watcher, POLLIN | POLLOUT)), function uv__stream_destroy, file /Users/travis/build/Samsung/iotjs/deps/libtuv/src/unix/stream.c, line 461.`

Got the same error when using `socket.destroy()` or `socket.destroySoon()`

yichoi:
It would be better to fix the assertion error and include this patch.

DanielBallaSZTE:
There was an issue in net.js when a socket connect was called, the dns lookup was being called asynchronously, allowing a special case where the socket could be destroyed before the dns lookup was finished. Therefore, later on if it tried to connect, the connect(self, ip, port) is being called without the socket even existing, causing a crash.
A special testcase was added to the test covering this exact issue.

DanielBallaSZTE:
Seems like that didn't solve it.

DanielBallaSZTE:
At the moment I have no OSX environment to find the solution to the error, can't reproduce it on Linux.

glistening:
@DanielBallaSZTE Do you mean above testcase `test_net_11.js` fails on osx only, not reproduced on linux?

DanielBallaSZTE:
@glistening Yes, I could not reproduce it on linux, and it fails on OSX only.

akosthekiss:
I've managed to track this down. The minimal test case that replicates the problem is (comments by me):
```js
var net = require('net');

var server = net.createServer();
server.listen(22701, 5); // first listen succeeds

var server4 = net.createServer();
server4.listen(22701, 5); // second listen does not succeed and calls close() automatically
server4.listen(22701, "127.0.0.1"); // third listen succeeds(?)

server.close();  // close succeeds
server4.close(); // close thinks that "Attempt to close uninitialized or already closed handle"
```
This is because in src/js/net.js around line 490:
```js
  if (!self._handle) {
    self._handle = createTCP();
  }
```
but later on (twice) we only have:
```js
  if (err) {
    self._handle.close();
    return err;
  }
```
However, if we look at `Server.prototype.close`, we can see that the proper way of closing is to null out the handle:
```js
  if (this._handle) {
    this._handle.close();
    this._handle = null;
  }
```
So, in the example, the failing second `listen` leaves a closed handle in the `server4` object, which will not be replaced by the third `listen`. And not completely sure how, but this closed handle will be reused by the third `listen`. So the final `close` will somehow get to `iotjs_handlewrap_close` where it finds an already closed but somehow still active handle, and do nothing (will not try to close it again as it thinks it's already closed). But that's bad. Because the final `uv__run_closing_handles`, `uv__finish_close`, `uv__stream_destroy` function call chain does not expect an active handle and asserts out.

There may be multiple possible fixes to the problem, I guess. For me, it worked to add `self._handle = null;` after the `self._handle.close();` calls in `Server.prototype.close`. Perhaps replacing `self._handle.close();` with `self.close();` would also work, but I didn't try. (Digging to the root of how can a stream be closed and active at the same time might also be interesting, but didn't do that either.)

PS: As nulling out the handle seems to be a must, I have no idea why this issue never turned up on Linux or anywhere else.

src/js/net.js
```diff
@@ -115,12 +115,15 @@ Socket.prototype.connect = function() {
 
   self._host = host;
   dns.lookup(host, dnsopts, function(err, ip, family) {
+    if (self._socketState.destroyed){
```
zherczeg: Space before {

hs0225:
LGTM

Issue #798, opened 2017-04-28
Use array for headers in http_outgoing and http_incoming


daeyeon:
`Object` was used as substitute for the lack of `Array` features. #776 was a fix to remove one of the cases. 

In #797, it's discussed if replacing the type of `_header` in `OutgoingMessage` with `Array` is needed.

In my brief study, it's a private property that stores HTTP headers for In/OutMessage, and the type of user's input to be saved is `Object`. It may or may not be better to replace it with `Array`. I'd like to hear opinions.

daeyeon:
While discussing `src/js/http_outgoing.js` on #833, @knightburton also agreed that `Object` is ok for `OutgoingMessage`. If there is no other opinion, I'd like to close this for now.

PR #810, opened 2017-05-03
Avoid manual argument copy, use Array slice instead in events.js

galpeter:
To remove the first argument form the arguments object previously
a manual copy was used. This can be replaced by an Array.slice call.


yichoi:
It would be great to use Array where it could be replaced with at further PRs

yichoi:
related with #798

PR #811, opened 2017-05-03
[WIP][Code Coverage] Improve coverage of module.js

DanielBallaSZTE:
Some parts of the code are not possible to be covered, since the test requires the module.js file at the beginning with the given environment variables and cwd, which can not be changed later on since requiring the module adds it to the require.cache, so it wouldn't be required again.

### DIFF
--- a/test/run_pass/test_module_cache.js
+++ b/test/run_pass/test_module_cache.js
@@ -20,3 +20,16 @@ module_cache.i = 100;
 module_cache = require('run_pass/require1/module_cache.js');
 
 assert.equal(module_cache.i, 100);
+
+{
+  var test1 = require('run_pass/require1/node_modules/test_pkg');
+  var test2 = require('run_pass/require1/node_modules/test_index');
+}
+
+assert.throws(function() {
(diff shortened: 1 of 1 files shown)


DanielBallaSZTE:
Fixing module paths right now.

chokobole:
I don't know what exactly you are trying to cover? Since reading your PR, I expected some statements  like `require(module)` to try to manage re-load `module.js`. But the things i can see are to add some tests to throw error when there is no module with the given path.

DanielBallaSZTE:
You can not reload module.js even with clearing require.cache, which has been stated before. Therefore those functions can not be tested. What I did was covering line 74 to 94 in module.js

chokobole:
Yes I understood `test1` to `test4`  are tests to cover line 74 to 94 in module.js, But still what is consuing me is with this reason `Therefore those functions can not be tested`, so you added something new? or literally you couldn't ?

DanielBallaSZTE:
I literally could not test re-requiring module.js.
If you have any tips in clearing the cache and re-requiring it, I am all ears.

yichoi:
LGTM

PR #812, opened 2017-05-03
Test case for the Event Emitter

galpeter:
Test if an event listener for a 'once' call can be removed.


yichoi:
LGTM

PR #815, opened 2017-05-03
Fix include order for newer clang-format

galpeter:
The clang-fromat (3.8+) also checks #include order to
make sure it is in alphabetical order.


galpeter:
great... newer vs older clang-format does things differently.

PR #817, opened 2017-05-05
make dns.lookup asynchonous (fix for nuttx)

vprunet:
On nuttx, I noticed that some net and dns events get fired too early, before the call stack has completed execution and all handlers are registered.

On nuttx, ``dnsBuiltin.getaddrinfo`` synchronously calls ``iotjs_make_callback``, with the effect of running the lookup callback synchronously and then processing the nextTick queue.
As a result, the ``lookup`` event is fired immediately after the callback is run. This has also indirect consequences on the ``net.connect``events which are also fired too early.

I think that this is an improper use of ``iotjs_make_callback`` which should only be called from asynchronous C callbacks ( such as libuv callbacks) and never from synchronous C code.

The proposed fix wraps the call to ``dnsBuiltin.getaddrinfo`` into a ``process.nextTick`` to ensure that the code will run its own tick.

IoT.js-DCO-1.0-Signed-off-by: Vincent Prunet vincent.prunet@gmail.com

### DIFF
--- a/src/js/dns.js
+++ b/src/js/dns.js
@@ -59,19 +59,31 @@ exports.lookup = function lookup(hostname, options, callback) {
   if (family !== 0 && family !== 4 && family !== 6)
     throw new TypeError('invalid argument: family must be 4 or 6');
 
-   process.nextTick(function() {
-     var err = dnsBuiltin.getaddrinfo(
-       hostname,
-       family,
-       hints,
-       function(err, address, family) {
-         var errObj = null;
-         if (err) {
-           errObj = dnsException(err, 'getaddrinfo', hostname);
-         }
-       callback(errObj, address, family);
-       });
-   });
+  function getaddrinfo() {
+    var err = dnsBuiltin.getaddrinfo(
+      hostname,
+      family,
+      hints,
+      function(err, address, family) {
+        var errObj = null;
+        if (err) {
+          errObj = dnsException(err, 'getaddrinfo', hostname);
+        }
+        callback(errObj, address, family);
+      });
+      if (err) {
+        callback(dnsException(err, 'getaddrinfo', hostname), address, family);
+      }
+    }
+  if (process.platform != 'nuttx' && process.platform != 'tizenrt') {
+    getaddrinfo();
+  } else {
+    // dnsBuiltin.getaddrinfo is synchronous on these platforms.
+    // needs to be wrapped into an asynchronous call.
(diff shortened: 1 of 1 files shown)


chokobole:
Ooh, yes, you're right. Just one thing I am worried is that deferring `dnsBuiltin.getaddrinfo` results in new side effect, we didn't expect before? For example, 
```
(1) dns.lookup // about 20ms consuming
JSCodes ...
(2) other async function needs io poll // about 20 ms consuming
JSCodes ... // about 40 ms consuming
```
Let's limit to not nuttx environment. Above codes before this commit, user expects (2) is called before (1) is called. AFAIK, because in `uv_loop event queue`, (1) and (2) will be run on the `io poll` stage, then (2) and (1) will compete to run earlier. But after this commit, (1) is definitely run on `process.nextTick` so (1) is always run after (2).  Does this kind of thing really bother users?

glistening:
@chokobole 

> (1) is definitely run on process.nextTick so (1) is always run after (2).

What do you mean (1) in here? `dns.lookup`? or callback of `dns.lookup`?

vprunet:
@chokobole @glistening 
Thank you for your reviews.

I've updated the patch so that:
- the call to ``process.nextTick`` is done only on nuttx and tizenrt.
- possible synchronous errors are handled in the ``dns.lookup``callback, as ``dns.lookup``is not supposed to return a value (and definitely could not in the nuttx and tizenrt cases). Previous patch was ignoring synchronous errors.

Question: what is the recommended best practice (git command) to merge two commits into one without breaking the discussion thread?

chokobole:
@glistening  Sorry for delay. Forget about the example above, instead let's think with this problem then. This is what I am worried. 
```javascript
var a = 5;
setTimeout(function() {
  a = 3;
}, 0);

dns.lookup ('localhost', 4, function(err, ip, family) {
  console.log(a);  // Here, will it be 3 or 5?
});
```
If `dns.lookup` is handled with `process.nextTick`, then it is always print `5`. Because after running this script, `Timeout` handler is registered and  `dns.lookup` is queued to `process.nextTick` queues. And for the next looping, callback function queued at `process.nextTick` are run first then `timeout` handler.  But originally it can be `3` or `5`.  because both of the callback functions are competing at `io poll` stage.

chokobole:
@vprunet  what about `git commit --amend && git push -f`?

yichoi:
LGTM

Issue #725, opened 2017-04-06
High memory consumption in 'setTimeout'


LaszloLango:
`setTimeout` calls consume too much memory in IoT.js. I've investigated the issue and here are my experiences. The net module use `resetSocketTimeout()` frequently. For example in `test_net3.js` the setTimeout is called 828 times. 

JS backtrace of `Timeout.prototype.activate`:
```
Frame 0: timers:43 (in function() at line:42, col:30)
Frame 1: timers:90 (in function() at line:68, col:22)
Frame 2: iotjs.js:48 (in function() at line:46, col:25)
Frame 3: net:308 (in resetSocketTimeout() at line:302, col:1)
Frame 4: net:98 (in function() at line:77, col:28)
Frame 5: net:642 (in function() at line:639, col:46)
Frame 6: test/run_pass/test_net3.js:70 (in function() at line:1, col:2)
Frame 7: module:188 (in function() at line:180, col:36)
Frame 8: module:168 (in function() at line:152, col:23)
Frame 9: module:193 (in function() at line:192, col:26)
Frame 10: iotjs.js:31 (in startIoTjs() at line:23, col:3)
Frame 11: iotjs.js:221 (in function() at line:17, col:2)
```

Each `setTimeout` call allocate 176 byte in release and 232 byte in debug build. This means 140K peak of timer calls on `test_net3.js`. One big problem is that the timer handler structure in libtuv is too big. It isn't optimized for low memory environments. For example it has a reserved field that wastes a lot of memory.
The other problem is that this handler is only freed when the JS wrapper object is freed.

Is it OK to do complex reworks in `libtuv` or not? Do we plan to upstream `libuv` changes frequently? Do you have suggestion how to reduce the memory consumption of timers?

zherczeg:
Probably it could help if we could delete the timer when it is cleared. Unfortunately there is a delayed delete, which unreferences the object, but does not call its native destructor.

Perhaps it could help if the native part of such objects would be deleted as well.

glistening:
@LaszloLango For latter, please give us some time for investigating this issue. Probably we can fix the former first, because it is simpler than latter and independent problem.

glistening:
@LaszloLango 
1) _Is it OK to do complex reworks in libtuv or not?_ 
OK. If it is necessary to reduce memory consumption.
2) _Do we plan to upstream libuv changes frequently?_ 
Presently, no. We have other high priority tasks.
[...]

PR #818, opened 2017-05-05
Optimize memory usage of iotjs_timerwrap_t.

LaszloLango:
Results on RPi2, `test_net3.js`:
**before**
```
[1] Peak: 51.3Kb (52576b) Allocs: 5738 Reallocs: 0 Total: 186.7Kb (191188b) 
Group: iotjs
  [2] Peak: 47.5Kb (48640b) Allocs: 828 Reallocs: 0 Total: 64.6Kb (66240b) 
  Group: timer
```
**after**
```
[1] Peak: 48.9Kb (50152b) Allocs: 6566 Reallocs: 0 Total: 183.4Kb (187876b) 
  Group: iotjs
    [2] Peak: 45.0Kb (46132b) Allocs: 1656 Reallocs: 0 Total: 61.4Kb (62928b) 
    Group: timer
```

Related issue: #725

### DIFF
--- a/src/modules/iotjs_module_timer.c
+++ b/src/modules/iotjs_module_timer.c
@@ -24,15 +24,15 @@ IOTJS_DEFINE_NATIVE_HANDLE_INFO(timerwrap);
 
 iotjs_timerwrap_t* iotjs_timerwrap_create(const iotjs_jval_t* jtimer) {
   iotjs_timerwrap_t* timerwrap = IOTJS_ALLOC(iotjs_timerwrap_t);
+  uv_timer_t* uv_timer = IOTJS_ALLOC(uv_timer_t);
   IOTJS_VALIDATED_STRUCT_CONSTRUCTOR(iotjs_timerwrap_t, timerwrap);
 
   iotjs_handlewrap_initialize(&_this->handlewrap, jtimer,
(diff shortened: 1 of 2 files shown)


glistening:
@LaszloLango In my understanding, it separates the allocation of `uv_timer_t` from `iotjs_timerwrap_t`. So there is increase of the number of allocation, ( I'm a little afraid of defragmentation. ) And I fail to grasp the reason of reduce of total and peak memory. Could you explain why it reduce the memory consumption?

LaszloLango:
@glistening the fixes the "delayed free" mentioned by @zherczeg in #725. The problem is that the UV handler object is only freed when the GC free the JS object. Because the uv_handler is not part of the JS heap this leads to a high peak value. The other problem is that the handler is duplicated in the IoT.js. The handler wrapper stores a pointer and the extended object stores the uv_handler object. Even if decide to handle JS object and UV handlers together this should be solved. IMHO it is better to separating the life cycle of libUV and JS objects.

zherczeg:
I think the motivation is that if a cancelled timer is referenced from somewhere, it also keeps the data of the cancelled UV timer, which will never be used anymore. So it is good if we can free the internal part of an object without freeing the object. This idea is not limited to timers.

yichoi:
LGTM

PR #823, opened 2017-05-10
Use absolute path for the file system resources in tests

robertsipka:
NuttX requires absolute path for file operations.

### DIFF
--- a/test/run_pass/test_fs_event.js
+++ b/test/run_pass/test_fs_event.js
@@ -19,8 +19,8 @@ var fs = require('fs');
 var assert = require('assert');
 
 
-var srcFilePath = "resources/test1.txt";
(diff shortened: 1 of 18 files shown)


glistening:
@robertsipka Could you also change test/test_stat.js? I don't know why this testcase is in test root. Could you also move this file to run_pass directory? If it does not work on NuttX, please modify testsets.json as skip for NuttX.

robertsipka:
@glistening : Thanks, I've modified this testcase. The `test_stat.js` contains dynamically changing data, because it synchronously calls the operating system's stat() function on the `test` folder, that hopefully will grow. Therefore, this test should be skip on every platform.

yichoi:
LGTM

PR #824, opened 2017-05-10
Remove 'parser=this' constructs from http_common.js

galpeter:


hs0225:
Because of code readability, `this` seems to be assigned as a `parser`. This code refers to node.js. If for no other reasons, I think the current code is better now.

galpeter:
Removing the `var parser=this` reduces the bytecode size a bit.

hs0225:
LGTM

PR #826, opened 2017-05-10
Remove unused require calls from http server module

galpeter:
EventEmmitter, HTTPParser, IncomingMessage, and Buffer
modules were not used in the http_server.js module so
just remove them.

### DIFF
--- a/src/js/http_server.js
+++ b/src/js/http_server.js
@@ -13,13 +13,9 @@
  * limitations under the License.
  */
 
-var EventEmitter = require('events');
 var util = require('util');
 var net = require('net');
-var HTTPParser = process.binding(process.binding.httpparser).HTTPParser;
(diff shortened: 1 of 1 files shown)


glistening:
@galpeter I think #825 and #826 can be merged into one PR. It removes unused require call in http module. Could you merge these two PR? I think #824 is also similar one in that it removes unnecessary code in http module.

zherczeg:
Patch is ok for me.

yichoi:
LGTM

yichoi:
Please consider @glistening 's above opinion at next time.

galpeter:
@yichoi, @glistening sorry for the late answering, just got round to reply on this pr. I'll keep in mind your comments. The #824 issue was a different kind so merging it into this one could be a bit misleading.

PR #827, opened 2017-05-10
Fix build type parameter for js2c call

galpeter:
Previously even in debug mode the js2c was called
with the --buildtype=release option. For snapshot
mode this does nothing however for a no-snapshot build
the js files were always stripped of empty lines and commends
thus making debugging js code a bit harder as the line infos
were off.

By correctly detecting the build type in the CMAKE files
the buildtype problem will be resolved.


galpeter:
@glistening I've update the PR.

PR #828, opened 2017-05-10
Avoid creating anonymous functions in stream writable js module

galpeter:


galpeter:
Two reasons:
1) by removing the anonymous functions the bytecode size is reduced a bit.
2) having an anonymous function refer to an outer variable could result in having all of the parent function's variables in the lexical scope to be alive till the anonymous function is "freed". By removing we can reduce the runtime memory usage a bit.

hs0225:
Okay. Thanks for your answers. LGTM

yichoi:
LGTM

PR #829, opened 2017-05-11
Hide the native binding variable in GPIO module

hs0225:


yichoi:
LGTM

PR #830, opened 2017-05-11
[Code coverage] Test the double response in http.

knightburton:
Note: The http_client.js modification just a workaround because
the httpparser asserting otherwise. Working on the followup.


yichoi:
LGTM

PR #832, opened 2017-05-11
[Code coverage] Improve the http client module coverage.

knightburton:
- Add a test file to cover the client parse error.
 - Add a test file to cover the client socket error.

### DIFF
--- a/test/run_pass/test_net_httpclient_error.js
+++ b/test/run_pass/test_net_httpclient_error.js
@@ -1,0 +1,42 @@
+/* Copyright 2017-present Samsung Electronics Co., Ltd. and other contributors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+var assert = require('assert');
+var http = require('http');
+
+var recievedResponse = false;
+
+var server = http.createServer(function(request, response) {
+  server.close();
+});
+
+server.listen(3085, 'localhost');
+
+// Connection refusal: The request will try to connect to a server,
+// however the connection can not be created because the port is invalid.
+// Therefore the client will destroy the socket without an error emit.
+http.request({
+  host: 'localhost',
+  port: 3089,
+  path: '/',
+  method: 'GET'
+}, function(response) {
+  response.on('end', function() {
+    recievedResponse = true;
+  });
+}).end();
+
+process.on('exit', function() {
+  assert.equal(recievedResponse, false);
+});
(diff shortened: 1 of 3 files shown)


test/testsets.json
```diff
@@ -49,6 +49,8 @@
     { "name": "test_net_9.js" },
     { "name": "test_net_10.js" },
     { "name": "test_net_headers.js", "skip": ["nuttx"], "reason": "not implemented for nuttx" },
+    { "name": "test_net_http_client_error.js", "timeout": 10, "skip": ["nuttx"], "reason": "not implemented for nuttx" },
+    { "name": "test_net_http_client_parse_error.js", "timeout": 10, "skip": ["nuttx"], "reason": "not implemented for nuttx" },
```
knightburton: Done.

test/run_pass/test_net_http_client_parse_error.js
```diff
@@ -0,0 +1,50 @@
...
+    request.on('error', function(err) {
+      errors++;
+    });
+    request.end();
+  }
+});
+
+process.on('exit', function() {
+  assert.strictEqual(connects, 2);
+  assert.strictEqual(errors, 2);
```
knightburton: I've wrote a comment into the code.

yichoi:
@knightburton please rebase it to the master

knightburton:
@yichoi rebased

test/run_pass/test_net_httpclient_error.js
```diff
@@ -0,0 +1,42 @@
...
+
+var server = http.createServer(function(request, response) {
+  server.close();
+});
+
+server.listen(3085, 'localhost');
+
+// Connection refusal: The request will try to connect to a server,
+// however the connection can not be created because the port is invalid.
+// Therefore the client will destroy the socket without an error emit.
```
daeyeon: I think that we need to treat this as an issue. Because user should be notified that the connection is refused. For being consistent with node, the way to inform needs to be exception error. If agreed, can you make an issue for this?

test/run_pass/test_net_httpclient_parse_error.js
```diff
@@ -0,0 +1,53 @@
...
+    var request = http.request({
+      host: '127.0.0.1',
+      port: 3085,
+      method: 'GET',
+      path: '/'
+    });
+    // The first request will get a parse error while trying to parse the data,
+    // therefor the client will destroy the socket. After the socket is
+    // destroyed the second request will get a socket hang up error.
+    request.on('error', function(err) {
```
daeyeon: In my perspective, the first request will NOT get a parse error since the server sends a response with the proper header. It seems successfully ended. So I expect socket hang up error should not occur at the second response. Please clarify my query.

test/run_pass/test_net_httpclient_error.js
knightburton: I agree, this is not a user friendly state right now. Also I think, this test is okay at this moment. I'll create an issue for this later.

test/run_pass/test_net_httpclient_parse_error.js
knightburton: Well after a little investigation, this is working different as a thought and this test is unnecessary complicated. Let me update this.

knightburton:
@daeyeon  The server response will trigger the Parse Error (this is the first error), therefore the socket will be destroyed but I think the client will keep trying to listening. Of course that never gets a response and this is the reason of the second error (socket hang up).

daeyeon:
LGTM

test/run_pass/test_net_httpclient_parse_error.js
```diff
@@ -0,0 +1,45 @@
...
+  method: 'GET',
+  path: '/'
+});
+request.on('error', function(err) {
+  errors++;
+});
+request.end();
+
+process.on('exit', function() {
+  // The first errpor is a Parse Error.
```
hs0225: `errpor` is a typo.

knightburton: :(, fixed

yichoi:
LGTM

PR #833, opened 2017-05-11
[Code coverage] Improve the http module coverage.

knightburton:
- Add new test cases to the existing files.
- Add new test file to cover the request response messages.

Note: The setTimeout function from the outgoungMessage is deleted,
because every parents are implements their own setTimeout function.

### DIFF
--- a/test/run_pass/test_net_headers.js
+++ b/test/run_pass/test_net_headers.js
@@ -29,6 +29,8 @@ var server = http.createServer(function (req, res) {
   });
 
   var endHandler = function () {
+    // this should return.
+    res.removeHeader('h1');
 
     res.setHeader('h1','h1');
     res.setHeader('h2','h2');
@@ -39,8 +41,13 @@ var server = http.createServer(function (req, res) {
     }
     // final res.headers = { 'h1' : 'h1', 'h3': 'h3prime' }
 
+    // Large header on response.
+    for (var i = 0; i < 500; i++) {
+      res.setHeader('h' + (5 + i), 'h' + (5 + i));
+    }
+
     res.end(function(){
-        server.close();
+      server.close();
     });
   };
 
--- a/test/run_pass/test_net_http_request_response.js
+++ b/test/run_pass/test_net_http_request_response.js
@@ -1,0 +1,195 @@
+/* Copyright 2017-present Samsung Electronics Co., Ltd. and other contributors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+var assert = require('assert');
+var http = require('http');
+var net = require('net');
+
+// Messages for further requests.
+var message = 'Hello IoT.js';
+
+// Options for further requests.
+var options = {
+  method: 'POST',
+  port: 3005,
+  path: '/',
+  headers : {'Content-Length': message.length}
+};
+
+var server1 = http.createServer(function(request, response) {
+  var str = '';
+
+  request.on('data', function (chunk) {
+    str += chunk.toString();
+  });
+
+  request.on('end', function() {
+    assert.equal(str, message);
+    response.end();
+  });
+});
+server1.listen(3005, 5);
+
+// Simple request with valid utf-8 message.
+var isRequest1Finished = false;
+var request1 = http.request(options, function(response) {
+  var str = '';
+
+  response.on('data', function(chunk) {
+    str += chunk.toString();
+  });
+
+  response.on('end', function() {
+    isRequest1Finished = true;
+    server1.close();
+  });
+});
+request1.end(message, 'utf-8');
+
+
+var server2 = http.createServer(function(request, response) {
+  response.end();
+});
+server2.listen(3006, 5);
+
+// Simple request with end callback.
+var isRequest2Finished = false;
+options.port = 3006;
+var request2 = http.request(options);
+request2.end(message, function() {
+  server2.close();
+  isRequest2Finished = true;
+});
+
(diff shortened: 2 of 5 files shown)


yichoi:
@knightburton Please rebase it to the master

test/run_pass/test_net_httpserver_timeout.js
```diff
@@ -25,6 +25,8 @@ var server = http.createServer(function(req, res) {
 });
 
 server.listen(3003);
+// Overwrite the default timeout value.
+server.setTimeout(100);
```
daeyeon: There is no verification for this test.

src/js/http_outgoing.js
```diff
@@ -152,8 +152,8 @@ OutgoingMessage.prototype._storeHeader = function(statusLine) {
 
 OutgoingMessage.prototype.setHeader = function(name, value) {
 
-  if (this._headers === null) {
-    this._headers = {};
+  if (util.isNull(this._headers)) {
+    this._headers = [];
```
daeyeon: First of all, this commit is to improve code coverage. So please don't fix any original code. Plus, I don't agree this change. Kindly let me know why using `Array` is needed. Seeing the code below, `this._headers` is still used as `Object` although you set it as `Array`.

test/run_pass/test_net_httpserver_timeout.js
knightburton: I've added an assert below this line.

src/js/http_outgoing.js
knightburton: You are right! I will separate the coverage and fix in the future. The other hand, you are right again, there is no point to use array instead of object, this was my inattention.

test/run_pass/test_net_httpserver_timeout.js
daeyeon: I suggest the test like below.

```
var timer = setTimeout(function () {
    defaultTimeout = true;
}, server.timeout);

server.setTimeout(XXX, function(socket) {
    ....
    assert.equal(defaultTimeout, false);
    clearTimeout(timer);
```

yichoi:
@knightburton please rebase it to the master

src/js/http_outgoing.js
daeyeon: #798 is related to this discussion.

knightburton:
@daeyeon I've updated the PR. I would like to know what do you think about it. First of all, I've removed the code fixes (I'll update them in a separate PR later), and isolated every test case into a server-request pair in the request-response test file. In the other hand, I removed the server default timeout overwrite test for now, because I think neither of our solution is not the best. I would like to update this patch with a proper timeout test and invalid utf-8 test later.

test/run_pass/test_net_http_request_response.js
```diff
@@ -0,0 +1,206 @@
...
+// the default port is 80 and the default method is GET.
+var request0 = http.get(function(response) {
+  var str = '';
+
+  response.on('data', function(chunk) {
+    str += chunk;
+  });
+
+  response.on('end', function() {
+    assert.equal(str, 'default');
```
daeyeon: This seems too natural connection failure case to test. Do we need to test if our api doesn't work when user doesn't fill up the mandatory field for the api?

test/run_pass/test_net_http_request_response.js
```diff
@@ -0,0 +1,206 @@
...
+});
+server4.listen(3008, 5);
+
+var request4 = http.request({
+  method: 'HEAD',
+  port: 3008,
+  path: '/'
+}, function(response) {
+  response.on('end', function() {
+    assert.equal(response.statusCode, 200);
```
daeyeon: Please clearly use `assert` on what you check for each test. Since this compares the status code, it seems to assume that the test needs any data from server. I suggest the below pattern. The same problem is found elsewhere in this code.

```
var isTestCase1Called = false;
var request1 = http.get(function(res) {
  ....
  res.on('end', function () {
      isTestCase1Called = true;
  });
  ....
});


process.on('exit', function () {
  assert.equal(isTestCase1Called, false);
});
``

File: test/run_pass/test_net_http_request_response.js (hunk 3 above)
Comment (knightburton) [inline test/run_pass/test_net_http_request_response.js]: Yes, I think it is necessary. I think we should test the default, natural cases. too.

File: test/run_pass/test_net_http_request_response.js (hunk 4 above)
Comment (knightburton) [inline test/run_pass/test_net_http_request_response.js]: I understand what are you suggest, but also I think this test case is fine because the communication should be OK without any body data and that is why there is the status code check.

File: test/run_pass/test_net_headers.js [hunk 5]
```diff
@@ -29,6 +29,7 @@ var server = http.createServer(function (req, res) {
   });
 
   var endHandler = function () {
+    res.removeHeader('h1'); // this should return.
```
zherczeg: Put comment before this line.

test/run_pass/test_net_http_request_response.js
daeyeon: In the case, it seems right to check if the api doesn't crash, or else if it throws any information for the wrong usage. However, assert here checks str although it's even never be called. Confusing.

daeyeon: You maybe misunderstood `options`. It's mandatory, so that the api doesn't need to gurantee the case where user doesn't fill it up. If the given `options` is just an object having nothing, the api fill it up with default values. https://github.com/Samsung/iotjs/blob/master/docs/api/IoT.js-API-HTTP.md#httpgetoptions-callback

zherczeg: Should it throw an error if no options passed then?

knightburton: No, it should not. It is just fills up the options values with default parameters. That is what i've tested, if we can't define any options value, the connection should be fine.

daeyeon: Throwing an error is needed. e.g) calling `server0.listen()` without the first mandatory param, port, throws an error. The criteria is the API reference, not the current observation.

knightburton: Sorry for the late response, then what should I do with this test? Should I remove it for now?

daeyeon: Yes, please remove this test for now. It'd be better to treat it on another commit for changing the condition of `options` to optional.

knightburton: Done, removed the test.

daeyeon:
LGTM

Issue #816, opened 2017-05-05
Internal types are not checked by IoT.js


zherczeg:
This is currently crashing:

    var Buffer = require('buffer');

    var buf = new Buffer("ABCDEF");

    var o = { _builtin: { compare:buf._builtin.compare }, compare: buf.compare };
    o.compare(buf);

We are creating a _builtin object which is a normal object, and this object has no internal property. It is even more dangerous if we use an object created by another module (and has an internal property to a different type).

Question: do we want to do something with it, or this is a misuse of the API?

LaszloLango:
I think this is bug. The internal objects must be hidden. `_onNextTick` and `_onUncaughtException` also available in the global object. This might caused by #748.

FYI, sample code to find internal objects:
```javascript
var keys = Object.keys(global);

for(var i=0; i < keys.length; i++)
{
  var str = /^_/.exec(keys[i])
  if (str)
    console.log(keys[i])
}
```

hs0225:
@zherczeg I agree that if a user uses the internal object indiscreetly, it causes a problem in the program. But this is not an easy problem to remove it. Especially, modules with native dependency are not easy to fix. Is there a good idea to solve this problem?

@LaszloLango I think  `_onNextTick` and `_onUncaughtException` should be hidden.

zherczeg:
I have a simple solution. Currently JS callbacks do something like this:

    JHANDLER_CHECK_THIS(object);
    const iotjs_jval_t* builtin = JHANDLER_GET_THIS(object);

The `JHANDLER_GET_THIS` do no checks. However with IOTJS_DEFINE_NATIVE_HANDLE_INFO, we could easily validate the type of an object. I would propose something:

    IOTJS_DEFINE_NATIVE_HANDLE_INFO(type);

    JHANDLER_THIS_TO_OBJ(builtin, object, type);

The `JHANDLER_THIS_TO_OBJ` would check if `object` is object and its type matches to `type`. If not it returns with error. Otherwise it defines `builtin` as `const iotjs_jval_t*` and stores the internal pointer there (no need another `JHANDLER_GET_THIS` call).

yichoi:
#829 can solve a part of problem, GPIO.

hs0225:
@zherczeg The best way is to hide the internal object. In the case of `Buffer`, we should consider how to hide internal object like node.js. 
In the sample above, if user change `this` by mistake or intention and call the internal function, I think your solution has a meaning.

kolipka:
[...]

PR #834, opened 2017-05-11
[Hotfix] Fix an addition in iotjs.js

DanielBallaSZTE:
Fixing an addition in iotjs.js which caused an error when running measure_coverage.sh, since it's in strict mode, and the mentioned line deletes a local variable, which is not allowed.

### DIFF
--- a/src/js/iotjs.js
+++ b/src/js/iotjs.js
@@ -55,22 +55,21 @@ Native.prototype.compile = function() {
 global.console = Native.require('console');
 global.Buffer = Native.require('buffer');
 
-var timers = undefined;
+(function() {
+  var timers = undefined;
 
-var _timeoutHandler = function(mode) {
-  if (timers == undefined) {
-    timers = Native.require('timers');
+  var _timeoutHandler = function(mode) {
+    if (timers == undefined) {
+      timers = Native.require('timers');
+    }
+    return timers[mode].apply(this, Array.prototype.slice.call(arguments, 1));
   }
-  return timers[mode].apply(this, Array.prototype.slice.call(arguments, 1));
-}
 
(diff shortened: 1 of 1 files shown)


src/js/iotjs.js
```diff
@@ -69,9 +69,6 @@ setInterval = _timeoutHandler.bind(this, 'setInterval');
 clearTimeout = _timeoutHandler.bind(this, 'clearTimeout');
 clearInterval = _timeoutHandler.bind(this, 'clearInterval');
 
-// Make sure that it is not accessible elsewhere
-delete _timeoutHandler;
```
hs0225: When you delete this line, we can access this variable anywhere. (related issue #816)
I suggest the following.
```js
(function() {
  var timers = undefined;

  var _timeoutHandler = function(mode) {
    if (timers == undefined) {
      timers = Native.require('timers');
    }
    return timers[mode].apply(this, Array.prototype.slice.call(arguments, 1));
  }

  global.setTimeout = _timeoutHandler.bind(this, 'setTimeout');
  global.setInterval = _timeoutHandler.bind(this, 'setInterval');
  global.clearTimeout = _timeoutHandler.bind(this, 'clearTimeout');
  global.clearInterval = _timeoutHandler.bind(this, 'clearInterval');
})();
```

DanielBallaSZTE: Yes, that seems like a good solution, I didn't really think it through honestly, just wanted to hotfix the issue, and leave the solution to the original committer.

DanielBallaSZTE:
@hs0225 Thanks for the suggestion, added it to the commit.

yichoi:
LGTM

PR #836, opened 2017-05-12
[Code Coverage] Improve coverage of stream_writable.js

DanielBallaSZTE:
Removed the if check, since it would never be called, as doWrite only happens when state.writing is false.

### DIFF
--- a/test/run_pass/test_stream_writable.js
+++ b/test/run_pass/test_stream_writable.js
@@ -1,0 +1,42 @@
+/* Copyright 2015-present Samsung Electronics Co., Ltd. and other contributors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+
+var assert = require('assert');
+
+var Writable = require('stream').Writable;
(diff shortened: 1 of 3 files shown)


test/run_pass/test_stream_writable.js
```diff
@@ -0,0 +1,43 @@
...
+var assert = require('assert');
+
+var Writable = require('stream').Writable;
+var writable = new Writable({highWaterMark: 0});
+
+var writable2 = Writable({highWaterMark: 0});
+
+assert.throws(function() { writable._write(0, 0, 0); }, Error);
+
+process.platform = 'nuttx';
```
hs0225: Why did you change `process.platform`?

yichoi:
@DanielBallaSZTE please rebase it to the master

test/run_pass/test_stream_writable.js
DanielBallaSZTE: Wanted to test if the EOF event runs without errors, making the process believe it's nuttx, so the code runs.

DanielBallaSZTE: Nevertheless, I don't think it's needed, so I'm removing it, just left it there.

DanielBallaSZTE:
@yichoi Rebased, however the master build fails.

daeyeon:
We should avoid adding tests for private/protected members since it make tests verbose and make sources much harder to maintain.

This test doesn't give exact information about what are tested. A common pattern for unit test in my mind is AAA (Arrange, Act, Assert). In this code, I can find sections to arrange values and act testing. However, the section for assert, which verifies that the action behaves as expected, are not enough. This is a common problem I can find these days from commits related to code coverage.

yichoi:
Closing

DanielBallaSZTE:
@daeyeon These tests are to cover the uncovered branches, which point to nothing, like in line 261 the else case. There's nothing to assert.

daeyeon:
Despite your explanation, the test still looks like a cheat to just pass the uncovered branches, and they are unclear. If it's impossible to cover them via normal operations, that may mean we need to consider refactoring it.

PR #842, opened 2017-05-16
Check if Writeable Stream's options is undefined while setting highWaterMark

akhilkedia:
The previous code caused the program to stop if options was not defined while creating a Writable Stream.
This reverts to the default highWaterMark if options is undefined.
The options argument for Writable Constructor should ideally be an optional argument, since the only property it must supply is highWaterMark, which we have a default value for.

IoT.js-DCO-1.0-Signed-off-by: Akhil Kedia <akhil.kedia@samsung.com>

### DIFF
--- a/src/js/stream_writable.js
+++ b/src/js/stream_writable.js
@@ -37,8 +37,8 @@ function WritableState(options) {
 
   // high water mark.
   // The point where write() starts retuning false.
-  var hwm = options.highWaterMark;
-  this.highWaterMark = (hwm || hwm === 0) ? hwm : defaultHighWaterMark;
+  this.highWaterMark = (options && util.isNumber(options.highWaterMark)) ?
+    options.highWaterMark : defaultHighWaterMark;
 
   // 'true' if stream is ready to write.
   this.ready = false;


yichoi:
@akhilkedia, Travis CI failed
```
tools/check_signed_off.sh --travis
Signed-off-by message is incorrect. The following line should be at the end of the 858051f commit's message: 'IoT.js-DCO-1.0-Signed-off-by: Akhil akhil.kedia@samsung.com'. 
[Failed - 1] tools/check_signed_off.sh --travis
```

hs0225:
@akhilkedia I don't know why the previous code is a problem. If `options.highWaterMark` is `undefined` or `null`, it would be `defaultHighWaterMark`. And `highWaterMark` can be zero that is a valid value.

akhilkedia:
@hs0225 My mistake. What I meant was to fix if the options object is undefined. I have updated my PR.

src/js/stream_writable.js
```diff
@@ -37,7 +37,9 @@ function WritableState(options) {
 
   // high water mark.
   // The point where write() starts retuning false.
-  var hwm = options.highWaterMark;
+  var hwm = defaultHighWaterMark;
+  if(options)
+    hwm = options.highWaterMark;
   this.highWaterMark = (hwm || hwm === 0) ? hwm : defaultHighWaterMark;
```
hs0225: I suggest implementing the above 4 lines as below.
``` js
this.highWaterMark = (options && util.isNumber(options.highWaterMask)) ? 
  options.highWaterMask : defaultHighWaterMask;
```
And space is required between `if` and `(`.

hs0225:
@akhilkedia My mistake. Please change `...HighWaterMask` to `...HighWaterMark`.

yichoi:
LGTM

PR #848, opened 2017-05-18
Typo fix in jerry.cmake file for JERRY_PORT_DEFAULT

galpeter:
Usage of JERRY_PORT renamed to JERRY_PORT_DEFAULT as
that is the correct name of the cmake define.


yichoi:
LGTM

PR #851, opened 2017-05-18
Fix bug for family return value in dns.lookup

hs0225:
`family` must be 4 or 6. However, the value of `AF_INET` on Nuttx is 2, not 4. Therefore, we should not return a defined value.


yichoi:
LGTM

PR #864, opened 2017-05-24
Fix BLE defects issued by coverity scan

daeyeon:
This commit fixes 6 defects in the medium level risk.


hs0225:
LGTM

PR #865, opened 2017-05-24
Fix a Buffer defect issued by coverity scan

daeyeon:
This commit fixes a defect in the high level risk.


yichoi:
LGTM

PR #868, opened 2017-05-24
Typo fix in net.js

galpeter:
The error message should refer to 'family' instead of 'port' as
the check is for the family option.


yichoi:
LGTM

Issue #743, opened 2017-04-10
Proposal: Test naming


galpeter:
Currently the test case names does not have a "strict" rule.
There are test file like `test_i2c.js` and also `test_uncaught1.js`. In the first example the test name explicitly contains the module under testing (or related module under testing). But the `test_uncaught1.js` does not contain any module info thus it is not clear for which module is tested (the `test_uncaught1.js` tests one of the `process` module's functionality).

Additionally the names of the test cases are used for test skipping. For example if someone specifies the `skip-module=i2c` the `test_i2c.js` will not be executed which is great. However if the test file is named like `test_writebyte.js` the file would have been executed.

The detection of which tests should be skipped is done by an `indexOf` call, that is we only check if the module name (specified by the `skip-module` argument) is in the target test's name. In the case of the `test_uncaught1.js` to skip it the `skip-module=uncaught` would be necessary.

What I'm proposing is to have all test file names the same naming convention. That is all test files should be named like: `test_<module name>[_<functionallity].js`. Most of the tests already adhere to this convention.

The affected test files (and their new proposed name) in the `test/run_pass` are the following:

Current name |  New name
-|-
test_cwd.js | test_process_cwd.js
test_exit.js | test_process_exit.js
test_experimental_on.js | test_process_experimental_on.js
test_experimental_off.js | test_process_experimental_off.js
test_fs1.js | test_fs_1.js
test_fs2.js | test_fs_2.js
test_gpio1.js | test_gpio_1.js
test_gpio2.js | test_gpio_2.js
test_httpclient_timeout.js | test_net_httpclient_timeout.js
test_httpclient_timeout2.js | test_net_httpclient_timeout2.js
test_httpserver.js | test_net_httpserver.js
test_httpserver_timeout.js | test_net_httpserver_timeout.js
test_net10.js | test_net_10.js
test_net1.js | test_net_1.js
test_net2.js | test_net_2.js
test_net3.js | test_net_3.js
test_net4.js | test_net_4.js
test_net5.js | test_net_5.js
test_net6.js | test_net_6.js
test_net7.js | test_net_7.js
test_net8.js | test_net_8.js
test_net9.js | test_net_9.js
test_next_tick.js | test_process_next_tick.js
test_timers2.js | test_timers_2.js
test_uncaught1.js | test_process_uncaught1.js
test_uncaught2.js | test_process_uncaught2.js
test_uncaught_error1.js | test_proces_uncaught_error1.js
[...]

PR #869, opened 2017-05-24
Rename test_promise.js to test_iotjs_promise.js

galpeter:
As per #743 the test_promise.js name for the test
would expect that the iotjs would be built with
promise iotjs(!) module. However the promise is
not an iotjs feature but it is provided by the
underlying JS engine JerryScript. Thus the test
runner incorrectly returns the reason for skipping
the test, stating that is not supported by the
iotjs build but that is the incorrect message.

By renaming the testcase to test_iotjs_promise.js
the correct skip reason is displayed and will
make it easier in the future to enable the test case.


glistening:
@galpeter I didn't recognized that the file name is related to test runner's running. The PR is good for me. However, at least for me, this PR's new name is counter-intuitive. Because it requires iotjs infix, although promise is JavaScript engine (jerryscript) 's feature, not iotjs's feature. Anyway, I think it would be better to make a document about test file naming rule. Could you make a document in doc/dev?

glistening:
@galpeter I think #743 explanation is enough. We can use #743 for a new document for making test cases in developer guide. Also, we need to add comment for exceptional cases, that "_For tests that don't belong to any iotjs module, use iotjs infix._ " (although I still feel strange).

galpeter:
We can introduce another "virtual" module name which is hardcoded into the testing system so it knows how to handle them. So if you have another idea instead of the 'iotjs' name we can change it. I picked the 'iotjs' as it is a must have module.

Should I add the doc into this PR or a separate one?

glistening:
@galpeter `iotjs` infix is enough for me if the rule is documented explicitly. I found there is already [To write a testcase](https://github.com/Samsung/iotjs/blob/master/docs/help/Getting-Started.md#to-write-a-test-case) in Getting Started. Could you please add a bullet for testcase naming rule? Thank you.

galpeter:
I've updated the PR with the requested documentation change.

yichoi:
LGTM

Issue #862, opened 2017-05-23
Fix net tests on stm32f4dis-nuttx


rtakacs:
The following net tests are failing (caused by #851):

* test_net_1.js
* test_net_2.js
* test_net_4.js
* test_net_5.js
* test_net_6.js
* issue-266.js

You can check it on the testresult webpage: https://samsung.github.io/iotjs-test-results/

PR #871, opened 2017-05-26
Use AF_INET by default on NuttX when no address family specified.

rtakacs:
Added `AF_INET` as a default family on NuttX because the `inet_pton` function can handle only `AF_INET` or `AF_INET6` families. You can check it in the [NuttX source code](https://bitbucket.org/nuttx/nuttx/src/710376291a7a4f2201a77137bd9ccdcc4405180d/libc/net/lib_inetpton.c?at=master&fileviewer=file-view-default#lib_inetpton.c-372).

This patch is a fix for the #862.


yichoi:
LGTM

Issue #873, opened 2017-05-29
Documents


daeyeon:
1) Ensuring that it precisely describes API interface and its parameter conditions.
2) Applying the most of Node.js format (e.g, title, subtitle, leveling rules, and so on) except the following cases.
  - Place 'Platform Support' section at the beginning.
  - Use the current format for callback arguments
3) Improving lacks and faults of the documents.

---
Common Checklists
---
1) For primitive types, start with lowercase. e.g) `{String}` should be `{string}`. Its criteria is the form of Node.js.
2) Use single quotes for contents and samples.
3) Remove TOC
4) For callback arguments, follow the below example.

```
* `someCallback` {Function}
  * `arg1` {integer}
  * `arg2` {Object}
```
5) Remove the section title, `constructor`, and write it simply like below.
```
new Buffer(array)
```
6) Remove all the section titles such as Class Methods, Prototype Function, Module Functions/Properties, Properties, Event, and so on.
```
## Class Methods <--- should be deleted.
### dgram.createSocket(options[, callback]) <-- keep the three hashmarks
```
7) Use the below format like Node.js

```
Buffer.from(string,[ encoding]) // for Class Method
buf.copy(...) // for Prototype Function or Module Function
```

8) In case of Class and Event, use the prefix as below, but not for others like Class Method.
```
Class: http.ClientRequest // Ok
Event: 'abort' // Ok 

Class Method: Buffer.alloc // Bad
```

:books: Please update this section once you find any.

---
Group A
---

- [x] Net
- [x] HTTP    
- [x] DGRAM
- [x] File-System
- [x] Buffer
- [x] DNS

Group B
---

- [x] Assert
- [x] Events
- [x] Module
- [x] Process
- [x] Stream
- [x] Timers

Group C
---

- [x] ADC
- [x] BLE
- [x] GPIO    
- [x] I2C
- [x] PWM
- [x] SPI
- [x] UART

zherczeg:
I would like to work on the DGRAM module (testing + documentation).

LaszloLango:
I'd kike to work on the HTTP module.

galpeter:
I'll work on the dns module.

bzsolt:
I'll work on the **fs** module.

glistening:
@daeyeon, @zherczeg, @LaszloLango, @galpeter, @bzsolt 

I found some inconsistency between the each module documentation:

1. "Prototype Function" or "Class Method"
[...]

PR #879, opened 2017-05-30
Update Dgram documentation.

zherczeg:


docs/api/IoT.js-API-DGRAM.md
```diff
@@ -5,265 +5,405 @@ The following shows dgram module APIs available for each platform.
...
-  * `reuseAddr <Boolean>` - it indicates to allow socket to bind the address and port used previously.
-* `type <String>` - it indicates an address family either `udp4` or `udp6`.
-* `createListener <Function(msg: Buffer, rinfo: Object)>`
-  * `rinfo <Object>`
-    * `address <String>`
-    * `family <String>` - it indicates an address family either `IPv4` or `IPv6`
-    * `port <Number>`
+### dgram.createSocket(options[, callback])
+* `options` {Object}
+** `type` {String}
```
galpeter: I think the `String` should be `string`.

src/js/dgram.js
```diff
@@ -132,7 +139,7 @@ Socket.prototype.bind = function(port /*, address, callback*/) {
 
     self._handle._reuseAddr = self._reuseAddr;
 
-    var err = self._handle.bind(ip, port || 0);
+    var err = self._handle.bind(ip, port | 0);
```
daeyeon: Isn't `||` right?

daeyeon:
Could you use single quotes for all the samples? I've updated common mistakes in #873.

src/js/dgram.js
zherczeg: I made it stronger to get error early. E.g. if port is an empty `{}` object, then `{} || 0` returns with `{}` and we get an error later in the native part. However the `| 0` forces an int32 conversion, which is enough for a port number. Also `new Number(13) | 0` will give back 13, rather than an object, which simplifies conversions later as well (e.g. native functions complains on type mismatch).

zherczeg:
I would like to enable dgram module on Posix systems, but I get various errors. E.g `Detected module(s) both in include and exclude list. Module name(s): dgram` on NuttiX. It seems `iotjs-include-module` needs change as well. Could somebody explain me how to do this properly?

galpeter:
@zherczeg This error comes from the `tools/module_analyzer.py`. Previously it was invalid to both include and exclude a module but now as the exclude list is target dependent I think we can remove this check. The check is done in the lines 50-58 in https://github.com/Samsung/iotjs/blob/master/tools/module_analyzer.py#L54

src/js/dgram.js
daeyeon: Thanks for clarifying my query.

zherczeg:
Shall I cut the re-enabling of dgram testing to another PR? Seems more complicated than I hoped.

zherczeg:
Text updated according to format.

galpeter:
@zherczeg after #894 you can add the dgram module to the include list and to the nuttx exclude list.

yichoi:
LGTM

Issue #853, opened 2017-05-19
Unit Tests for release 1.0


daeyeon:
This task is not for migrating Node.js tests as many as possible, but for bugfixing.

1) Find tests, which test 1 feature what IoT.js should support.
2) If fails, then fix it, and migrate the test

Common Checklists
---
- For the adopted tests, put the Samsung license and Node.js one in order, and update the year on Samsung's one with 2017.

:books: Please update this section once you find any.

Group A
---

- [x] Net
- [x] HTTP
- [x] File-System

Group B
---

- [x] Timers

yichoi:
Need to report Unit Test Coverage

daeyeon:
For release 1.0, we investigated `http`, `http`, `fs`, and `timers` which seem high priority, and adopted some of them with bug fixes. The items confirmed and left were reported. 

This task turns out quite big to finish up for 1.0 release schedule. So I think we need to close this for now and divide them smaller and make them 1.x tasks.

PR #881, opened 2017-05-30
Adapt test-assert.js test from Node.js to IoT.js.

zherczeg:


daeyeon:
@zherczeg Kindly let me have your thought on my suggestion.

First of all, we don't need to add the unit test for `assert.ok` since it's not described in the document. (not supported.) 

Second, we can find many node.js tests that use `assert.ok`. So, for using them with minimal changes, we could extend `assert` in common.js like below.

`test.js`
```
var common = require('./common');
var assert = require('assert');

// Assuming the below is a test from node.js, it doesn't work since 
// IoT.js doesn't support assert.ok. So, assert.ok is extended in common.js.

test(function serverTimeout(cb) {
    var server = https.createServer(....);
    server.listen(0, common.mustCall(function () {
        ...
        assert.ok(s instanceof https.Server); // <-- This works now.
        ...
    }));
});
```

`common.js`
```
var assert = require('assert');

assert.ok = function (value, message) {
    assert.equal(!!value, true, message);
}
```

test/node/parallel/test-assert.js
```diff
@@ -0,0 +1,730 @@
...
+// Test starts here
+
+function makeBlock(f) {
+  var args = Array.prototype.slice.call(arguments, 1);
+  return function() {
+    return f.apply(this, args);
+  };
+}
+
+assert.ok(a.AssertionError.prototype instanceof Error,
```
daeyeon: we don't need to add the unit test for assert.ok

zherczeg:
Could you review this patch?

daeyeon:
This a kind of skeleton for #853 and will be incrementally updated.

zherczeg:
@daeyeon could you check this again?

yichoi:
LGTM

PR #888, opened 2017-06-01
TizenRT support for PWM

kolipka:

### DIFF
--- a/docs/targets/tizenrt/artik05x/IoT.js-API-artik053.md
+++ b/docs/targets/tizenrt/artik05x/IoT.js-API-artik053.md
@@ -1,0 +1,33 @@
+## Artik053 module
+
+## `Pin`
+To use system IO, such as PWM, you must know pin name.
+
+
+### `PWM Pin Number` <a name="pwm-pin"></a>
+
+For example,
+``` javascript
+var pwm = new require('pwm')();
+var config = {
+  pin: 0,
+  period: 0.001,
+  dutyCycle: 0.5
+}
+var pwm0 = pwm.open(config, function() {
+  if (err) {
+    throw err;
+  }
+});
+```
+The following is a list of PWM pin numbers.
+
+| PWM Pin number | GPIO Name |
+| :---: | :---: |
+| 0 | XPWMTOUT_0 |
+| 1 | XPWMTOUT_1 |
+| 2 | XPWMTOUT_2 |
+| 3 | XPWMTOUT_3 |
+| 4 | XPWMTOUT_4 |
+| 5 | XPWMTOUT_5 |
+| 6 | XPWMTOUT_6 |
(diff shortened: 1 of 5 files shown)


test/run_pass/test_pwm.js
```diff
@@ -19,12 +19,17 @@ var Pwm = require('pwm');
...
+  period: 0.001,  // 1kHz
+  dutyCycle: 0.2  // 20%
+};
 
 if (process.platform === 'linux') {
   configuration.pin = 0;
 } else if (process.platform === 'nuttx') {
   configuration.pin = require('stm32f4dis').pin.PWM1.CH1_1;
+} else if (process.platform === 'tizenrt') {
+  configuration.pin = require('artik053').pin.PWM0;
```
kolipka: I think it is useful for end developer. It is looks better when you have: `open({pin: pin.PWM0})` than `open({pin: 0})`. This is meaningful, easy to read ID instead of hardcoded magic number. Additionally it is consistent with stm.

docs/targets/tizenrt/artik05x/IoT.js-API-artik053.md
```diff
@@ -0,0 +1,31 @@
+## Artik053 module
+
+## `Pin`
+To use system IO, such as GPIO, PWM and ADC you must know pin name.
```
kolipka: will it be OK if I change it to:
PWM (ADC, GPIO, etc in the future) ? I think we should have all pins described this way. It is easy and nice for external developers. 0 for PWM and 0 for GPIO is the same number but differend PIN. I think more readable is to have pin.PWM0 and pin.GPIO0

test/run_pass/test_pwm.js
hs0225: Memory usage will increase. In addition, developers should be include additional module. I think, this will decrease usability. The value in the pin module contains information for initializing pwm on stm. It includes pin number, port number, pin mode, and so on, not 1 2 3. The pin module of stm are required for advanced developers. However, I plan to support pwm for general developers without including pin modules.

src/modules/iotjs_module_pwm.h
```diff
@@ -16,11 +16,16 @@
 
 #ifndef IOTJS_MODULE_PWM_H
 #define IOTJS_MODULE_PWM_H
+#include <stdint.h>
```
kolipka: Ah, my bad - it is required only by TizenRT, I will move it below

test/run_pass/test_pwm.js
kolipka: I disagree with you, but you are the maintainer here.

Memory usage: This is 6 int props, propably much less space than default config for spi...
Usability: please try to find out where in Artik053 docs there is info about PWM pins, additionally, we will have to provide mapping from raw magic numbers to actual PWM lines, so developer will have to do:
var PWM0 = 0; open({pin: PWM0... 
to avoid hardcoded magic values in the source code. Exactly the same, but will have to read more docs to know the number...

yichoi:
LGTM

PR #890, opened 2017-06-02
Adapt test-buffer-alloc.js test from Node.js to IoT.js. (#881)

zherczeg:
There are some tests, which fails but I feel they should work. What is your opinion about them?

### DIFF
--- a/test/testsets.json
+++ b/test/testsets.json
@@ -107,6 +107,7 @@
     { "name": "test_process_implicit_exit.js", "expected-failure": true }
   ],
   "node/parallel": [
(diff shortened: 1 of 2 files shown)


zherczeg:
(These tests are commented out.)

test/testsets.json
```diff
@@ -106,5 +106,6 @@
   ],
   "node/parallel": [
     { "name": "test-assert.js" }
+    { "name": "test-buffer-alloc.js" }
```
galpeter: Missing a comma before the new entry.

PR #891, opened 2017-06-02
Remove test_net_11.js

DanielBallaSZTE:
Work in progress:
Removing test_net_11.js and splitting it to more files.


hs0225:
LGTM

PR #893, opened 2017-06-02
Implement fs.fstat() and fs.fstatSync()

bzsolt:


test/run_pass/test_fs_fstat_fstatsync.js
```diff
@@ -0,0 +1,77 @@
...
+ * limitations under the License.
+ */
+
+var fs = require('fs');
+var assert = require('assert');
+
+var testfile = process.cwd() + '/run_pass/test_fs_fstat_fstatsync.js';
+var testdir = process.cwd() + '/resources';
+
+// fstat - file
```
galpeter: I think it would be better to have the fstat (async) tests its own file (test_fs_fstat.js). Some of the other tests are also structured that way.

bzsolt: You're right, I've splitted the testfile.

yichoi:
LGTM

PR #894, opened 2017-06-02
Rework the module checker logic

galpeter:
Change the logic of include and exclude module handling.
Now the exclude list will remove elements from the
include list.

With this it is possible to add a module to the include list
and exclude it on specific platforms.


galpeter:
Issue raised during #879

PR #895, opened 2017-06-02
Disable global checks in Node common.js

galpeter:
Few modifications:
* For now disable the global leak checks for Node tests.
* Fix the usage of array include method which is not
  present in ES 5.1.
* Remove the 'opensslCli' entry as it does not work on IoT.js.

### DIFF
--- a/test/node/common.js
+++ b/test/node/common.js
@@ -184,28 +184,6 @@ Object.defineProperty(exports, 'localhostIPv4', {
   }
 });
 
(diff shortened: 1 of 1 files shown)


galpeter:
Without  the PR the `node/parallel/test-assert.js` returned an exit code of 2 when it was executed as  stand alone script.

yichoi:
LGTM

PR #896, opened 2017-06-02
Fix fs.read()

bzsolt:
Make the synchronous and asynchronous variant consistent.
When `position` is null should read from the current position.


yichoi:
LGTM

Issue #874, opened 2017-05-29
Validate arguments on C side for Release 1.0


daeyeon:
Improve all argument checkers on C side in release running. Can be disabled, although not recommended only for trusted code.

- #898 and #911 are landed as a PoC for `Buffer`.

---
- [x] Process
<del>- [ ] Console</del>
- [x] Timers
<del>- [ ] Module</del>
---
<del>- [ ] Assert</del>
- [x] Buffer
- [x] DGRAM
<del>- [ ] DNS</del>
- [x] Events
<del>- [ ] File-System</del>
- [x] HTTP
- [x] Net
- [x] Stream
- [x] ADC
- [x] BLE
- [x] GPIO
- [x] I2C
- [x] PWM
- [x] SPI
- [x] UART

kolipka:
Do we know how it will impact binary size / memory usage?

zherczeg:
Please check #816

kolipka:
@zherczeg I don't get it :/ You are refering to my question about memory impact? I don't see how those are related. Your solution in #816 is to add checks in C code. Description here states 

> Remove all argument checkers from the C codes and do it on JS side

Anyway: adding something on JS side does have bigger impact(mem+runtime performance) than native check. The module list here is quite long, and if we removed capacity from string to have savings, I think we should be careful also in other places.

zherczeg:
@kolipka no, the issue is something else. Its security, not resource. In that issue I was able to hijack the internal parts of a Buffer object, and native code crashes rather than throwing an error (or do something nasty for a cleverly constructed input). If we go further into this direction (do argument validation in JS side), the problem might deepen.

daeyeon:
@zherczeg Could you update the above checklist?

PR #898, opened 2017-06-06
Add new argument validators enabled in debug

daeyeon:
This commit adds validators and applies them to the modules, so that
we can remove validating arguments in release build. This feature is protected
under experimental status until #874 is done.


glistening:
I omitted the word `not` in above comment by mistake. I intended to write "_I am not 100% sure._"

yichoi:
LGTM

PR #861, opened 2017-05-22
Minor fix in the documentation of NuttX build.

LaszloLango:
IoT.js-DCO-1.0-Signed-off-by: László Langó llango.u-szeged@partner.samsung.com

PR #899, opened 2017-06-06
Update the Makefile to build the IoT.js app in nuttx.

cedricDum:
The Makefile runs in the nuttx apps tree. It needs IOTJS_ROOT_DIR to link to IoT.js.
The commit #861 suggests to define IOTJS_ROOT_DIR to the relative path of IoT.js as seen from nuttx root, which is a good thing.
But the actual Makefile is in apps/system/iotjs, forbidding the use of a relative path ( ../../../iotjs would work, but is counter intuitive).
The patch allows to provide either an absolute path or a relative path starting from the main nuttx Makefile.

IoT.js-DCO-1.0-Signed-off-by: cedricDum duminy.cedric@gmail.com

### DIFF
--- a/config/nuttx/stm32f4dis/app/Makefile
+++ b/config/nuttx/stm32f4dis/app/Makefile
@@ -52,14 +52,15 @@
 -include $(TOPDIR)/Make.defs
 include $(APPDIR)/Make.defs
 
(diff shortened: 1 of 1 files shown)


daeyeon:
LGTM

PR #911, opened 2017-06-08
Add a more secure type validator to Buffer

daeyeon:
This commit add a validator to check whether the execution
context on `this` is not the expected C type. #816

### DIFF
--- a/src/iotjs_binding.h
+++ b/src/iotjs_binding.h
@@ -128,6 +128,8 @@ iotjs_jval_t iotjs_jval_get_property(THIS_JVAL, const char* name);
 void iotjs_jval_set_object_native_handle(THIS_JVAL, uintptr_t ptr,
                                          JNativeInfoType native_info);
 uintptr_t iotjs_jval_get_object_native_handle(THIS_JVAL);
+uintptr_t iotjs_jval_get_object_from_jhandler(iotjs_jhandler_t* jhandler,
+                                              JNativeInfoType native_info);
 
 void iotjs_jval_set_property_by_index(THIS_JVAL, uint32_t idx,
                                       const iotjs_jval_t* value);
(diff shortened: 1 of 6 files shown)


zherczeg:
Good patch. I have a question though. It seems we call jerry_get_object_native_pointer twice: once for validating, and once more for getting the internal pointer. It could be an option to make this function a real function (not a macro). Something like this:

```
void *get_native_pointer(jval_t, expected_type)
```

It would first check whether the value is object, then it checks the expected_type, and if all is well it returns with a pointer. In all other cases it returns with NULL. This would reduce the binary size of iot.js.

You could create a macro which casts the void * to an appropriate type. When the returned value is NULL, we should throw an exception, which should be the same (and we could also create a global function for it).

glistening:
@daeyeon I checked binary size changes with this PR. Because I wondered about binary size when we add type check for almost all binding functions. In short, it seems okay with following simple experiment.

With this patch, binary size is increased by **60** bytes (622,040 -> 622,100).
After inserting one more `JHANDLER_CHECK_THIS_SAFE`, it remains 622,100 bytes. (There is no increase.)

@zherczeg Thank you for reporting bug and suggesting solution.

daeyeon:
@zherczeg Thanks for the review. I was in a similar way to merge the two but it wasn't likely to get that with simple replacement. So I chose only the one not to impact much. I will make a patch involving that. We could progressively apply it once agreed.

src/iotjs_binding.h
```diff
@@ -293,6 +295,24 @@ static inline bool ge(uint16_t a, uint16_t b) {
...
 #define JHANDLER_FUNCTION(name) static void name(iotjs_jhandler_t* jhandler)
 
+#define JHANDLER_CHECK_NATIVE_INFO(x, native_info)               \
+  if (!iotjs_jval_get_object_native_handle_s(x, &native_info)) { \
+    JHANDLER_THROW(COMMON, "Unsafe access");                     \
+    return;                                                      \
+  }
+
+#define JHANDLER_CHECK_THIS_SAFE(type) \
+  JHANDLER_CHECK_NATIVE_INFO(JHANDLER_GET_THIS(type), this_module_native_info)
```
zherczeg: These two macros seem unused.

src/iotjs_binding.h
```diff
@@ -293,6 +295,24 @@ static inline bool ge(uint16_t a, uint16_t b) {
...
+
+#define JHANDLER_DECLARE_WRAPPER(type, name)                           \
+  JHANDLER_CHECK_THIS(object);                                         \
+  iotjs_##type##_t* name = (iotjs_##type##_t*)                         \
+      iotjs_jval_get_object_native_handle_s(JHANDLER_GET_THIS(object), \
```
zherczeg: I would do the object check (`JHANDLER_CHECK_THIS(object)`) inside `iotjs_jval_get_object_native_handle_s`. This would reduce binary size since this is a frequently used function. Whit is the _s postfix? static?

daeyeon: You're very right. I'll fix it. The suffix is inspired from MS's security enhanced api like `printf_s`

yichoi:
LGTM

Issue #854, opened 2017-05-19
Some known issues for Release 1.0


daeyeon:
**Infra**
- [x] Add daily benchmark (memory usage and binary footprint size) for NuttX/STM32 and Linux/RPi (https://samsung.github.io/iotjs-test-results/)
- [x] Integrate license verification tool (https://goo.gl/SXMcbq)
---
**Defects**
- [x] Fix static analysis defects (https://scan.coverity.com/projects/samsung-iotjs)
- [x] Validating target supports on RPi/Linux and NuttX/STM32.
---
**Code Coverage**
- [x] Goal: Functions 100% (Public in JS only), Statement over 80% (checker: ./tools/measure_coverage.sh)

yichoi:
Somwhat dupe with #932.
let's close together with the issues are resolved

PR #914, opened 2017-06-08
Split test_net_11.js

DanielBallaSZTE:
Split test_net_11.js into multiple tests for better readability.

Had to recreate the PR, as I accidentally deleted then restored the branch.


LaszloLango:
Past reviews and conversations: #903

daeyeon:
This commit was removed in #891 because 1) the tests was in testing process for about 10 min 2) many tests looked only for increasing branch coverage, but has less point as you mentioned before.

Since it would have less meaning to reach 100% of branch coverage, we have no goal for it. (#854)
This commit still look to be made for such a purpose. If right, I would stay without this. Please close this.

Plus, double quote and indentations aren't fixed yet. Some cases uses compulsory changes of internal property.

PR #925, opened 2017-06-12
TizenRT support for UART

kolipka:


LaszloLango:
The new file looks the same as the NuttX support of UART. Is there any difference? I think we should revise the platform folder to reduce the code duplications in the near future.

kolipka:
no, it is exactly the same. 
If I may, I would also suggest changing file naming rules, since adding folder name as a prefix (the same with os name suffix) seems to hide meaningful part of file name. The same with mixing _ with -

yichoi:
@kolipka please rebase it to the master

yichoi:
@kolipka BTW, please add your test result on real target like https://samsung.github.io/iotjs-test-results/
I hope you wil make a site like this for TizenRT & Tizen  sooner or later

PR #630, opened 2017-02-16
Add mbedtls submodule

mmiodeksamsung:
- add submodule
 - building and linking mbedtls libraries with iotjs

PR #926, opened 2017-06-12
HTTPS Module for Tizen

akhilkedia:
This implements an `HTTPS` client module for `Tizen` platform. The API is almost the same as the `HTTP` module, minus the `server` parts.

`Https` is needed for connections to most cloud services and APIs, which often do not have an `Http` fallback.

We use `libcurl` to handle the backend of the request. Since `tizen` already has `libcurl`, and we dynamically link against it, this results in only a very small increase in binary size (compared to say statically linking against `MbedTLS`).

Here is a binary size increase comparison - 

Build Type | Master | With HTTPS 
---|---|---
Debug | 19,92,117 Bytes | 20,62,718 (around 3.5% increase)
Release | 4,62,179 Bytes | 4,84,393 (around 4% increase)

The code was tested on the latest `Tizen 3.0` snapshot on a `TM1` device.

_Note -_ While by default it only builds for `Tizen`, by applying this [patch](https://gist.github.com/akhilkedia/756b2e060b861e11523d4717c2f9849c) you can also build it for Linux and run the `valgrind` testcases on the Linux host if you have `libcurl4-openssl-dev` or `libcurl4-gnutls-dev` installed. However this method will not build when cross-compiling for `arm` or for `32-bit linux` since we need to link against `libcurl` compiled for those platforms.

IoT.js-DCO-1.0-Signed-off-by: Akhil akhil.kedia@samsung.com


src/js/https_client.js
```diff
@@ -0,0 +1,185 @@
...
+// Concrete stream overriding the empty underlying _write method.
+ClientRequest.prototype._write = function(chunk, callback, onwrite) {
+  httpsNative._write(this, chunk.toString(), callback, onwrite);
+};
+
+ClientRequest.prototype.headersComplete = function() {
+  var self = this;
+  self.emit('response', self._incoming);
+  var isHeadResponse = (self.method == 'HEAD');
+  return isHeadResponse;
```
zherczeg: `return (self.method == 'HEAD')` ?

src/js/https_client.js
```diff
@@ -0,0 +1,185 @@
...
+      utftext += String.fromCharCode((c & 63) | 128);
+    } else {
+      utftext += String.fromCharCode((c >> 12) | 224);
+      utftext += String.fromCharCode(((c >> 6) & 63) | 128);
+      utftext += String.fromCharCode((c & 63) | 128);
+    }
+
+  }
+
+  return utftext;
```
zherczeg: Cant we use buffers for this conversion?

src/js/https_client.js
```diff
@@ -0,0 +1,185 @@
...
+      enc3 = enc4 = 64;
+    } else if (isNaN(chr3)) {
+      enc4 = 64;
+    }
+
+    output = output +
+    _keyStr.charAt(enc1) + _keyStr.charAt(enc2) +
+    _keyStr.charAt(enc3) + _keyStr.charAt(enc4);
+
+  }
```
zherczeg: There are lots of empty lines before `}`. Are they intentional?

src/js/https_client.js
akhilkedia: Makes sense. I'll change.

src/js/https_client.js
akhilkedia: No they are not. The are the leftover of since deleted debug statements, and I had (mistakenly) expected `clang-format` will just take care of such issues. I will remove such extra lines.

src/js/https_client.js
akhilkedia: Do you mean to directly use Buffer(String) and then Buffer.toString('utf8')?

From a quick glance of the IoT.js module, i wasn't sure of the level of support in IoT.js for this. While I can see 'hex' as an encoding explicitly, and I can see uint-8, I wasnt sure it handled UTF-8.

akhilkedia: Going by a new example in the Buffer API documentation, it would appear I can indeed use the above method. I'll remove this function utf8Encode and use buffers for this.

akhilkedia:
@zherczeg I've updated with info on binary size increase. 
Also, one of the Travis jobs seems to be failing with an error unrelated to mine - perhaps that build can be restarted?

zherczeg:
Seems already restarted.

kolipka:
What about TizenRT/Nuttx? There is no dynamic linking on TizenRT, so this has to be static. There is already MBedTLS there. It is designed for such small systems and used by Iotivity.
Do we want to have different TLS library for Linux and embedded systems?

akhilkedia:
@kolipka My intention with making this module was to make it for only `Tizen`, which already has `libcurl` so I could take advantage of it.
Dynamic linking results in around **4%** increase in binary size, **compared to 50%** when using MbedTLS (according to #630). This difference is, frankly, huge.
Forcing static linking would be **extremely** suboptimal for `Tizen`, to say the least.

akhilkedia:
Right now, there aren't any other `HTTPS` implementations for any platform.
Until then, why not atleast support `Https` on atleast one platform?
When there is another `Https` module which works for all platforms, perhaps this issue could be raised again? (though the above size considerations would still apply.)

kolipka:
@akhilkedia:

1. I am not against dynamic linking on linux/tizen. MbedTLS can be linked in both ways.
2. On TizenRT there is no way to dynamic link(it is whole static blob). MbedTLS is build in.
3. Comparing size impact of dynamic libcurl with static mbedtls is kind of apples to oranges comparison. static MbedTLS takes 500kb. How much takes to put libcurl (with dependencies) on arm?
4. TizenRT is based on Nuttx, so hopefully if mbedtls works on TizenRT, it will work on Nuttx too.

I am against using libcurl for Linux/Tizen and MbedTLS for TizenRT/Nuttx.  It is not a matter of temporary support for HTTPS just for linux/Tizen, but maintenance effort later. While I'm pretty sure MbedTLS will work for all OS we support currently, I fear that libcurl(with deps) will not fit small embedded platforms.

akhilkedia:
(long post, see tl;dr below.)
@kolipka 
The following is true for Tizen devices (as of now) - 
1. No `HTTPS` on `IoT.js` right now.
2. `Libcurl` already present in the system.

Now, for anyone using `Tizen`, there are 2 alternatives - 
1. Use this `HTTPS` implementation - The advantages of this are : 
    - It's already implemented.
    - Only 4% increase in size from a system which had an older version of `IoT.js`. 
Yes, hypothetically, if you had to place `libcurl` and its deps on a vanilla system, the size increase will be much more.  But `Tizen` is not a vanilla system. 
In practise, because `libcurl` is already there, the size increase in Indeed only 4%
2. The `mbedTLS` implementation - The disadvantages of this are :
    - There is no such implementation yet.
    - Even after waiting, the size increase in `IoTjs` will be 50%.

Now, the disadvantages of using this implementation of `Https` for `Tizen` that you mentioned - 
1. Increased maintenance work going forward -
    - Until a new `https` module supporting all platforms land, I can offer to take up the maintaince work on this module for Tizen. Most changes to `IoT.js` rarely impact side modules substantially, and whenever there are changes that need to addressed, I offer to do those.
2. `Libcurl` will not fit small embedded platforms
    - This module is simply not for any other platforms except `Tizen`. If you see the `build.config`, this module will only be enabled for `Tizen`. For any other platforms, adding this module to `IoTjs` changes nothing.
3. Using multiple backends for `Https` on different platforms
    - Right now, there are no other backends other than this `libcurl` implementation. Wouldn't this question of which backend to use make sense once someone PRs an alternative that the community can actually consider?
    - Even when there is a PR for `Https` for all platforms using `mbedTLS`, there is no one size fits all. Using `mbedTLS` it will not have an 'ideal' solution for `Tizen` at all, and this `libcurl` version will not be ideal for embedded systems. Perhaps multiple implementations of `https` backends should be considered in the future when the `mbedTLS` PR does land?

**tl;dr -**  This adds extra features to IoT.js on Tizen, and does not affect any other platform. I also offer to handle future maintenance work until there are alternatives - at which point, their benefits/disadvantages should be considered and the better selected. Until then, I see absolutely no reason to not use this `Https` version.

yichoi:
We don't need to use `mbedTLS` for all Tizen and Tizen RT. just use for each requirements.

docs/api/IoT.js-API-HTTPS.md
```diff
@@ -0,0 +1,133 @@
...
+
+|  | Linux<br/>(Ubuntu) | Raspbian<br/>(Raspberry Pi) | Nuttx<br/>(STM32F4-Discovery) | Tizen<br/>(Artik 10) |
+| :---: | :---: | :---: | :---: | :---: |
+| https.request  | X | X | X | O |
+| https.get  | X | X | X | O |
+
+
+# Https
+
+IoT.js provides HTTPS to support HTTPS clients enabling users to send HTTPS request easily.
```
zherczeg: request -> requests

docs/api/IoT.js-API-HTTPS.md
```diff
@@ -0,0 +1,133 @@
...
+
+### https.request(options[, callback])
+* `options` {Object}
+  * `host` {string} A domain name or IP address of the server to issue the request to. Defaults to 'localhost'.
```
zherczeg: `**Default:**` not Defaults to. Please fix it everywhere.

docs/api/IoT.js-API-HTTPS.md
```diff
@@ -0,0 +1,133 @@
...
+
+Note that in the example `req.end()` was called. With `https.request()` one must always call `req.end()` to signify that you're done with the request - even if there is no data being written to the request body.
+
+### https.get(options[, callback])
+* `options` {Object}
+* `callback` {Function}
+  * `response` {https.IncomingMessage}
+* Returns: {https.ClientRequest}
+
+Same as `https.request` except that `https.get` automatically call `req.end()` at the end.
```
zherczeg: Could be a link. Options are also the same? Perhaps we should list them here for completeness.

docs/api/IoT.js-API-HTTPS.md
```diff
@@ -0,0 +1,133 @@
...
+A list of HTTPS methods supported by the parser as `string` properties of an `Object`. Currently supported methods are `'DELETE'`, `'GET'`, `'HEAD'`, `'POST'`, `'PUT'`, `'CONNECT'`, `'OPTIONS'`, `'TRACE'`.
+
+
+## Class: https.ClientRequest
+
+This object is created internally and returned from https.request(). It represents an in-progress request whose header has already been queued.
+
+https.ClientRequest inherits [`Stream.writable`](IoT.js-API-Stream.md). See it's documentation to write data to an outgoing HTTP request. Notable methods are `'writable.write()'` (to send data as request body), `'writable.end()'` (to signal an end and flush remaining request body), and the event `'finish'`.
+
+### request.aborted
```
zherczeg: `Event:` comes first as far as I understood.

yichoi:
@akhilkedia Have you tested with this on RPI3 Tizen ? if so, please share the result of this patch

src/js/https_incoming.js
```diff
@@ -0,0 +1,242 @@
...
+  }
+}
+
+//------------ LIBCURL PARSER CALLBACKS -----------------
+// Called by libcurl when Request is Done. Finish parser and unref
+function cbOnEnd() {
+  var incoming = this;
+  var parser = incoming.parser;
+  if (parser) {
+    // unref all links to parser, make parser GCed
```
zherczeg: unref -> Unref

src/js/https_incoming.js
```diff
@@ -0,0 +1,242 @@
...
+    // Socket closed before we emitted 'end'
+    incoming.on('end', function() {
+      incoming.emit('close');
+      clientRequest.emit('close');
+    });
+    incoming.push(null);
+  } else if (!incoming.started) {
+    incoming.emit('close');
+    clientRequest.emit('close');
+    // socket closed before response starts.
```
zherczeg: socket -> Socket

src/js/https_incoming.js
```diff
@@ -0,0 +1,242 @@
...
+
+  var parser = incoming.parser;
+  var clientRequest = incoming.clientRequest;
+
+  chunk = new Buffer(chunk);
+  var ret = parser.execute(chunk);
+
+  if (ret instanceof Error) {
+    parser.finish();
+    // unref all links to parser, make parser GCed
```
zherczeg: unref -> Unref

test/testsets.json
```diff
@@ -41,6 +41,9 @@
     { "name": "test_fs_open_read_sync_3.js", "skip": ["nuttx"], "reason": "not implemented for nuttx" },
     { "name": "test_gpio_input.js", "skip": ["all"], "reason": "needs hardware" },
     { "name": "test_gpio_output.js", "skip": ["all"], "reason": "need user input"},
+    { "name": "test_https_get.js", "timeout": 40 },
+    { "name": "test_https_request_response.js", "timeout": 40 },
+    { "name": "test_https_timeout.js", "timeout": 40 },
```
zherczeg: According to the doc, this module is supported only at tizen.

src/js/https_client.js
```diff
@@ -0,0 +1,156 @@
...
+
+    if (isNaN(chr2)) {
+      enc3 = enc4 = 64;
+    } else if (isNaN(chr3)) {
+      enc4 = 64;
+    }
+
+    output = output +
+    _keyStr.charAt(enc1) + _keyStr.charAt(enc2) +
+    _keyStr.charAt(enc3) + _keyStr.charAt(enc4);
```
zherczeg: Should be indented by two spaces.

yichoi:
@akhilkedia please check @zherczeg comments and rebase it to the master

docs/api/IoT.js-API-HTTPS.md
akhilkedia: Done

docs/api/IoT.js-API-HTTPS.md
akhilkedia: Done

docs/api/IoT.js-API-HTTPS.md
akhilkedia: Done. Listed all options here for completeness.

docs/api/IoT.js-API-HTTPS.md
akhilkedia: Done.

src/js/https_incoming.js
akhilkedia: Done

src/js/https_incoming.js
akhilkedia: Done.

src/js/https_incoming.js
akhilkedia: Done.

src/js/https_client.js
akhilkedia: Done.

test/testsets.json
akhilkedia: Since the HTTPS module is in the "exclude" of all platforms except Tizen, and Tizen is not a platform offered by Travis, these test cases will never actually run for any platform. 
These test cases are here just for reference, so they can be used to test manually if needed as shown in [https://github.com/Samsung/iotjs/blob/master/docs/build/Build-for-ARTIK.md#testing](https://github.com/Samsung/iotjs/blob/master/docs/build/Build-for-ARTIK.md#testing)
Hence there is no need for a `"skip": ["all"]`, though I could add that in the JSON if you say so.

akhilkedia:
@yichoi I have tested this on an `Rpi3` on the `headless Tizen` profile, and `Artik 10` on the `unified Tizen` profile.
All 3 tests passed. You can find the valgrind log for `Rpi3` here - [https://gist.github.com/akhilkedia/030818887499fb9a712c278165af9fc4](url)

src/modules/iotjs_module_https.c
```diff
@@ -0,0 +1,862 @@
...
+  iotjs_https_t* https_data = (iotjs_https_t*)userp;
+  IOTJS_VALIDATED_STRUCT_METHOD(iotjs_https_t, https_data);
+  size_t real_size = size * nmemb;
+  if (iotjs_jval_is_null(&_this->jthis_native))
+    return real_size - 1;
+  iotjs_jargs_t jarg = iotjs_jargs_create(1);
+  iotjs_jval_t jresult_arr = iotjs_jval_create_byte_array(real_size, contents);
+  // iotjs_string_t jresult_string =
+  //    iotjs_string_create_with_size(contents, real_size);
+  // iotjs_jargs_append_string(&jarg, &jresult_string);
```
hs0225: Do you need this comment?

src/modules/iotjs_module_https.c
```diff
@@ -0,0 +1,862 @@
...
+  // Use the jresult_arr Byte Array in production, but in testing use
+  // string. Uncomment out above line in testing.
+  iotjs_jargs_append_jval(&jarg, &jresult_arr);
+
+  bool result =
+      iotjs_https_jcallback(https_data, IOTJS_MAGIC_STRING_ONDATA, &jarg, true);
+
+  iotjs_jval_destroy(&jresult_arr);
+  // iotjs_string_destroy(&jresult_string);
```
hs0225: ditto.

build.config
```diff
@@ -98,19 +98,20 @@
...
       "nuttx": [],
-      "tizen": ["m", "rt"],
+      "tizen": ["m", "rt", "curl"],
       "tizenrt": []
     }
   },
   "module": {
     "always": ["buffer", "console", "events", "fs", "module", "timers"],
-    "include": ["assert", "dns", "http", "net", "stream", "testdriver"],
+    "include": ["assert", "dns", "http", "net", "stream", "testdriver", "https"],
```
akhilkedia: Having `https` in `include` made Travis CI atleast check if the `https` module builds (it cannot ofcourse run the tests for `Tizen`.)
I will remove it from the `includes` and `excludes` if the above Travis build check for `https` module is not needed.

build.config
```diff
@@ -98,19 +98,20 @@
...
       "all": [],
-      "linux": ["adc", "ble", "dgram", "gpio", "i2c", "pwm", "spi", "uart"],
-      "nuttx": ["adc", "dgram", "gpio", "i2c", "pwm", "stm32f4dis", "uart"],
-      "darwin": [],
-      "tizenrt": ["gpio", "pwm"]
+      "linux": ["adc", "ble", "dgram", "gpio", "i2c", "pwm", "spi", "uart", "https"],
+      "nuttx": ["adc", "dgram", "gpio", "i2c", "pwm", "stm32f4dis", "uart", "https"],
+      "darwin": ["https"],
+      "tizen": ["adc", "ble", "dgram", "gpio", "i2c", "pwm", "spi", "uart"],
+      "tizenrt": ["gpio", "pwm", "https"]
```
akhilkedia: Same as above.

src/modules/iotjs_module_https.c
akhilkedia: Removed the comment. Leftover debugging statement.

src/modules/iotjs_module_https.c
akhilkedia: Removed the comment. Leftover debugging statement.

src/js/https_client.js
```diff
@@ -0,0 +1,156 @@
...
+  var isMethodGood = false;
+  for (var key in methods) {
+    if (methods.hasOwnProperty(key)) {
+      if(this.method == methods[key]) {
+        isMethodGood = true;
+      }
+    }
+  }
+
+  if(!isMethodGood) {
```
akhilkedia: Done.

src/iotjs_module.h
```diff
@@ -52,7 +52,8 @@ typedef iotjs_jval_t (*register_func)();
   E(F, TCP, Tcp, tcp)                            \
   E(F, TIMER, Timer, timer)                      \
   E(F, UART, Uart, uart)                         \
-  E(F, UDP, Udp, udp)
+  E(F, UDP, Udp, udp)                            \
+  E(F, HTTPS, Https, https)
```
akhilkedia: Done.

src/js/https_client.js
```diff
@@ -0,0 +1,156 @@
...
+  this.host = protocol + '//' + host + ':' + port + path;
+  this.method = options.method || 'GET';
+  this.ca = options.ca || '';
+  this.cert = options.cert || '';
+  this.key = options.key || '';
+
+  var isMethodGood = false;
+  for (var key in methods) {
+    if (methods.hasOwnProperty(key)) {
+      if(this.method == methods[key]) {
```
akhilkedia: Done. I did not check manually because I assumed `tools/check-tidy.py` and the `clang-format` will take care of things like this.

test/testsets.json
glistening: @akhilkedia Currently Tizen is the only platform that support https. And we use `testsets.json` to identify which tests we can run and expected to pass for each platform. Without any description, we and other developers would think it can be run on all platform and supposed to pass.  Could you please specify "skip" and "reason"? I prefer to specify like following:

```
  { "name": "test_https_get.js", "timeout": 40 }, "skip": ["all"], "reason": "Only tizen support https, and on tizen we test manually" },
```

src/modules/iotjs_module_https.c
```diff
@@ -0,0 +1,862 @@
...
+  else if (strcmp(method, STRING_HEAD) == 0)
+    _this->method = HTTPS_HEAD;
+  else if (strcmp(method, STRING_CONNECT) == 0)
+    _this->method = HTTPS_CONNECT;
+  else if (strcmp(method, STRING_OPTIONS) == 0)
+    _this->method = HTTPS_OPTIONS;
+  else if (strcmp(method, STRING_TRACE) == 0)
+    _this->method = HTTPS_TRACE;
+  else {
+    // Will never reach here cuz checked in JS
```
glistening: @akhilkedia What about adding `assert(0)` here as you commented?

src/modules/iotjs_module_https.c
```diff
@@ -0,0 +1,862 @@
...
+  curl_multi_remove_handle(_this->curl_multi_handle, _this->curl_easy_handle);
+  curl_easy_cleanup(_this->curl_easy_handle);
+  _this->curl_easy_handle = NULL;
+  curl_multi_cleanup(_this->curl_multi_handle);
+  _this->curl_multi_handle = NULL;
+  curl_slist_free_all(_this->header_list);
+
+  if (_this->poll_data != NULL)
+    iotjs_https_poll_close_all(_this->poll_data);
+
```
glistening: @akhilkedia Please remove additional blank line.

daeyeon:
Please get commits merged into a single one.

src/modules/iotjs_module_https.c
```diff
@@ -0,0 +1,862 @@
...
+}
+
+// Write Callback is actually to read data from incoming response
+size_t iotjs_https_curl_write_callback(void* contents, size_t size,
+                                       size_t nmemb, void* userp) {
+  iotjs_https_t* https_data = (iotjs_https_t*)userp;
+  IOTJS_VALIDATED_STRUCT_METHOD(iotjs_https_t, https_data);
+  size_t real_size = size * nmemb;
+  if (iotjs_jval_is_null(&_this->jthis_native))
+    return real_size - 1;
```
glistening: @akhilkedia Could you explain me what does returning real_size - 1 for null instance means?

hs0225:
@akhilkedia I have one question. Can you combine `http_client.js` and `https_client.js` into one? `http_incoming.js` is the same.

src/modules/iotjs_module_https.c
akhilkedia: This function is a callback which `libcurl` sends us to write data (see  [https://curl.haxx.se/libcurl/c/CURLOPT_WRITEFUNCTION.html](https://curl.haxx.se/libcurl/c/CURLOPT_WRITEFUNCTION.html)). Qoute from the above link - 

> Your callback should return the number of bytes actually taken care of. If that amount differs from the amount passed to your callback function, it'll signal an error condition to the library. This will cause the transfer to get aborted.
Callback should return the number of bytes actually taken care of. If that amount differs from the amount passed to your callback function, it'll signal an error condition to the library. This will cause the transfer to get aborted.

The `real_size - 1` is used to signal an error in writing data to `libcurl`, and `real_size -1` will **always** differ from the "amount passed to your callback function" (which is `real_size`)

src/modules/iotjs_module_https.c
akhilkedia: Done.

src/modules/iotjs_module_https.c
akhilkedia: Done.

src/js/https_client.js
```diff
@@ -0,0 +1,156 @@
...
+        isMethodGood = true;
```
akhilkedia: Done.

akhilkedia:
@hs0225 I had considered merging `https_client` and `https_client` (and the same for `*_incoming`), but they have significantly different control flows.
The `http_*` rely **heavily** on `socket` API, which I did not implement for `https`. `https_*` rely **heavily** on the `libcurl` callbacks, the logic/control flow for which differ from those of `socket` significantly.
The ideal way to merge these would be perhaps using an `Agent` API as NodeJS does.

akhilkedia:
@yichoi I heard you wanted me to run _all_ tests with this patch.
Here are the results - [https://gist.github.com/akhilkedia/cbe625702cf8fbb8f20987a996109ad1](https://gist.github.com/akhilkedia/cbe625702cf8fbb8f20987a996109ad1).
I have added the results of the master branch as well as with this patch. The `https` tests pass, and all other tests remain unchanged.

Note that 3 tests, `test_dns_lookup`, `test_fs_mkdir_rmdir`, `test_events`, seem to fail on master branch, but these were probably due to incorrectly configured test environment (Tizen smack permissions and network config). If needed I can investigate these more and open a separate issue.

glistening:
@akhilkedia Thank you for update. Could you also see my comment on `testset.json`?

test/testsets.json
akhilkedia: Makes sense. Will do.

akhilkedia:
@glistening Done.

test/run_pass/test_https_timeout.js
```diff
@@ -0,0 +1,39 @@
+/* Copyright 2015-present Samsung Electronics Co., Ltd. and other contributors
```
yichoi: 2015->2017

yichoi:
Except this minor issue (2015->2017), LGTM

test/run_pass/test_https_timeout.js
akhilkedia: Done.

PR #935, opened 2017-06-13
Fix convertMode() in fs.js

bzsolt:
Parse mode-string as an octal number.


yichoi:
LGTM

PR #936, opened 2017-06-13
Remove test_stat.js

bzsolt:
Details: This test was skipped in all platform due to dynamically changing results,
which were not tested, basically this test didn't test anything.
The existing `test_fs_stat.js` covers all testable case except
the exception throwing, which part was added to it.


yichoi:
LGTM

PR #939, opened 2017-06-14
Remove Pause and Resume from HTTP parser.

zherczeg:
This is a feature of the http parser which is not used by IoT.js.
If it will ever be needed we can reintroduce them.


daeyeon:
@zherczeg please PR this to a brand new branch named `1.0`.

PR #940, opened 2017-06-14
Remove Pause and Resume from HTTP parser.

zherczeg:
This is a feature of the http parser which is not used by IoT.js.
If it will ever be needed we can reintroduce them.


daeyeon:
@zherczeg As for function coverage, the audit criteria has been changed after more discussions on internal unused functions. As the result, the new criteria confirmed is to meet 100% only for public functions in JS. So I think that we can close this. Sorry to have you confused.

PR #941, opened 2017-06-14
Remove unused functions from C files

DanielBallaSZTE:
Removing functions that were never used.


daeyeon:
@DanielBallaSZTE As for function coverage, the audit criteria has been changed after more discussions on internal unused functions. As the result, the new criteria confirmed is to meet 100% only for public functions in JS. So I think that we can close this. Sorry to have you confused.

PR #947, opened 2017-06-15
Fix test_net_10.js on stm32f4-nuttx target. (version 1)

rtakacs:
The `connect` function is synchronous on NuttX thus an error (non exist host) happens earlier than the required timeout.

### DIFF
--- a/test/run_pass/test_net_10.js
+++ b/test/run_pass/test_net_10.js
@@ -17,6 +17,7 @@ var net = require('net');
 var assert = require('assert');
 
 var timedout = false;
+var error = false;
 var connected = false;
 
 // Try connect to host that is not exist (Reserved address of TEST-NET-1)
@@ -30,7 +31,8 @@ socket1.on('timeout', function() {
 });
 
 socket1.on('error', function() {
-  assert.fail();
+  error = true;
+  socket1.destroy();
 });
 
 socket1.on('connect', function() {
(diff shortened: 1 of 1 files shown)


test/run_pass/test_net_10.js
```diff
@@ -39,6 +41,11 @@ socket1.on('connect', function() {
 });
 
 process.on('exit', function() {
-  assert(timedout);
+  if (process.platform == 'nuttx' || process.platform == 'tizenrt') {
```
daeyeon: Looks strange. I believe we should be able to execute the same code as possible on all supported platforms. Is this an inevitable problem?

zherczeg: The problem is that connect is synchronous on NUTTX. Perhaps we can invent something for this, but the test must be changed (hopefully in a platform independent way).

test/run_pass/test_net_10.js
```diff
@@ -30,7 +31,8 @@ socket1.on('timeout', function() {
 });
 
 socket1.on('error', function() {
-  assert.fail();
+  error = true;
```
daeyeon: Can we identify what kind of error comes in nuttx as the argument of the event handler? If we could use it, the following could be for thought.

```
error = false;

socket1.on('error',  function(e) {
      if(e.code == 'ErrorOccuredInNuttX') 
          error = true;
   }
}

process.on('exit', function() {
   assert(error || timedout);
   ..
}
```

yichoi:
resolved by #962

PR #949, opened 2017-06-15
Revive test_require.js

bzsolt:
Details:
 * Add this test to the testrunner as `test_module_require.js`.
 * Move module `process` related code into `test_process_readsource.js`.
 ** Move the used JSON-file to `test/resources/process/`.


test/run_pass/test_module_cache.js
```diff
@@ -22,8 +22,8 @@ module_cache = require('run_pass/require1/module_cache.js');
 assert.equal(module_cache.i, 100);
 
 {
-  var test1 = require('run_pass/require1/node_modules/test_pkg');
-  var test2 = require('run_pass/require1/node_modules/test_index');
+  var test1 = require('run_pass/require1/test_pkg');
```
daeyeon: add `assert` for test1 and test2.

test/run_pass/test_module_cache.js
```diff
@@ -22,8 +22,8 @@ module_cache = require('run_pass/require1/module_cache.js');
 assert.equal(module_cache.i, 100);
 
 {
```
daeyeon: Remove brackets

test/run_pass/test_module_cache.js
bzsolt: This part is already covered by the introduced `test_module_require.js`, so I've removed it. The invalid modules part is moved into `test_module_require.js` (because I think it doesn't belong to the module caching).

bzsolt:
Currently the `node_modules` only used explicitly, the implicit module loading is not tested yet. (I already have a patch to load modules under an `iotjs_modules` directory indirectly, but that patch depends on this PR.)
Furthermore the current directory structure is wrong.
Let suppose we would like to load the `require1` module. Based on the current behaviour the IoT.js looks for this module in `cwd/iotjs_modules/require1`, but the current folder-structure looks like `run_pass/require1/node_modules`, so the order of `require1` and `iotjs_modules` are swapped.

yichoi:
LGTM

PR #956, opened 2017-06-16
Cover a few missing functions

DanielBallaSZTE:
Covering some of the missing public functions, migrating parallel/test-net-keepalive.js.

### DIFF
--- a/test/node/parallel/test-net-keepalive.js
+++ b/test/node/parallel/test-net-keepalive.js
@@ -1,0 +1,65 @@
+/* Copyright 2017-present Samsung Electronics Co., Ltd. and other contributors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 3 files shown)


DanielBallaSZTE:
The test `test-net-bind-twice.js` seems to fail, somehow I was thinking it was my test that was failing, as the test system says it's failing, however it doesn't, the test before it fails.

DanielBallaSZTE:
As both @robertsipka and my tests are failing I assume that it is caused by the previous `test-net-bind-twice.js` not closing properly, (or in time) as both of our tests are net tests, and both cause that test to fail. I think a timeout in testsets.json would solve the issue, or putting it in the test itself would solve it either.

yichoi:
@DanielBallaSZTE  please rebase it to the master

DanielBallaSZTE:
Rebased to master

yichoi:
LGTM

PR #960, opened 2017-06-19
Make fs.write() and fs.writeSync() consistent

bzsolt:
Now fs.write() can accept `null` or `undefined` as `position` argument like fs.writeSync().


yichoi:
LGTM

PR #962, opened 2017-06-19
Rework test_net_10.js.

zherczeg:
Does not depend on lookup timeout anymore.


yichoi:
#947, #957 could be closed if this landed?

zherczeg:
Yes!

PR #963, opened 2017-06-19
Enable two additional process test

bzsolt:


bzsolt:
The output of `test_console.js` isn't checked also, but the test is allowed to run.
They run on Linux (on a desktop machine and on a RPi2 as well) and on NuttX platform properly.

yichoi:
LGTM

PR #977, opened 2017-06-21
Add a more secure type validator to TCP

robertsipka:


src/modules/iotjs_module_tcp.c
```diff
@@ -323,7 +316,7 @@ JHANDLER_FUNCTION(Connect) {
 
   if (err == 0) {
     // Get tcp wrapper from javascript socket object.
-    iotjs_tcpwrap_t* tcp_wrap = iotjs_tcpwrap_from_jobject(jtcp);
+    JHANDLER_DECLARE_THIS_PTR(tcpwrap, tcp_wrap);
```
LaszloLango: This macro contains a `return` statement, so this should be before declaration of `address` to avoid memory leak.

robertsipka: Thanks, I've updated it.

PR #978, opened 2017-06-21
Add a more secure type validator to BleHciSocket based on #911

LaszloLango:
Related issue: #816


LaszloLango:
BLE tests are on skip list by default, so this is only a speculative fix.

yichoi:
LGTM

PR #980, opened 2017-06-21
Add a more secure type validator to ADC module based on #911

DanielBallaSZTE:
I could not test the ADC module itself, as it is on the skiplist by default.
Manual testing doesn't reproduce unsafe access errors, however this is only a speculative fix.

### DIFF
--- a/src/modules/iotjs_module_adc.c
+++ b/src/modules/iotjs_module_adc.c
@@ -18,15 +18,15 @@
 #include "iotjs_objectwrap.h"
 
 
-static void iotjs_adc_destroy(iotjs_adc_t* adc);
-IOTJS_DEFINE_NATIVE_HANDLE_INFO(adc);
+IOTJS_DEFINE_NATIVE_HANDLE_INFO_THIS_MODULE(adc);
 static iotjs_adc_t* iotjs_adc_instance_from_jval(const iotjs_jval_t* jadc);
 
 
 static iotjs_adc_t* iotjs_adc_create(const iotjs_jval_t* jadc) {
   iotjs_adc_t* adc = IOTJS_ALLOC(iotjs_adc_t);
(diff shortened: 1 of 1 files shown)


src/modules/iotjs_module_adc.c
```diff
@@ -19,14 +19,15 @@
 
 
 static void iotjs_adc_destroy(iotjs_adc_t* adc);
```
LaszloLango: Remove this declaration.

src/modules/iotjs_module_adc.c
```diff
@@ -231,12 +232,11 @@ JHANDLER_FUNCTION(AdcConstructor) {
 
 
 JHANDLER_FUNCTION(Read) {
-  DJHANDLER_CHECK_THIS(object);
+  JHANDLER_DECLARE_THIS_PTR(adc, adc);
   DJHANDLER_CHECK_ARG_IF_EXIST(0, function);
 
   const iotjs_jval_t* jcallback = JHANDLER_GET_ARG_IF_EXIST(0, function);
   const iotjs_jval_t* jadc = JHANDLER_GET_THIS(object);
```
LaszloLango: You can write `const iotjs_jval_t* jadc = iotjs_jobjectwrap_jobject(adc);` instead, but this is also a good solution. @daeyeon What do you think? Which would be better?

DanielBallaSZTE:
@LaszloLango I've removed the function declaration as requested.
As for the other, I don't know which one should we use, as I used the one I'm using here in my previous commits I'll just leave it like that, so it matches the other ones.

LaszloLango:
@DanielBallaSZTE please check #993 Could that work here too?

DanielBallaSZTE:
@LaszloLango I think it works nicely, I think I'll change it on the gpio too.

yichoi:
LGTM

PR #982, opened 2017-06-21
Move process argument initialization to process internals

galpeter:
Main points of the change:
* Removed process.argv and process._initArgv from iotjs.js.
* Reworked InitArgv method to create the process.argv array.

The main motivation was to add object type check for the InitArgv
method but by moving the argv initialization into the C side
we can avoid the check and reduce the number of properties
for the process module.

Also this should reduce the memory usage a bit. As one
function property and one magic string was removed.


yichoi:
LGTM

PR #983, opened 2017-06-21
Add a more secure type validator to I2C module based on #911

LaszloLango:
Related issue: #816


LaszloLango:
I2C tests are on the skip list by default, so this is only a speculative fix.

yichoi:
LGTM

PR #985, opened 2017-06-21
Add a more secure type validator to SPI

robertsipka:
SPI tests are on the skip list by default, so this is only a speculative fix.

### DIFF
--- a/src/modules/iotjs_module_spi.c
+++ b/src/modules/iotjs_module_spi.c
@@ -23,15 +23,15 @@
 /*
  * SPI instance function
  */
-static void iotjs_spi_destroy(iotjs_spi_t* spi);
 static iotjs_spi_t* iotjs_spi_instance_from_jval(const iotjs_jval_t* jspi);
-IOTJS_DEFINE_NATIVE_HANDLE_INFO(spi);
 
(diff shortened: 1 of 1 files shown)


src/modules/iotjs_module_spi.c
```diff
@@ -342,18 +342,19 @@ JHANDLER_FUNCTION(SpiConstructor) {
...
-  const iotjs_jval_t* jspi = JHANDLER_GET_THIS(object);
-  iotjs_spi_t* spi = iotjs_spi_get_instance(jspi);
 
   iotjs_spi_set_array_buffer(spi, JHANDLER_GET_ARG(0, array),
                              JHANDLER_GET_ARG(1, array));
 
   if (jcallback) {
+    const iotjs_jval_t* jspi = JHANDLER_GET_THIS(object);
+
     SPI_ASYNC(transfer, jspi, jcallback, kSpiOpTransfer);
```
LaszloLango: I'd modify `iotjs_spi_reqwrap_create` to accept the `iotjs_spi_t*`. It will be used that way later.

robertsipka: Thanks, I've updated it accordingly.

yichoi:
LGTM

PR #990, opened 2017-06-22
Add a more secure type validator to UART module based on #911

knightburton:
The UART tests are on skip list by default, this is a speculative fix.


yichoi:
LGTM

PR #991, opened 2017-06-22
Fix module include/exclude priority logic.

### LINKED ISSUE #924: included / excluded modules configuration logic bug
There is commit **Rework the module checker logic** (#894) that changes how the logic works, but without update in build.config file. 
In this commit there is removed line that [exclude included modules from excluded list](https://github.com/Samsung/iotjs/commit/57b750f5a9d6f244be0c98d5a9c0820c48b23a62#diff-5dcd07ec20b9418df938cff319e62ac6L79). As a result if build.config file have excluded something, there is no way to include it with command line(with --iotjs-include-module).

We have to decide whether we want --iotjs-include-module have:
- higher priority than build.config exclude and add something like: 
`exclude_modules -= options.iotjs_include_module` 
- or lower and modify build.config and remove all modules from excluded that we want to have actually included.

zherczeg:
Command line include modules should have higher
priority than default exclude modules.

Fixes #924.

### DIFF
--- a/tools/module_analyzer.py
+++ b/tools/module_analyzer.py
@@ -39,11 +39,13 @@ def resolve_modules(options):
         build_modules_excludes |= set(
             options.config['module']['exclude'][system_os])
 
+    # Build options has higher priority than defaults
+    build_modules_excludes -= options.iotjs_include_module
+
     # By default the target included modules are:
     #  - always module set from the build config
     #  - modules specified by the command line argument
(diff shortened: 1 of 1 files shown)


LaszloLango:
#992 must fix the travis failures

zherczeg:
Lets wait #992 first.

LaszloLango:
#992 is landed. Please rebase the PR.

PR #993, opened 2017-06-22
Improve type validator in PWM module

robertsipka:
I've modified the iotjs_pwm_reqwrap_create function to accept the iotjs_pwm_t*.
PWM tests are on the skip list by default, so this is only a speculative fix.


yichoi:
LGTM

PR #1002, opened 2017-06-26
Correctly check number of cmd line arguments

galpeter:
If the IoT.js binary is started with one of the
`--memstat`, `--show-opcodes`, or `--start-debug-server` arguments
the program crashes when the process module initilizes its
argv array.

The reason for the crash is that the number of arguments
are checked before the previously mentioned arguments are
processed and after they processed the environment's argc
is set to 2, regardless that there is no valid filename/module
argument.

Test case:
```
./iotjs --memstat
```


yichoi:
LGTM

PR #1009, opened 2017-06-29
Adding an additional argument validation to Buffer module

DanielBallaSZTE:
This commit adds an additional check to Buffer module's Compare function.
Before it only checked if the given argument was an object, now it also verifies that the type is correct.

This is a sample patch only, if it succeeds I'll start implementing it in the other functions too.


yichoi:
LGTM

PR #1011, opened 2017-06-30
Remove module analyzer call from build.py

galpeter:
The module analyzer was used in the build.py
to build a list of modules which should be excluded
during test execution. However the test system now
knows which modules are built into the iotjs binary.


yichoi:
LGTM

PR #1012, opened 2017-06-30
Removed ADC JavaScript module wrapper in favor of exposing native ADC…

tosyu:
As the JS ADC wrapper was not really necessary, it was removed, and the native binding was exposed instead


LaszloLango:
Please rebase to the current master.

tosyu:
@LaszloLango PR rebased

kolipka:
@hs0225 
Important here is that this simplify API via removing unnecessary calls. If this proposition is OK, all other SystemIO APIs (UART, GPIO, SPI, I2C, PWM) should be changed in the same way.

src/modules/iotjs_module_adc.c
```diff
@@ -216,16 +198,32 @@ static void iotjs_adc_close_worker(uv_work_t* work_req) {
 
 JHANDLER_FUNCTION(AdcConstructor) {
   DJHANDLER_CHECK_THIS(object);
-  DJHANDLER_CHECK_ARGS(2, object, function);
```
tosyu: Because this dumps an unreadable error when funcion is not supplied, later in the code I check the param and supply a more userfriendly message when the funcion was not passed

tosyu:
@hs0225 about the callbacks, when I was rewriting fs module I have been informed that callback functions were mandatory (I did disagree) https://github.com/Samsung/iotjs/pull/839#issuecomment-301691435 so this change was made with that in mind and those callbacks are now mandatory

src/modules/iotjs_module_adc.c
```diff
@@ -216,16 +198,32 @@ static void iotjs_adc_close_worker(uv_work_t* work_req) {
...
+    JHANDLER_THROW(TYPE, "Bad arguments - configuration should be Object");
+    return;
+  }
+
+#if defined(__linux__)
+  DJHANDLER_GET_REQUIRED_CONF_VALUE(jconfiguration, _this->device,
+                                    IOTJS_MAGIC_STRING_DEVICE, string);
+#elif defined(__NUTTX__)
+  DJHANDLER_GET_REQUIRED_CONF_VALUE(jconfiguration, _this->pin,
+                                    IOTJS_MAGIC_STRING_DEVICE, number);
```
tosyu: done

src/modules/iotjs_module_adc.c
hs0225: You can use as below:
```js
  DJHANDLER_CHECK_ARGS(1, object);
  DJHANDLER_CHECK_ARG_IF_EXIST(1, function);
```

hs0225:
In case of FS, `callback` is mandatory because of Node.js compatibility. But ADC does not have to. `callback` is not always required on System I/O.
```js
gpio.close();
gpio.write(ture);
pwm.setPeriod(1);
pwm.setFrequency(1);
uart.write('');
...
```
I wanna change callback to option because of code usability.

src/modules/iotjs_module_adc.c
tosyu: This gives an error Internal Error (AdcConstructor), which in my opinion does not say why the error occured. I would rather stay with my jcallback check and error message

tosyu:
@hs0225 done, except for read which would be weird not to require a callback for a function that sole purpose is to put values in supplied callbacks :)

src/modules/iotjs_module_adc.c
```diff
@@ -216,17 +198,36 @@ static void iotjs_adc_close_worker(uv_work_t* work_req) {
...
+  DJHANDLER_GET_REQUIRED_CONF_VALUE(jconfiguration, _this->pin,
+                                    IOTJS_MAGIC_STRING_PIN, number);
+#endif
 
-  const iotjs_jval_t* jcallback = JHANDLER_GET_ARG(1, function);
-  ADC_ASYNC(open, adc, jcallback, kAdcOpOpen);
+  if (iotjs_jhandler_get_arg_length(jhandler) > 1) {
+    const iotjs_jval_t* jcallback = iotjs_jhandler_get_arg(jhandler, 1);
+    if (iotjs_jval_is_function(jcallback)) {
+      ADC_ASYNC(open, adc, jcallback, kAdcOpOpen);
```
hs0225: If there is no `jcallback`, ADC not open. Please change this line to execute even if it is not.

src/modules/iotjs_module_adc.c
```diff
@@ -257,15 +263,22 @@ JHANDLER_FUNCTION(Close) {
 
   if (jcallback) {
     ADC_ASYNC(close, adc, jcallback, kAdcOpClose);
-  } else {
-    if (!iotjs_adc_close(adc)) {
-      JHANDLER_THROW(COMMON, "ADC Close Error");
-    }
+  } else if (!iotjs_adc_close(adc)) {
```
hs0225: `adc.close` is async function. It should run asynchronously, even if there is no `jcallback`.

tosyu: ok, will fix it asap

src/modules/iotjs_module_adc.c
tosyu: ok, will fix it asap

yichoi: no more updates?

tosyu: This was updated and finished, I am waiting for you guys to review it again ;)

Issue #176, opened 2015-08-17
Complete process module


ILyoan:
https://github.com/Samsung/iotjs/wiki/IoT.js-API:-Process
- [ ] Documentation
- [ ] Implementation
  - [x] event: exit
  - [x] event: uncaughtException
  - [x] process.argv
  - [x] process.exit()
  - [x] process.exticode
  - [x] ~~process.version~~
  - [x] process.arch
  - [x] process.platform
  - [x] process.nextTick()

yichoi:
#1014 landed

PR #1014, opened 2017-07-03
Add process.version

glistening:
Now, process.version returns 1.0 since we've released 1.0 on June.


glistening:
It is related to #176.

PR #1019, opened 2017-07-04
Make the DECLARE_SIZE_T_FROM_DOUBLE macro a function

galpeter:
The macro DECLARE_SIZE_T_FROM_DOUBLE is defined in the
buffer module and used in multiple places.

By replacing the macro with a function we can reduce
the binary size a bit (almost 1 Kb).


yichoi:
LGTM

PR #1020, opened 2017-07-04
Add test result information to the README.md

rtakacs:
It is useful to see that all enabled tests are passed or not.


rtakacs:
The remote testrunner supports icons that related to the latest measurement. These are in a `status` folder where the `stm32f4dis.svg` and the `rpi2.svg` are either a pass or a fail icon.

yichoi:
LGTM

PR #1023, opened 2017-07-06
Change Tizen Build method and Support GBS build

sumin-lim:
Support GBS build, for following reasons.

1. Remove compiler specified lines.
 GBS uses own toolchain on the build. Thus, there should be no compiler specified lines on cmake file. If it contains, GBS returns compile error.
2. To use the latest dependencies based on Tizen latest packages. 
(http://download.tizen.org/snapshots/tizen/unified/latest/repos/standard/packages/)
3. Remove the test case from Travis.
This new profile will not run testsuite on Travis, because there is no specified compiler.

### DIFF
--- a/.gbs.conf
+++ b/.gbs.conf
@@ -1,0 +1,4 @@
+[general]
+upstream_branch = ${upstreamversion}
+upstream_tag = ${upstreamversion}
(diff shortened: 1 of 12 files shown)


sumin-lim:
Leave a comment

PR #1032, opened 2017-07-11
Move fs.Stats JS code into C side

galpeter:
For the creation of an fs.Stats object C and JS
codes were intermixed. By removing the JS parts
the code is simplified and the code size
will decrease by few hundred bytes.


yichoi:
```
Detected style problems:
/home/travis/build/Samsung/iotjs/src/iotjs_binding.c:276: line exceeds 80 characters
/home/travis/build/Samsung/iotjs/src/modules/iotjs_module_fs.c:537: line exceeds 80 characters
/home/travis/build/Samsung/iotjs/src/modules/iotjs_module_fs.c:538: line exceeds 80 characters
* total lines of code: 35224
* total non-blank lines of code: 28003
* style errors: 3
* clang-format errors: 2
* total errors: 5
Failed tidy check
```

daeyeon:
A  related basic rule has recently been created. Please refer to `Support Node.js API usages` section in [Native Module vs JS Module](https://github.com/Samsung/iotjs/blob/master/docs/devs/Native-Module-vs-JS-Module.md#when-to-write-js-module). In accordance with the rule, I think this should be closed.

galpeter:
@daeyeon I see. I've created the change before the document became available and didn't want to throw away the change before someone takes a look at it.

On a side note: I've also did a bit of memstat measuring and found out that with this PR the check_tests.js reports ~2 kb less peak allocated memory (on intel).

zherczeg:
Hm, as for me memory reduction has a higher priority than style.

daeyeon:
Personally my thoughts on this is same as you, but there are others who have concerns on the maintance. I respect the opinions also. As of now, the rule we have is a kind of compromise. We may eventually change it for memory reduction.

hs0225:
IMHO, it is important to reduce memory usage in our system. Except for APIs that are hard to maintain, I agree to move to C side. If the memory is reduced by ~2kb, your PR is okay for me.

PR #1034, opened 2017-07-12
artik053/tizenrt integration with Travis

tadziopazur:
Squashed into a single commit.
TizenRT build fix: invalid conversion in gpio platform module.
TizenRT build fix: wrong include order in pwm platform module.
Artik053/TizenRT Travis integration added.

PR #1037, opened 2017-07-13
Romfs patch for artik053

haesik:
- Currently, the size of romfs on artik053 is not decided yet, 
  We can use this patch temporary. 
  If TizenRT supports ROMFS officialy, I will delete this patch.
- Please copy the attached file into the TIzenRT root folder and patch it 
  $ patch -p0 < romfs.patch


yichoi:
I would like to specify `053` not `05x` in directory name since there will be no commercial product of `051`

haesik:
If I change the directory name, #1034 should also be changed...
And we are using artik05x name in two. 
First, we are using it on folder name, and second is on --target-board option when we build iotjs. 
So, I  think it's better to make another PR to chagne folder name and build option after landing #1034

PR #1041, opened 2017-07-14
Add a fix for http-catch-uncaughtexception test

daeyeon:

### DIFF
--- a/src/modules/iotjs_module_httpparser.c
+++ b/src/modules/iotjs_module_httpparser.c
@@ -213,7 +213,15 @@ static int iotjs_httpparserwrap_on_headers_complete(http_parser* parser) {
   iotjs_jargs_append_jval(&argv, &info);
 
   iotjs_jval_t res = iotjs_make_callback_with_result(&func, jobj, &argv);
-  bool ret = iotjs_jval_as_boolean(&res);
+
+  int ret = 1;
+  if (iotjs_jval_is_boolean(&res)) {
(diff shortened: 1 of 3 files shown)


src/modules/iotjs_module_httpparser.c
```diff
@@ -213,7 +213,13 @@ static int iotjs_httpparserwrap_on_headers_complete(http_parser* parser) {
   iotjs_jargs_append_jval(&argv, &info);
 
   iotjs_jval_t res = iotjs_make_callback_with_result(&func, jobj, &argv);
-  bool ret = iotjs_jval_as_boolean(&res);
+
+  // if exception throw occurs in iotjs_make_callback_with_result, then the
+  // result can be an object.
+  int ret = 1;
+  if (iotjs_jval_is_boolean(&res) || iotjs_jval_is_object(&res)) {
```
LaszloLango: Why don't you use `jerry_value_has_error_flag` instead?

glistening: @daeyeon If your intention is to check an error is occurred, I have same opinion with LaszloLango. There are some cases that we use `jerry_value_has_error_flag` directly.

daeyeon: The error_flag is cleared before leaving `iotjs_make_callback_with_result` out. And, I want to fix this with minimal changes.

yichoi:
LGTM

PR #1048, opened 2017-07-17
Add minimal heapsize finder tool

tkeri:
Added a python tool which finds the smallest possible size of
JerryHeap without failing to run the given js file.

IoT.js-DCO-1.0-Signed-off-by: Tamas Keri tkeri@inf.u-szeged.hu

### DIFF
--- a/tools/jerry_heaplimit_measure.py
+++ b/tools/jerry_heaplimit_measure.py
@@ -1,0 +1,129 @@
+#!/usr/bin/env python
+
+# Copyright 2017-present Samsung Electronics Co., Ltd. and other contributors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import argparse
+import json
+import os
+import subprocess
+import sys
+
+from subprocess import Popen, PIPE, STDOUT
+
+
+BIN_IOTJS_DEBUG = os.path.join(os.getcwd(),'build/x86_64-linux/debug/bin/iotjs')
+BIN_IOTJS_RELEASE = os.path.join(os.getcwd(),
+                                        'build/x86_64-linux/release/bin/iotjs')
+FLAG_BUILDTYPE = '--buildtype'
+FLAG_CLEAN = '--clean'
+FLAG_NO_TEST = '--no-check-test'
+FLAG_HEAPLIMIT = '--jerry-heaplimit'
+IOTJS_BUILDER = os.path.join(os.getcwd(), 'tools/build.py')
+TEST_DIR = os.path.join(os.getcwd(), 'test')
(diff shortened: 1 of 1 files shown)


tools/jerry_heaplimit_measure.py
```diff
@@ -0,0 +1,138 @@
...
+    if (low + hi) % 2:
+        return (low + hi) / 2
+    else:
+        return (low + hi + 1) / 2
+
+
+def build_bin(heapsize, opts):
+    """ Run tools/build.py script """
+    buildtype = opts.buildtype
+    hsize = str(heapsize)
```
rtakacs: I think the type conversation is not necessary.

tools/jerry_heaplimit_measure.py
```diff
@@ -0,0 +1,138 @@
...
+    except subprocess.CalledProcessError as err:
+        output = err.output
+        ret = err.returncode
+
+    return opts.testfile, output, ret
+
+
+def heap_limit(opts):
+    """ Find the minimal size of jerryheap to pass """
+    goodheap = int(opts.heapsize)
```
rtakacs: The type conversion is not necessary since you defined the type in the argument parser (type = int).

tools/jerry_heaplimit_measure.py
```diff
@@ -0,0 +1,138 @@
...
+    lowheap = int(0)
```
rtakacs: Unnecessary type conversion.

tools/jerry_heaplimit_measure.py
```diff
@@ -0,0 +1,138 @@
...
+    hiheap = int(opts.heapsize)
+
+    while lowheap < hiheap:
+        build_bin(hiheap, opts)
+
+        if opts.buildtype in ['debug']:
+            is_exe(BIN_IOTJS_DEBUG)
```
rtakacs: You should handle the case when there is no binary. You can add an assert for this line.

tools/jerry_heaplimit_measure.py
```diff
@@ -0,0 +1,138 @@
...
+        else:
+            is_exe(BIN_IOTJS_RELEASE)
```
rtakacs: Ditto.

tools/jerry_heaplimit_measure.py
tkeri: This one is it, because the subprocess.call() arg must contain only strings.

tools/jerry_heaplimit_measure.py
tkeri: Done.

tools/jerry_heaplimit_measure.py
tkeri: Done.

tools/jerry_heaplimit_measure.py
tkeri: Done.

tools/jerry_heaplimit_measure.py
tkeri: Ditto.

zherczeg:
This sounds like an interesting experiment, but do we need it in the main repository?

tkeri:
It could be useful tool.
For example, https://github.com/Samsung/iotjs/pull/1144. I used it in this PR.

Or the users could check their js files' size of JerryHeap.

yichoi:
I prefer this to be jerry repo

tkeri:
The jerry repo: https://github.com/jerryscript-project/jerryscript/pull/1991. So, I will close this this PR.

PR #1066, opened 2017-07-20
Enable ENABLE_MEMORY_CONSTRAINTS for http-parser

glistening:
Recently, we reduced the binary size of http-parser (Samsung/http-parser#6).
This patch enables this feature.


yichoi:
LGTM

PR #1067, opened 2017-07-20
fix TizenRT configuration

haesik:
We require to run 'make menuconfig' to setup the TizenRT,
With this patch, we can skip it..


yichoi:
```
tools/check_signed_off.sh --travis
Signed-off-by message is incorrect. The following line should be at the end of the 91639de commit's message: 'IoT.js-DCO-1.0-Signed-off-by: Haesik, Jun haesik.jun@samsung.com'. 
[Failed - 1] tools/check_signed_off.sh --travis
```

yichoi:
LGTM

PR #1065, opened 2017-07-20
TizenRT: skip tests to run without hung and coredump

kolipka:
+ test_spi.js file was added in order to comply with testsets.json

IoT.js-DCO-1.0-Signed-off-by: Konrad Lipner k.lipner@samsung.com

PR #1071, opened 2017-07-20
Added timeout for uart test to prevent hanging.

marcinsmsg:
In case the uart is not responding (for example due to lack of set up environment) after 10 seconds the test is skipped with TIMEOUT status.
The IoT.js test suite was run and checked for this test on Raspberry Pi 3. 

```
FAIL : test_stream.js
PASS : test_stream_duplex.js
PASS : test_timers_arguments.js
PASS : test_timers_error.js
PASS : test_timers_simple.js
TIMEOUT : test_uart.js                            # here after change
PASS : test_util.js


>>>> run_pass/issue
PASS : issue-133.js
PASS : issue-137.js
```

### DIFF
--- a/test/testsets.json
+++ b/test/testsets.json
@@ -91,7 +91,7 @@
     { "name": "test_timers_arguments.js" },
     { "name": "test_timers_error.js" },
     { "name": "test_timers_simple.js", "timeout": 10 },
-    { "name": "test_uart.js", "skip": ["all"], "reason": "need to setup test environment" },
+    { "name": "test_uart.js", "timeout": 10},
     { "name": "test_util.js" }
   ],
   "run_pass/issue": [


hs0225:
@marcinsmsg The UART test app skips on all platform because of setting the environment. Your PR now seems to have no meaning since test runner can not test the UART test app. I think it's better to add timeout after building System I/O auto test.

kolipka:
@hs0225 Currently I am asked to test SysIO along with submitting PR. First: all SysIO are skipped - this is bug, second some are broken (eg. no SPI test file, timeout on UART, etc). 
Our tests should work despite we have auto-test or not. auto-test solution needs it as a prerequisite.

glistening:
@kolipka 

> Currently I am asked to test SysIO along with submitting PR.

Yes, all system I/O PRs must be tested before submitting. Since there is no way to run automatically, it would be done manually for now.

> First: all SysIO are skipped - this is bug, second some are broken (eg. no SPI test file, timeout on UART, etc).

Yes, we know. It's the reason we asked you to prepare system I/O automation.

> Our tests should work despite we have auto-test or not. auto-test solution needs it as a prerequisite.

Could you please why this timeout is a prerequisite for auto-test solution?
I am guessing JS test driver hangs on test_uart.js on TizenRT. You want to get timeout so that JS test runner can run next tests. Is it right?

kolipka:
@glistening 
Yes you are right. I think that all tests involving some async actions should have timeouts. Perhaps a default timeout for all tests should be defined, and just extended in long running tests. Without it test suite can hung and requires manual actions. We will see test results before first hung only, and we can't submit such partial results to our webpage.

Actually we can (and we should) run SysIO tests automatically. UT should check whatever it can without external device. For reference please check [PR #1065 ](https://github.com/Samsung/iotjs/pull/1065/files#diff-e2477c4d3df8a43cfdc91bc1fe235209R1) and my first comment to this PR.

PR #1074, opened 2017-07-21
ADC implemented and tested for TizenRT port.

pmarcinkiew:
1. Pin field enabled for TizenRT in iotjs_module_adc.
2. New binding for TizenRT introduced based on sensorboard demo from
   TizenRT.

### DIFF
--- a/src/modules/iotjs_module_adc.h
+++ b/src/modules/iotjs_module_adc.h
@@ -34,7 +34,7 @@ typedef struct {
 
 #if defined(__linux__)
   iotjs_string_t device;
-#elif defined(__NUTTX__)
+#elif defined(__NUTTX__) || defined(__TIZENRT__)
   uint32_t pin;
 #endif
   int32_t device_fd;
--- a/src/platform/tizenrt/iotjs_module_adc-tizenrt.c
+++ b/src/platform/tizenrt/iotjs_module_adc-tizenrt.c
@@ -1,0 +1,96 @@
+/* Copyright 2017-present Samsung Electronics Co., Ltd. and other contributors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#if defined(__TIZENRT__)
+
+#include <errno.h>
+#include <fcntl.h>
+#include <stdlib.h>
+#include <sys/ioctl.h>
+#include <tinyara/analog/adc.h>
+#include <tinyara/analog/ioctl.h>
+
+#include "iotjs_def.h"
+#include "modules/iotjs_module_adc.h"
+
+#define S5J_ADC_MAX_CHANNELS 4
+
+// There is one file for all ADC inputs so wee need one common file descriptor
+static int32_t device_fd;
+// This is simple ref counter. Each time ADC is opened, it is increased.
+static size_t device_fd_counter = 0;
+// Path of ADC device.
+#define TIZENRT_ADC_DEVICE "/dev/adc0"
+
+int32_t iotjs_adc_read(iotjs_adc_t* adc) {
+  IOTJS_VALIDATED_STRUCT_METHOD(iotjs_adc_t, adc);
+  int ret, nbytes;
+  size_t readsize, i, nsamples;
+  struct adc_msg_s samples[S5J_ADC_MAX_CHANNELS];
+  ret = ioctl(_this->device_fd, ANIOC_TRIGGER, 0);
+  if (ret < 0) {
+    return -1;
+  }
+  readsize = sizeof(samples);
+  while (true) {
+    nbytes = read(_this->device_fd, samples, readsize);
+    if (nbytes > 0) {
+      nsamples = (size_t)nbytes / sizeof(struct adc_msg_s);
+      int sample = -1;
+      for (i = 0; i < nsamples; ++i) {
+        if (_this->pin == samples[i].am_channel) {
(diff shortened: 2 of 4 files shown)


src/modules/iotjs_module_adc.h
```diff
@@ -34,7 +34,11 @@ typedef struct {
 
 #if defined(__linux__)
   iotjs_string_t device;
-#elif defined(__NUTTX__)
+#endif
+#if defined(__NUTTX__)
+  uint32_t pin;
+#endif
+#if defined(__TIZENRT__)
```
LaszloLango:
```
#if defined(__NUTTX__) || defined(__TIZENRT__)
  uint32_t pin;
```

pmarcinkiew:
Would you be so kind to review my changes again?

I tested this commit again using test_adc.js with the latest TizenRT and it passed:

```
TASH>>iotjs test_adc.js
adc_setup: Enable the ADC interrupt: irq=106
ADC initialized
test1 start(read async test)
4008
4003
4007
4002
4007
test1 complete
adc_setup: Enable the ADC interrupt: irq=106
ADC initialized
test2 start(read sync test)
3657
3650
3650
3651
3653
test2 complete
iotjs thread end
TASH>>
```

I use potentiometer connected to ground and 3.3 VC from Artik053. The middle connector is providing signal to ADC0 input. I changed resistance for sync part of test and it is clear that output changes.

pmarcinkiew:
I use ROMfs feature from TizenRT to copy all IoT.js test into Artik053. Please see @kolopka 's PR [TizenRT#237](https://github.com/Samsung/TizenRT/pull/237). It contains description how to configure your Tizen RT. You should use genromfs to compress test files and flash them to the board.

src/platform/tizenrt/iotjs_module_adc-tizenrt.c
```diff
@@ -0,0 +1,96 @@
...
+
+bool iotjs_adc_close(iotjs_adc_t* adc) {
+  IOTJS_VALIDATED_STRUCT_METHOD(iotjs_adc_t, adc);
+  if (_this->device_fd > 0) {
+    device_fd_counter--;
+  }
+  if (0 == device_fd_counter) {
+    close(_this->device_fd);
+    device_fd = -1;
+    _this->device_fd = -1;
```
hs0225: Your PR is work well, but this code is right to be to the if statement above.

src/platform/tizenrt/iotjs_module_adc-tizenrt.c
```diff
@@ -0,0 +1,96 @@
...
+void iotjs_adc_open_worker(uv_work_t* work_req) {
+  ADC_WORKER_INIT;
+  IOTJS_VALIDATED_STRUCT_METHOD(iotjs_adc_t, adc);
+  if (0 == device_fd_counter) {
+    device_fd = open(TIZENRT_ADC_DEVICE, O_RDONLY);
+  }
+  _this->device_fd = device_fd;
+  if (_this->device_fd < 0) {
+    req_data->result = false;
+    _this->device_fd = -1;
```
hs0225: This seems unnecessary.

src/platform/tizenrt/iotjs_module_adc-tizenrt.c
```diff
@@ -0,0 +1,96 @@
...
+      int sample = -1;
+      for (i = 0; i < nsamples; ++i) {
+        if (_this->pin == samples[i].am_channel) {
+          sample = samples[i].am_data;
+        }
+      }
+      if (-1 != sample) {
+        return sample;
+      }
+    }
```
hs0225:
```js
for (i = 0; i < nsamples; ++i) {
  if (_this->pin == samples[i].am_channel) { 
    return samples[i].am_data;
  } 
}
```
As above, you can remove the `sample` variable.

src/modules/iotjs_module_adc.h
```diff
@@ -34,7 +34,8 @@ typedef struct {
 
 #if defined(__linux__)
   iotjs_string_t device;
-#elif defined(__NUTTX__)
+#endif
+#if defined(__NUTTX__) || defined(__TIZENRT__)
```
LaszloLango: `#elif` would be better here

src/platform/tizenrt/iotjs_module_adc-tizenrt.c
```diff
@@ -0,0 +1,96 @@
...
+  struct adc_msg_s samples[S5J_ADC_MAX_CHANNELS];
+  ret = ioctl(_this->device_fd, ANIOC_TRIGGER, 0);
+  if (ret < 0) {
+    return -1;
+  }
+  readsize = sizeof(samples);
+  while (true) {
+    nbytes = read(_this->device_fd, samples, readsize);
+    if (nbytes <= 0) {
+      // Loop
```
LaszloLango: Could you explain this? Why is this branch empty? Please write a more detailed comment at least. But I think the following style would be better:
```c
if (nbytes > 0) {
  nsamples = (size_t)nbytes / sizeof(struct adc_msg_s);
  int sample = -1;
  for (i = 0; i < nsamples; ++i) {
    if (_this->pin == samples[i].am_channel) {
      sample = samples[i].am_data;
    }
  }
  if (-1 != sample) {
    return sample;
  }
} /* else {
  // Some explanation of the missing else branch, maybe a TODO comment
} */

src/modules/iotjs_module_adc.h
pmarcinkiew: It is not any better. You are suggesting introducing dependency between not related lines. Why we should change TizenRT/Nuttix code if someone will decide that linux structure is not necessary or remove support for Linux?  An #elseif is not faster because precompiler generates header during compilation not at runtime, there is not final else with #error so there is not point in creating any relation between precompiler directives for Linux and Tizen.

LaszloLango: It is only a minor style issue. BTW, you wrote the `#elif` style in `src/modules/iotjs_module_adc.c`. I just made a quick search in the code base and we are using the same style for this in the whole code base. Please don't break it.

src/platform/tizenrt/iotjs_module_adc-tizenrt.c
hs0225: Do you have any updates on this?

yichoi:
LGTM

PR #1076, opened 2017-07-24
Speeding up the travis test

glistening:
This patch tries to reduce the travis test duration.
   
The `misc` job is spliited into two jobs:
{ default, minimal-profile } and { no-snapshot }
    
The `coverity` job seems strange because it runs all unit test.
Coverity is static anaylzer, so it should not require runtime test.
The coverity site also says just providing make command is enough.
    
### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,6 +27,7 @@ env:
     - OPTS="--test=artik10" INSTALL_TIZEN_DEPS=yes
     - OPTS="--test=artik053" INSTALL_TIZENRT_DEPS=yes
     - OPTS="--test=misc"
+    - OPTS="--test=no-snapshot"
 
 matrix:
   include:
--- a/tools/precommit.py
+++ b/tools/precommit.py
@@ -26,7 +26,7 @@ from common_py.system.executor import Executor as ex
 from common_py.system.platform import Platform
 from check_tidy import check_tidy
 
-TESTS=['host-linux', 'host-darwin', 'rpi2', 'nuttx', 'misc',
+TESTS=['host-linux', 'host-darwin', 'rpi2', 'nuttx', 'misc', 'no-snapshot',
        'artik10', 'artik053', 'coverity']
 BUILDTYPES=['debug', 'release']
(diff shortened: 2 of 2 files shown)


.travis.yml
```diff
@@ -27,6 +27,8 @@ env:
     - OPTS="--test=artik10" INSTALL_TIZEN_DEPS=yes
     - OPTS="--test=artik053" INSTALL_TIZENRT_DEPS=yes
     - OPTS="--test=misc"
+    - OPTS="--test=nosnapshot"
+    - OPTS="--test=minimalprofile"
```
LaszloLango: I'd write it with dashes to be similar to the build options (`no-snapshot` and `minimal-profile`).

LaszloLango:
> The coverity job seems strange because it runs all unit test.
> Coverity is static anaylzer, so it should not require runtime test.
> The coverity site also says just providing make command is enough.

Please check https://github.com/jerryscript-project/jerryscript/blob/master/.travis.yml#L39 and https://github.com/jerryscript-project/jerryscript/blob/master/.travis.yml#L71
This is how we solved this problem in JerryScript.

glistening:
@LaszloLango I also prefer the names you suggest. I was not certain that the travis allows dash in name. I will fix it.

For coverity scan, I will refer your link. Thank you.

LaszloLango:
@glistening I was thinking a bit more on this coverity issue. I think the correct fix would be this:

```diff
diff --git a/tools/precommit.py b/tools/precommit.py
index 93e7a06..f7047b5 100755
--- a/tools/precommit.py
+++ b/tools/precommit.py
@@ -255,4 +255,4 @@ for test in option.test:
         build("debug", ['--iotjs-minimal-profile'] + build_args)
 
     elif test == "coverity":
-        build("debug", os_dependency_module['linux'] + build_args)
+        build("debug", ['--no-check-test'] + os_dependency_module['linux'] + build_args)
```

glistening:
@LaszloLango I've added dash in name, and fixed 80 column rule, and I keep `--no-check-test` option since both you and I seem to be thinking same thing.

PR #1089, opened 2017-08-01
fix RPi3-Tizen guide document

haesik:
- fix RPi3-Tizen guide document
 - adding a guide of IP settings in profile 


docs/build/Build-for-RPi3-Tizen.md
```diff
@@ -105,8 +120,8 @@ sdb push ./test/run_pass/test_console.js /home/owner/iotjs/
 sdb root on
```
glistening: I like your convention of using (ubuntu) and (target) to specify on which machine I must type the command. Could you please add `(ubuntu) ` in front of sdb command? Also, for consistency, it would be better to change `(target)$` to `(target) $`. Please notice 1 space.

PR #1091, opened 2017-08-01
Skip the test_uart.js test file.

rtakacs:
This file fails on [nuttx-stm32f4 target](https://samsung.github.io/iotjs-test-results/). As you can see in the reason, the binary does not contain the `uart` module. But, when enabling `uart` module, the test is waiting for a test device (that will cause a timeout if the test device is not available).

Since test environment is not established everywhere, we should skip all similar test files.

### DIFF
--- a/test/run_pass/test_uart_api.js
+++ b/test/run_pass/test_uart_api.js
@@ -1,0 +1,24 @@
+/* Copyright 2017-present Samsung Electronics Co., Ltd. and other contributors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


bzsolt:
I think we should skip this test on linux as well. because `test_uart.js` also fails on rpi2 target.

hs0225:
I agree with @bzsolt .

rtakacs:
@bzsolt, @hs0225: you are right, I've updated the PR.

kolipka:
If I may: I would rather suggest changing the test similar manner to SPI (#1093). UART test should check compilance with API description and does OS does not throw exception when API calls are performed. As much as we can without need for external device. There should be another test, skipped, that require device.

Otherwise this test should be removed at all from the suite - since it require additional steps (adding test env), it should be skipped for all platforms, and just started manually on specially crafted environment. Please check test_pass/test_spi_mcp3008.js for reference.

zherczeg:
Can we do that check with the current infrastructure?

kolipka:
@zherczeg Yes, you can test all API except on('data') event. I can make fast PR for that it you want.

PR #1093, opened 2017-08-02
Add generic SPI test

kolipka:
This test check api existance and does OS calls fail.
There is no need for hardware, since response is not checked.
The test will be skipped only if SPI module is not included in IoT.js

### DIFF
--- a/test/run_pass/test_spi.js
+++ b/test/run_pass/test_spi.js
@@ -1,0 +1,80 @@
+/* Copyright 2017-present Samsung Electronics Co., Ltd. and other contributors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+var assert = require('assert');
+var Spi = require('spi');
+
+var spi = new Spi();
+
+var configuration = {};
+
+if (process.platform === 'linux') {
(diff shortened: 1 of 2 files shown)


zherczeg:
Does this test require hardware setup?

kolipka:
@zherczeg No, it only checkes does API exist and calls function that does not require any hardware attached(open, write). Actually I could verify also transfer result. I guess on most boards it will return 0xff, but on some boards it could require external pullup resistor, so input pin state will be floating.

hs0225:
@kolipka You should resolve test fail on Travis and [test-result](https://samsung.github.io/iotjs-test-results/).
```
TypeError: Cannot read property 'transfer' of undefined
FAIL : test_spi.js
```

kolipka:
@hs0225 The question is how?
spi.open looks like async function, but it is not. Callback is called from the inside, before it returns value. This is missleading. This is why in other tests we use this.transfer, while from setTimeout they call via spi1.

This is not described in docs, and I think API is invalid. So actually we need update API (remove SPI.open) and then describe properly with example use case how to use it.

test/run_pass/test_spi.js
```diff
@@ -0,0 +1,79 @@
...
+             'spi module does not export construction function');
+assert.assert(spi.MODE,
+              'spi module does not provide \'MODE\' property');
+assert.assert(spi.CHIPSELECT,
+              'spi module does not provide \'CHIPSELECT\' property');
+assert.assert(spi.BITORDER,
+              'spi module does not provide \'BITORDER\' property');
+assert.equal(typeof spi.open, 'function',
+             'spi does not provide \'open\' function');
+assert.equal(typeof spi1.transfer, 'function',
```
hs0225: `transfer` is called before `spi1` is declared.

kolipka:
@hs0225 Now there is:

> uncaughtException: AssertionError: spi.open failed: Error: Failed to export SPI device
> FAIL : test_spi.js

I suspect that travis linux will be differend than desktop linux and rpi2 linux. 
We can disable linux, or check is the linux + rpi board, and stop test if not.

hs0225:
@kolipka We need to solve this problem in different ways. Your test can't be solved by only Travis, so it would be better skip your test app.

PR #1094, opened 2017-08-02
Fix recursive load of modules.

### LINKED ISSUE #1077: Require recursive load.
Hello,

I have a problem with modules with the same name that loads itself recursively. There are 2 situations described below.

1. Core module vs local module
I have test file ./test/gpio.js
```javascript 
var gpio = require('gpio');
```
When run: script loads itself until recursion limit is reached.
```bash
iotjs ./test/gpio.js
```
I would expect that require will load gpio core module not local one, if I would want to load local module I would use: ` require('./gpio.js');`

In docs there are native first so I guess there must be some caching problem?

2. User module vs local module
I have create similar setup with user module:
* iotjs_modules/my-module/index.js
* test/my-module.js

In this case behavior is as documented in your docs. But in this situation I have no way of referencing  user modules ('iotjs_modules') from files of the same name.

In [Node.JS docs ](https://nodejs.org/api/modules.html#modules_all_together) there is described algorithm. When require path start from '/ ', './' or '../' then it tries to load local script otherwise it tries core module and then user modules.

This may be enhancement proposal.

Regards

LaszloLango:
Fixes #1077

### DIFF
--- a/src/js/module.js
+++ b/src/js/module.js
@@ -111,6 +111,9 @@ iotjs_module_t.resolveFilepath = function(id, directories) {
 
 
 iotjs_module_t.resolveModPath = function(id, parent) {
+  if (parent != null && id === parent.id) {
+    return false;
+  }
 
   // 0. resolve Directory for lookup
   var directories = iotjs_module_t.resolveDirectories(id, parent);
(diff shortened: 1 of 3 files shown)


kolipka:
@LaszloLango I am not sure this is proper solution for the #1077 .
require('gpio') should return core module, despite it is performed from local gpio.js file. This is according https://github.com/Samsung/iotjs/blob/master/docs/api/IoT.js-API-Module.md. Please check how Node.js solves module:
https://nodejs.org/api/modules.html#modules_all_together

If you want to keep this fix, module docs should be updated to reflect it.

LaszloLango:
@kolipka It loads tha native module first if it's built.

yichoi:
LGTM

PR #1096, opened 2017-08-03
Change socket error handling to fix #1078

### LINKED ISSUE #1078: Http request - server unreachable bug
Hello,
I have a problem with http request case when server is unreachable. 
Running test code bellow, doesn't return any error/exception when request is sent to unreachable host.
```
var http = require('http');

var request = http.request({
  host: 'localhost',
  port: 5000,
  path: '/',
  method: 'GET'
}, function (response) {
  response.on('data', function (chunk) {});
  response.on('end', function (chunk) {
    console.log('RESPONSE');
  });
  response.on('error', function (error) {
    console.log('RESPONSE', error);
  });
});
request.on('error', function (error) {
  console.log('REQUEST', error);
});
request.end();
```
Commenting the last line, hides(removes?) bug and example works properly. However, if I understand correctly API reference, this line is mandatory.

akhilkedia:
Fixes #1078 

If net module's socket errors, the pending writes are never finished, and as such the socket 'close' event is never fired. Since HTTPClient depends on the close event to fire its on error event if the request could not be started, this error event was actually never fired.

This patch fixes the above by making net module's socket "pretend" to write everything in its stream buffer and calls end so that the socket can call close as expected by the HTTPClient class.

This same approach is also used currently in the HTTPS module to handle this very case. Please feel free to suggest alternative ways of fixing this if there is a better way.

The older test case seems to have intentionally waited before calling request.end in an attempt to bypass this issue. Removed that wait.

### DIFF
--- a/src/js/net.js
+++ b/src/js/net.js
@@ -39,6 +39,7 @@ function SocketState(options) {
   this.readable = true;
 
   this.destroyed = false;
+  this.errored = false;
 
   this.allowHalfOpen = options && options.allowHalfOpen || false;
 }
@@ -139,7 +140,6 @@ Socket.prototype.write = function(data, callback) {
   if (!util.isString(data) && !util.isBuffer(data)) {
     throw new TypeError('invalid argument');
   }
-
   return stream.Duplex.prototype.write.call(this, data, callback);
 };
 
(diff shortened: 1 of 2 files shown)


src/js/net.js
```diff
@@ -150,16 +150,27 @@ Socket.prototype._write = function(chunk, callback, afterWrite) {
...
   var self = this;
 
-  resetSocketTimeout(self);
-
-  self._handle.owner = self;
-
-  self._handle.write(chunk, function(status) {
-    afterWrite(status);
+  if (self.errored) {
+    console.log('About to setTimeout for afterwrite');
```
kolipka: Please remove console.log

src/js/net.js
```diff
@@ -150,16 +150,27 @@ Socket.prototype._write = function(chunk, callback, afterWrite) {
...
+    process.nextTick(afterWrite, 1);
     if (util.isFunction(callback)) {
-      callback.call(self, status);
+      console.log('About to setTimeout for callback');
```
zherczeg: Please remove this condol.log as well.

src/js/net.js
akhilkedia: Oops. Will fix.

src/js/net.js
akhilkedia: Will fix.

yichoi:
LGTM

PR #1097, opened 2017-08-03
Adding error handling

haesik:
- adding error handling for iotjs_file_read()
- changing free()->iotjs_buffer_release()

### DIFF
--- a/src/iotjs_util.c
+++ b/src/iotjs_util.c
@@ -27,7 +27,10 @@
 
 iotjs_string_t iotjs_file_read(const char* path) {
   FILE* file = fopen(path, "rb");
-  IOTJS_ASSERT(file != NULL);
+  if (file == NULL) {
+    iotjs_string_t empty_content = iotjs_string_create();
+    return empty_content;
+  }
 
   fseek(file, 0, SEEK_END);
   long ftell_ret = ftell(file);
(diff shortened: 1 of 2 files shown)


src/iotjs_util.c
```diff
@@ -28,6 +28,10 @@
 iotjs_string_t iotjs_file_read(const char* path) {
   FILE* file = fopen(path, "rb");
   IOTJS_ASSERT(file != NULL);
+  if (file == NULL) {
```
zherczeg: Isn't this contradict the assert above? I think we should remove the assert.

haesik:
There is no test cases that cover the iotjs_file_read()

src/iotjs_util.c
```diff
@@ -27,7 +27,10 @@
 
 iotjs_string_t iotjs_file_read(const char* path) {
   FILE* file = fopen(path, "rb");
-  IOTJS_ASSERT(file != NULL);
+  if (file == NULL) {
+    iotjs_string_t empty_content = iotjs_string_create();
+    return empty_content;
+  }
```
glistening: Thank you for removing assert that could be false. I think we need to handle differently for actual empty content and `fopen` error. I found there is the only use case for iotjs_file_read for process.readsource. We need to add test case for this not-handled-till-now cases.

haesik: @glistening I found a test for this, test/run_pass/test_process_readsource.js. I am not sure that we need to distinguish empty content and fopen error. process.readsource is not a official API and it's only case that is used..

glistening: @haesik  For testcase, adding following case to `test_process_readsource.js` would be enough to cover your change. It causes segmentation fault.

```
process.readSource('notExist');
```

Although I still think we need to distinguish them (not existing source vs. existing and empty content), the fix to segmentation fault itself is good.

PR #1099, opened 2017-08-04
Remove regular expressions

jeuxdeau:
As the former version cannot turn off regular expression config in Jerryscript, I removed the codes relating to regular expressions. It can reduce memory usage of apps which do not use regular expressions.

IoT.js-DCO-1.0-Signed-off-by: Woobin Park keyxiz@gmail.com

### DIFF
--- a/src/js/ble_hci_socket_hci.js
+++ b/src/js/ble_hci_socket_hci.js
@@ -38,6 +38,7 @@ var debug = console.log; // require('debug')('ble_hci');
 
 var events = require('events');
 var util = require('util');
+var uuidUtil = require('ble_uuid_util');
 
 var BluetoothHciSocket = require('ble_hci_socket');
 
@@ -562,7 +563,7 @@ Hci.prototype.processCmdCompleteEvent = function(cmd, status, result) {
     this.emit('readLocalVersion', hciVer, hciRev, lmpVer, manufacturer, lmpSubVer);
   } else if (cmd === READ_BD_ADDR_CMD) {
     this.addressType = 'public';
-    this.address = result.toString('hex').match(/.{1,2}/g).reverse().join(':');
+    this.address = uuidUtil.reverseByteOrder(result.toString('hex'), ':');
 
     debug('address = ' + this.address);
 
@@ -605,7 +606,7 @@ Hci.prototype.processLeConnComplete = function(status, data) {
   var handle = data.readUInt16LE(0);
   var role = data.readUInt8(2);
   var addressType = data.readUInt8(3) === 0x01 ? 'random': 'public';
-  var address = data.slice(4, 10).toString('hex').match(/.{1,2}/g).reverse().join(':');
+  var address = uuidUtil.reverseByteOrder(data.slice(4, 10).toString('hex'), ':');
   var interval = data.readUInt16LE(10) * 1.25;
   var latency = data.readUInt16LE(12); // TODO: multiplier?
   var supervisionTimeout = data.readUInt16LE(14) * 10;
--- a/src/js/ble_uuid_util.js
+++ b/src/js/ble_uuid_util.js
@@ -36,8 +36,25 @@
 
 module.exports.removeDashes = function(uuid) {
   if (uuid) {
-    uuid = uuid.replace(/-/g, '');
+    uuid = uuid.split('-').join('');
   }
 
   return uuid;
 };
+
+module.exports.reverseByteOrder = function(str, joint) {
+  str = String(str);
+  var uuid = '';
+  var len = str.length;
+  var i = 0;
+
+  while (i < (len & ~0x1)) {
+    uuid = joint + str.slice(i, i + 2) + uuid;
+    i = i + 2;
+  }
+
+  if (len & 0x1) uuid = str.slice(i, i + 1) + uuid;
+  else if (joint !== '') uuid = uuid.substring(1, uuid.length);
+
+  return uuid;
+}
--- a/src/js/ble.js
+++ b/src/js/ble.js
@@ -39,7 +39,7 @@ var debug = console.log; // require('debug')('ble');
 var events = require('events');
 var util = require('util');
 
-var UuidUtil = require('ble_uuid_util');
+var uuidUtil = require('ble_uuid_util');
 
 var PrimaryService = require('ble_primary_service');
 var Characteristic = require('ble_characteristic');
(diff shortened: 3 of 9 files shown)


src/js/ble_uuid_util.js
```diff
@@ -36,8 +36,36 @@
...
+    while (end < uuid.length) {
+      if (uuid.charAt(end) !== '-') end++;
+      else {
+        str += uuid.slice(start, end);
+        start = end = end + 1;
+      }
+    }
+
+    str += uuid.slice(start, end);
   }
```
glistening: I think line 39 ~ line 52 can be replaced with `uuid.split('-').join('')`;

glistening:
@jeuxdeau Could you give me the binary size after disabling jerrsycript regexp with this patch (including base (default build) binary size?

jeuxdeau:
@glistening Thank you, I updated it. And here is it below:

* original: 617.4 kB (617,392 B)
* current: 617.2 kB (617,216 B)
* current(regex disabled): 590.2 kB (590,152 B)

build option: --no-check-test --buildtype release

src/js/ble_uuid_util.js
```diff
@@ -36,8 +36,23 @@
...
 module.exports.removeDashes = function(uuid) {
   if (uuid) {
-    uuid = uuid.replace(/-/g, '');
+    uuid = uuid.split('-').join('');
   }
 
   return uuid;
 };
+
+module.exports.tokenize = function(str) {
```
glistening: @jeuxdeau What about `splitAtEvery2ndChars`? I cannot know the expected working by the name of `tokenize`.

src/js/ble_uuid_util.js
```diff
@@ -36,8 +36,23 @@
...
+  var i = 0;
+
+  while (i < len - (len % 2)) {
+    arr.push(str.charAt(i) + str.charAt(i + 1));
+    i = i + 2;
+  }
+
+  if (i < len) arr.push(str.charAt(i));
+  return arr;
+}
```
zherczeg: It seems the code always uses .reverse().join('') after this call, perhaps we could do that here.

src/js/ble_hci_socket_hci.js
```diff
@@ -562,7 +563,7 @@ Hci.prototype.processCmdCompleteEvent = function(cmd, status, result) {
     this.emit('readLocalVersion', hciVer, hciRev, lmpVer, manufacturer, lmpSubVer);
   } else if (cmd === READ_BD_ADDR_CMD) {
     this.addressType = 'public';
-    this.address = result.toString('hex').match(/.{1,2}/g).reverse().join(':');
+    this.address = UuidUtil.tokenize(result.toString('hex'), ':');
```
glistening: Where can I find the definition of `UuidUtil.tokenize`?

src/js/ble_uuid_util.js
glistening: @jeuxdeau Could you explain why did you restore the code?

src/js/ble_uuid_util.js
```diff
@@ -36,8 +36,36 @@
...
+    }
+
+    str += uuid.slice(start, end);
   }
 
-  return uuid;
+  return str;
 };
+
+module.exports.refineUuid = function(str, joint) {
```
glistening: @jeuxdeau What do you mean by refineUuid? is it general term used in uuid? I can't guess what the function `refineUuid` does by the name.

src/js/ble_hci_socket_hci.js
jeuxdeau: @glistening I did it by mistake. I'll put it back. Thank you.

src/js/ble_uuid_util.js
jeuxdeau: @glistening I did it by mistake. I'll put it back. Thank you.

src/js/ble_uuid_util.js
jeuxdeau: Originally, Tokenize was a function that slices a string into two chars as you suggested. At the advice of @zherczeg, I combined Tokenize with reverse().join(sth). In so doing it got to need more comprehensive name not just representing matching regular expressions but also including its other operations. In this regard, Could you suggest an alternative again? I admit that refine is too ambiguous and not general for uuid as you think.

src/js/ble_hci_socket_hci.js
```diff
@@ -38,6 +38,7 @@ var debug = console.log; // require('debug')('ble_hci');
 
 var events = require('events');
 var util = require('util');
+var UuidUtil = require('ble_uuid_util');
```
zherczeg: The otherst start with lowercase letter.

src/js/ble_uuid_util.js
```diff
@@ -36,8 +36,23 @@
...
+  str = String(str);
+  var len = str.length;
+  var arr = [];
+  var i = 0;
+
+  while (i < len - (len % 2)) {
+    arr.push(str.charAt(i) + str.charAt(i + 1));
```
zherczeg: what about `str.slice(i, i+1)` ?

src/js/ble_uuid_util.js
```diff
@@ -36,8 +36,23 @@
...
   return uuid;
 };
+
+module.exports.refineUuid = function(str, joint) {
+  str = String(str);
+  var len = str.length;
+  var arr = [];
+  var i = 0;
+
+  while (i < len - (len % 2)) {
```
zherczeg: `(len & ~0x1)` ?

src/js/ble_uuid_util.js
```diff
@@ -36,8 +36,23 @@
...
+    arr.push(str.charAt(i) + str.charAt(i + 1));
+    i = i + 2;
+  }
+
+  if (i < len) arr.push(str.charAt(i));
```
zherczeg: `len & 0x1` ?

src/js/ble_uuid_util.js
```diff
@@ -36,8 +36,23 @@
...
+  return arr.reverse().join(joint);
```
zherczeg: What about creating the string on-the-fly without using an array? Should be faster and use less memory. You can also iterate back to front.

src/js/ble_uuid_util.js
glistening: @jeuxdeau `splitIntoTwoLetterSegmentsAndReverseAndJoin` or `reverseByteOrder`?

src/js/ble_uuid_util.js
```diff
@@ -36,8 +36,23 @@
...
 module.exports.removeDashes = function(uuid) {
   if (uuid) {
-    uuid = uuid.replace(/-/g, '');
+    uuid = uuid.split('-').join('');
   }
 
   return uuid;
 };
+
+module.exports.refineUuid = function(str, joint) {
```
glistening: @jeuxdeau What about `splitIntoTwoLetterSegmentsAndReverseAndJoin` or `reverseByteOrder`?

yichoi:
LGTM

PR #1088, opened 2017-07-30
Tizen peripheral UART

pmarcinkiew:
This commit requires Tizen GBS build script changes so it contains temporarily GPIO commit.
I will rebase it for master after GPIO will be merged.

PR #1086, opened 2017-07-27
GPIO peripheral API introduced for Tizen

pmarcinkiew:
1. The new platform module for Tizen was created using CAPI System
   Peripheral API.
2. The iotjs.spec was extended with peripheral dependencies.
3. The operating system flag for Tizen build was changed from
   Linux to Tizen. We use Tizen specific APIs for hardware
   so this change is reasonable.
4. The compiler flag __linux__ was used in many places to check for
   Linux specific features. This flag is set in compiler not in our
   build environment. We need a way to distinguish between
   pure Linux (e.g. Raspbian) and Tizen, which is very specific flavour.
5. Tizen GPIO was tested using test_gpio_xx test executed manually.

PR #1102, opened 2017-08-07
UART refactored for Tizen IoT support

pmarcinkiew:
1. UART fails to support close operations, which are specific
   for platform.
2. Some UART macros obfuscate code and fail to provide a strict type safe
   compilation. They where replaced with functions and structures.
3. Code was tested at Linux. Not merge YET.

### DIFF
--- a/src/modules/iotjs_module_uart.c
+++ b/src/modules/iotjs_module_uart.c
@@ -109,45 +109,25 @@ iotjs_uart_t* iotjs_uart_instance_from_reqwrap(THIS) {
 
 #undef THIS
 
-
-static bool iotjs_uart_close(iotjs_uart_t* uart) {
-  IOTJS_VALIDATED_STRUCT_METHOD(iotjs_uart_t, uart);
-  uv_poll_t* poll_handle = &_this->poll_handle;
-
-  if (_this->device_fd > 0) {
-    if (!uv_is_closing((uv_handle_t*)poll_handle)) {
-      uv_close((uv_handle_t*)poll_handle, NULL);
-    }
-    if (close(_this->device_fd) < 0) {
-      return false;
-    }
-  }
-
-  return true;
+void iotjs_uart_worker_init(uv_work_t* work_req,
+                            iotjs_worker_t* worker_fields) {
+  worker_fields->req_wrap = iotjs_uart_reqwrap_from_request(work_req);
+  worker_fields->req_data = iotjs_uart_reqwrap_data(worker_fields->req_wrap);
+  worker_fields->uart =
+      iotjs_uart_instance_from_reqwrap(worker_fields->req_wrap);
 }
 
 
 static void iotjs_uart_write_worker(uv_work_t* work_req) {
-  UART_WORKER_INIT;
+  iotjs_worker_t worker;
+  iotjs_uart_worker_init(work_req, &worker);
 
-  if (!iotjs_uart_write(uart)) {
-    req_data->result = false;
+  if (!iotjs_uart_platform_write(worker.uart)) {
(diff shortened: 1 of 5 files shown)


yichoi:
I wonder why this patch is needed. Please discuss with @daeyeon

pmarcinkiew:
We need new features in UART support for Tizen IoT. It is not based on SYSFS or any similar API but at custom Tizen solution. Please see https://github.com/Samsung/iotjs/pull/1086 for reference. In my opinion is better to refactor old UART and later add support for new platform in to separate commit.

daeyeon:
a) The problem is the working sequence. Since we don't have auto-testing infra to verify side-effect yet, someone needs to verify this for all the devices manually. However, for system I/O modules, I tend not to take any refactorying affecting other code that we already verify if it's working. For verifying this refactorying, let me have your prove. 

b) The requirement is to support uart for Tizen IoT. So I want you to first make a commit having working uart module for Tizen IoT based on the current code base with minimum changes. This refactorying may natural steps for that though.

c) Regarding refactorying,  

i. Function naming
You seemingly use `_platform_` to indicate functions needed to port. It's PRed for uart and gpio. Then, who care about other modules after you? Since it make others confused, it seems better to apply it with all related functions.

ii. macros 
I agree reducing macros in terms of readabiliy and memory.

iii. functions like worker_XXX or libtuv_XX_close
Can we make the functions as common for other modules?

d. Instead of mentioning 'Not merge YET'. Please put [WIP] onto the title if you pushed the commit you're working on.

daeyeon:
No response so far. I'd like to close this with the above reasons.

daeyeon:
After further discussion with pmarcinkiew, this commit has been reopened.
Please share your verification of this at each platform as suggested.

umaciek:
I verified this pull request on Raspberry, work correct, pass UART tests

src/modules/iotjs_module_uart.h
```diff
@@ -58,25 +58,34 @@ typedef struct {
   iotjs_uart_t* uart_instance;
 } IOTJS_VALIDATED_STRUCT(iotjs_uart_reqwrap_t);
 
-#define THIS iotjs_uart_reqwrap_t* uart_reqwrap
+typedef struct {
+  iotjs_uart_reqwrap_t* req_wrap;
+  iotjs_uart_reqdata_t* req_data;
+  iotjs_uart_t* uart;
+} iotjs_worker_fields;
```
LaszloLango: `iotjs_worker_t`

src/modules/iotjs_module_uart.c
```diff
@@ -109,45 +109,24 @@ iotjs_uart_t* iotjs_uart_instance_from_reqwrap(THIS) {
...
-    if (!uv_is_closing((uv_handle_t*)poll_handle)) {
-      uv_close((uv_handle_t*)poll_handle, NULL);
-    }
-    if (close(_this->device_fd) < 0) {
-      return false;
-    }
-  }
-
-  return true;
+void uart_worker_init(uv_work_t* work_req, iotjs_worker_fields* worker_fields) {
```
LaszloLango: Use the iotjs prefix for public functions: `iotjs_uart_worker_init`

pmarcinkiew:
@LaszloLango I modified file and structure names as you requested.

daeyeon:
The refactorying itself is ok, but the verification on it and merge requests are not done yet.

- The code doesn't work well as mentioned in #1088. 
- #1102 and #1088 were asked to be merged into one.

The above requests has not been resolved for too long time. #1088 and #1102 are closed for anyone who could take over them fast.

PR #1103, opened 2017-08-08
Use an own http server for testing

daeyeon:
This fixes the test failure or delay upon Travis.

The old test assumes that a web server runs on the localhost with default port. So, the requested socket could be hangged up when the web server is busy. That could cause delays or timeout especially on Travis.

### DIFF
--- a/test/run_pass/test_net_http_request_response.js
+++ b/test/run_pass/test_net_http_request_response.js
@@ -170,18 +170,26 @@ request5.end();
 
 
 // Test the IncomingMessage read function.
+var server6 = http.createServer(function(request, response) {
+  request.on('end', function() {
+    response.end('ok');
+    server6.close();
+  });
+}).listen(8080, 5);
+
 var readRequest = http.request({
-  host: 'localhost',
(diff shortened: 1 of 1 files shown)


test/run_pass/test_net_http_request_response.js
```diff
@@ -170,18 +170,26 @@ request5.end();
...
+  request.on('end', function() {
+    response.end('ok');
+    server6.close();
+  });
+}).listen(8080, 5);
+
 var readRequest = http.request({
-  host: 'localhost',
-  port: 80,
+  host: '127.0.0.1',
```
zherczeg: Isn't `localhost` is converted to `127.0.0.1` ?

zherczeg:
Add **n** to title: Use a**n** own http server for testing

test/run_pass/test_net_http_request_response.js
daeyeon: Right. I just wanted to directly use it.

yichoi:
LGTM

PR #1107, opened 2017-08-09
Add default timeout and skip flaky tests

bzsolt:
The default timeout is 240 seconds now.


tools/test_runner.js
```diff
@@ -18,6 +18,8 @@ var console_wrapper = require('common_js/module/console');
 var builtin_modules =
   Object.keys(process.native_sources).concat(Object.keys(process.binding));
 
+var default_timeout = 240; // (in seconds)
```
zherczeg: That is a pretty long timeout. A bad patch (where all tests are timed out) could make the test system run kind of forever.

bzsolt: Yes, it seems a pretty big timeout, but a debug binary + valgrind combination can produce 100+ sec runtimes on Travis CI (moreover tomorrow I've seen a 169 sec runtime), so this value based on this corner-cases, but ofc we can adjust it. Now there is no limitation for the time of a testrun unless the test specifies a certain timeout in `testsets.json`.

kolipka: @bzsolt Would it be possible to have it as a command line parameter? 240 is fine for me, but if not I could override it.

bzsolt: @kolipka great idea, I've added an additional option to `check_test.js`, please check it.

tools/common_js/option_parser.js
```diff
@@ -60,6 +62,11 @@ OptionParser.prototype.parse = function() {
     var val = arg_val[1];
     var found = false;
 
+    if (arg == "default-timeout") {
```
kolipka: == is error prone, please use ===

tools/test_runner.js
kolipka: @bzsolt looks awesome, thanks!

yichoi:
LGTM

PR #1108, opened 2017-08-10
TizenRT fix net problems

kolipka:
Seems that like in Nuttx TizenRT can't poll EOF.
Solves test_net_connect.js and test_net_2.js fails.

Current status:
PASS : 82
FAIL : 1
TIMEOUT : 0
SKIP : 39

IoT.js-DCO-1.0-Signed-off-by: Konrad Lipner k.lipner@samsung.com

PR #1112, opened 2017-08-11
TizenRT: turn on dns test

kolipka:
After recent net fixes(#1108) it works OK on TizenRT


daeyeon:
There are 15 tests of basic tests which are skipped, but must be passed for TizenRT. Will you create commits one by one in this way? I think it's better to make WIP PR.

kolipka:
@daeyeon TizenRT is based on Nuttx and therefore skip tests that are skipped in nuttx. There are 3 tests skipped exclusively for TizenRT: test_dns.js, test_iotjs_syntax_error.js and test-net-bind-twice.js. 

dns test works OK. The other 2 cause problems in TizenRT OS.  Solution might be trivial or not. For sure it will take a lot of time to fix it. If you want, I can make WIP PR, but it will just delay dns testing, and help with nothing.

yichoi:
LGTM

PR #1115, opened 2017-08-13
Restore process._exiting

bzsolt:
08f7833 works well with `check_test.js` because it has a different implementation
of `process.exit()`, but in case of manual runs when
an uncaughtException occurs fall into a loop.

### DIFF
--- a/src/js/iotjs.js
+++ b/src/js/iotjs.js
@@ -140,11 +140,15 @@ function _onUncaughtException(error) {
 
 
 process.exitCode = 0;
+process._exiting = false;
 process.emitExit = function(code) {
-  if (code || code == 0) {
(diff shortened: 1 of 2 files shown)


bzsolt:
This patch fixes the 3 process-related fails on test-results site.

daeyeon:
@bzsolt Can you  let me know what exact tests you're referring to?

bzsolt:
@daeyeon 
```
test_process_exitcode_arg.js
test_process_exitcode_var.js
test_process_explicit_exit.js
```
Just run them with an iotjs binary and because of the loop the iotjs will reach the reference count limit (ofc. only the debug bin prints it) and terminate.

PR #1119, opened 2017-08-16
Fix to remove dots from module path before caching.

akhilkedia:
Previously, several paths pointing to the same module could be cached
by module.js as it did not realize the absolute paths for them were the same.

This resulted in Significantly higher memory usage as the same module was
loaded repeatedly if 'required' by files in different folders.

### DIFF
--- a/src/js/module.js
+++ b/src/js/module.js
@@ -33,25 +33,27 @@ iotjs_module_t.wrap = Native.wrap;
 
 
 var cwd;
-try { cwd = process.cwd(); } catch (e) { }
+try {
+  cwd = process.cwd();
+} catch (e) { }
 
-var moduledirs = [""]
-if(cwd){
+var moduledirs = [""];
+if (cwd) {
   moduledirs.push(cwd + "/");
   moduledirs.push(cwd + "/iotjs_modules/");
 }
-if(process.env.HOME){
+if (process.env.HOME) {
   moduledirs.push(process.env.HOME + "/iotjs_modules/");
 }
-if(process.env.IOTJS_PATH){
-  moduledirs.push(process.env.IOTJS_PATH + "/iotjs_modules/")
+if (process.env.IOTJS_PATH) {
+  moduledirs.push(process.env.IOTJS_PATH + "/iotjs_modules/");
 }
 
 
 iotjs_module_t.resolveDirectories = function(id, parent) {
   var dirs = moduledirs;
-  if(parent) {
-    if(!parent.dirs){
+  if (parent) {
+    if (!parent.dirs) {
       parent.dirs = [];
     }
     dirs = parent.dirs.concat(dirs);
@@ -61,8 +63,7 @@ iotjs_module_t.resolveDirectories = function(id, parent) {
 
 
 iotjs_module_t.resolveFilepath = function(id, directories) {
-
-  for(var i = 0; i<directories.length ; i++) {
+  for (var i = 0; i < directories.length; i++) {
     var dir = directories[i];
     var modulePath = dir + id;
 
@@ -76,30 +77,30 @@ iotjs_module_t.resolveFilepath = function(id, directories) {
     // 1. 'id'
     var filepath = iotjs_module_t.tryPath(modulePath);
 
-    if(filepath){
+    if (filepath) {
       return filepath;
     }
 
     // 2. 'id.js'
     filepath = iotjs_module_t.tryPath(modulePath + '.js');
 
-    if(filepath){
+    if (filepath) {
       return filepath;
     }
 
     // 3. package path id/
     var jsonpath = modulePath + "/package.json";
     filepath = iotjs_module_t.tryPath(jsonpath);
-    if(filepath){
+    if (filepath) {
       var pkgSrc = process.readSource(jsonpath);
       var pkgMainFile = JSON.parse(pkgSrc).main;
       filepath = iotjs_module_t.tryPath(modulePath + "/" + pkgMainFile);
-      if(filepath){
+      if (filepath) {
         return filepath;
       }
       // index.js
       filepath = iotjs_module_t.tryPath(modulePath + "/" + "index.js");
-      if(filepath){
+      if (filepath) {
         return filepath;
       }
     }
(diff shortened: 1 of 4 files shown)


zherczeg:
Detecting same files is not a trivial problem. This seems to improve the situation, but not fully. What does nodejs do to detect same files?

akhilkedia:
@zherczeg I have no idea how nodejs does it - I came across the above problem while developing an external module, and doing this "improves" the situation, atleast enough to let my external module run.
Please feel free to suggest better methods to do this.

zherczeg:
https://linux.die.net/man/2/stat
`stat()` has an inode number. I think that is unique. Not sure Tizen supports it though.

glistening:
@akhilkedia I think you're trying to add a new API like nodejs's [path.normalize](https://nodejs.org/api/path.html#path_path_normalize_path). If so, what about `normalizePath`? And Could you also add test case and documentation? In general, we add test case and documentation for API.

akhilkedia:
@glistening Makes sense, I will rename pathRemoveDot() to normalizePath(). Also I will add a testCase for the same.

kolipka:
@glistening @akhilkedia Maybe we could introduce path module?
Nodejs seems to store real/absolute paths. I would advice being careful with stat/inode since platforms other than full-blown linux might not have such features (tizenrt/nuttx?).

glistening:
@kolipka For now, I am tentative for introducing `path` module since additional module requires more time and resources to maintain. And tizenrt and nuttx support `stat`. We provide `fs.stat` based on `stat` on tizenrt and nuttx.

kolipka:
@glistening sure, just keep in mind that TizenRT stat have it's quirks that we already had to workaround:
https://github.com/Samsung/TizenRT/issues/320 / https://github.com/Samsung/iotjs/blob/master/src/js/module.js#L69

glistening:
@kolipka Thank you for comment. There are several limitations on almost all APIs on tizenrt and nuttx. For filesystem, they seem to assume absolute path for every path parameter, it does not interpret dot or dot dot. `stat` is not exceptional. We need to implement and review carefully to provide relative path module resolution with limited support from os.

akhilkedia:
@glistening I've renamed the function to normalizePath. I have also added test cases for this.
I agree that maintaining a full paths module for just this one function is overkill, and its simpler to just add that function here and use it internally only.
Is the current approach of trying to get the absolute path acceptable? Or should I shift to use `stat` as suggested by @zherczeg (though @kolipka and @glistening have pointed out some problems with using `stat`)?

glistening:
@akhilkedia We cannot use `stat` on nuttx since there is no inode number in `struct stat` on nuttx. (I guess tizenrt is similar.)

```C
/****************************************************************************
 * Type Definitions
 ****************************************************************************/

/* This is the simplified struct stat as returned by stat() and fstat().
 * This structure provides information about a specific file or directory in
 * the file system. 
 */

struct stat
{
  mode_t    st_mode;    /* File type, atributes, and access mode bits */
  off_t     st_size;    /* Size of file/directory, in bytes */
  blksize_t st_blksize; /* Blocksize used for filesystem I/O */
  blkcnt_t  st_blocks;  /* Number of blocks allocated */
  time_t    st_atime;   /* Time of last access */
  time_t    st_mtime;   /* Time of last modification */
  time_t    st_ctime;   /* Time of last status change */
};
```

@zherczeg Do you have further idea under this limitation of OS?

Personally I think current approach is acceptable although it is not fully and perfect, but it alleviates memory consumption.

By the way, we need to prohibit navigating below current directory since it may look around all the file system by exploiting `..`. For example, when iotjs runs on Tizen, it may access file/directories that is out of iotjs.

zherczeg:
I was just curious. Ok, lets use the current approach.

glistening:
@akhilkedia We can use current approach, but I would like to check two more cases:

Could you add two more test cases?

(1) accessing under current working directory
 ( e.g. cwd is /home/iotjs/iotjsapp/ , and require('../../userA/somefile.js')

(2) trying to access less than root
 ( e.g. cwd is /home/iotjs/iotjsapp/ , and require('../../../../somefile.js')

I am not sure what is appropriate working for case (1). Personally, I think case (1) should not be permitted . Although userA may give permission for everyone, I would like to protect strictly since userA may give full permission by mistake. For case (2), it should throw some error.

akhilkedia:
@glistening 
For (2), I will implement the test case, and it should be straight-forward.

For (1), there are some issues - 
1. IoT.js check for modules in `$HOME/iotjs_modules` when loading modules. How should the case be handled when `$HOME` is below current working directory? (say if CWD is `$HOME/FolderA/FolderB/`)
2. Same issue as above with `$IOTJS_PATH/iotjs_modules`.
3. How do I handle the "user giving permission" for access below CWD? Via some command-line flag while launching IoT.js?

Perhaps its a better idea to let the OS's access control/permissions system to restrict what/how files can be accessed by IoT.js? Such as for Tizen specifically, trying to access files of root by using `..` will ideally result in a `EACCESS: Permission Denied`. Any file that the user can access by launching by commandline, could be accessed directly from the same command line.
For Tizen specifically, even `Internet` access requires a special privilege - But IoT.js does not restict access to the internet by itself. The `Tizen` OS itself restricts internet access for IoT.js until it is granted privileges. Perhaps the same approach should be used for files?

glistening:
@akhilkedia NodeJS seems to permit case (1).
```
├── brian
│   └── app.js
└── otherUser
     └── hello.js

brian:/home$ cd brian
brian:~$ cat app.js
require('../otherUser/hello.js');
brian:~$ cat ../otherUser/hello.js
console.log('Hello World');
brian:~$ node app.js
Hello World
```
LGTM after case (2) test case is added and works correctly.

akhilkedia:
@glistening I have added a test case for (2).
Note that this issue would have already thrown an error, since before my new function `normalizePath` gets called, but such a situtation never arises. This is because in the function `resolveFilepath` already handles the case of non-existent files/modules, and a module path below root is one such case.

glistening:
@akhilkedia I found that `normalizePath` accepts only existent path in current use case. However, someone may try to use `normalizePath` and passes input like case (2). Here is the output from current `normalizePath` and nodejs `path.normalize`.

```
$ cat t.js
...
console.log(normalizePath('/home/iotjs/../../hello'));
console.log(path.normalize('/home/iotjs/../../hello'));

console.log(normalizePath('/home/iotjs/../../../hello'));
console.log(path.normalize('/home/iotjs/../../../hello'));

console.log(normalizePath('/home/iotjs/../../../../hello'));
console.log(path.normalize('/home/iotjs/../../../../hello'));

$ node t.js
/hello
/hello
/../hello
/hello
/../../hello
/hello
```

`normalizePath` does not remove `..` which is below root.
Nodejs remove `..` below root silently.

I don't want to follow nodejs's current behavior. Because it is not documented in [`path.normalize`](https://nodejs.org/api/path.html#path_path_normalize_path), and it handles silently for exceptional case.

What about just adding comment in front of `normalizePath` what it assumes or returns for exceptional input case (2).

yichoi:
@akhilkedia https://travis-ci.org/Samsung/iotjs/jobs/268246470
please make this code tidy.

LaszloLango:
The patch contains many style fixes, but the commit message does not mention it. BTW, I'm happy to see such changes, but it would better to open a separate PR for them.

yichoi:
LGTM. since you seem busy, please make it distinguishable at next time as @LaszloLango mentioned

Issue #750, opened 2017-04-12
Confusing "exclude" module list in `build.config`


akosthekiss:
In `build.config`, modules can be listed as "always" (i.e., must be included in the build), "include" (i.e., also to be included in the build, unless overridden by command line option), and "exclude" -- and this category is what needs to be discussed. The problem is that actually the "exclude" modules of `build.config` are not excluded from anywhere (as far as building IoT.js goes): the config files "exclude" list _must not_ overlap with the "include" list, only the above mentioned command line option (`--iotjs-exclude-module`) can remove (exclude) modules from there.

However, `build.config`'s "exclude" list is used
- for checking that those modules are not even required (transitively) by included modules (in `module_analyzer.py`),
- for defining the `skip-module` option to the test driver (in `build.py`), and
- for composing the `--iotjs-include-module` (!!!) command line option for build tests (in `precommit.py`).

This leads to some interesting situations:
- In current master, several modules are listed as "excluded" just to ensure that they are build-tested but not run-tested. Forcing the build tests also forces the dummy implementation of some native modules even for platforms where they cannot and will not be implemented ever (otherwise the build would fail).
- With PR #730, "exclude" becomes platform-dependent. There, e.g., the darwin-specific "exclude" list is empty. This removes the need for dummy native module implementations, as no modules will be build-tested in addition to the "always" and "include" list. However, it will also remove any require checks and will not parametrize the test driver either. (Same situation for TizenRT.)

I'm still not sure whether 
- "exclude" is a misnomer and it should simply have a different name (e.g., "build-test"), or
- it is a mix of two different concepts (e.g., keep the "exclude" name for checking that some modules are not transitively required and skipped when tested; and have a new separate "build-test" list as well to be used by `precommit.py`)

The answer perhaps depends on whether both features are really needed or not.

So, the question to be discussed is what to do with "exclude"? (Or is it good as it is?)

glistening:
@akosthekiss I agree with you. We need to refine the build option ''exclude". We will start to discuss and follow up this issue after high priority task (due date next Friday) is done. Thank you for suggestion and thorough review about #730.

hs0225:
The word "exclude" seems to be confusing. And I don't think exclude is necessary. However, we need a list of modules suppoted by platform. So I suggest the follwing.

Before,
```
"module": {
    "always": ["buffer", "console", "events", "fs", "module", "timers"],
    "include": ["assert", "dns", "http", "net", "stream", "testdriver"],
[...]

PR #1120, opened 2017-08-17
Enable ADC/DGRAM/GPIO/PWM/UART modules for Tizen RT

yichoi:


glistening:
@yichoi I cannot understand this PR. The title says "Enable ADC/DGRAM/GPIO/PWM/UART modules for Tizen RT", but the change is adding these in `exclude` section.

glistening:
@hs0225 Thank you for explaining me offline. This PR is trying to build "ADC/DGRAM/GPIO/PWM/UART" in travis. To do that, we need to add this modules in `exclude`. It is surprising that we must add in `exclude` to enable something. It is related to #750.

PR #1122, opened 2017-08-18
Fix process.nextTick to accept parameters

kolipka:
Will be compliant with nodejs.

In IoT.js it is used like this here:
https://github.com/Samsung/iotjs/blob/master/src/js/net.js#L156


yichoi:
LGTM

PR #1124, opened 2017-08-18
tizen: Fix tagname to use project release- prefix

rzr:
Currently Git tag is not matching version (release_1.0),
so GBS must align to this scheme.

This is needed for "gbs export" to generate tarballs
from upstream git tag (with patches serie on top of it if present).

If tag is not in master branch, then a native tarball (without patches)
will be exported and built.


yichoi:
LGTM

PR #1128, opened 2017-08-21
Arrange the build config clearer

### LINKED ISSUE #750: Confusing "exclude" module list in `build.config`

daeyeon:
I divided `build.config` into three files below to resolve #750.

a) `build.module` describes the information about all the supported modules. It's used for travis build in precommit.py.
b) `build.target` describes various configurations which seems rarely be changed. Moving them to cmake and removing the file could be considered as a future task.
c) `build.config` describes user's configurable options. You can add modules you want to build with the include / exclude option. It's used for build.py.

There seems still things left to improve in the scripts. I wrote this as the first step to resolve the confusing option related to `exclude-module`. Plus, the options users rarely care about were parted. Lastly, with `build.module`, we can see all the supported features with dependencies at a glance.

### DIFF
--- a/tools/build.py
+++ b/tools/build.py
@@ -28,7 +28,6 @@ import re
 import os
 
 from js2c import js2c
-from module_analyzer import resolve_modules, analyze_module_dependency
 from common_py import path
 from common_py.system.filesystem import FileSystem as fs
 from common_py.system.executor import Executor as ex
@@ -36,6 +35,15 @@ from common_py.system.platform import Platform
 
 platform = Platform()
 
+def get_config(build_option_path):
+    config_path_list = [path.BUILD_TARGET_CONFIG_PATH,
+                        build_option_path]
+    result = {}
+    for cpath in config_path_list:
+        with open(cpath, 'rb') as f:
+            module = json.loads(f.read().decode('ascii'))
+            result.update(module)
+    return result
 
 # Initialize build options.
 def init_options():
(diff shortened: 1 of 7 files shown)


yichoi:
Please review first anyone except who is in HQ.
@akosthekiss ??

tools/precommit.py
```diff
@@ -158,107 +158,104 @@ def build(buildtype, args=[]):
...
+    out_module = {}
+    for os_name in dependent_module.keys():
+        out_module[os_name] = list(dependent_module[os_name])
+    return out_module
 
+if __name__ == '__main__':
+    option = parse_option()
+    config = get_config()
+    extend_module = config['module']['supported']['extended'];
+    os_dependency_module = get_os_dependent_module(extend_module);
```
daeyeon: done

tools/precommit.py
```diff
@@ -158,107 +158,104 @@ def build(buildtype, args=[]):
...
-    all_module = set(exclude_module['all'])
-    for os_name in exclude_module.keys():
-        if not os_name == 'all':
-            os_dependency_module[os_name] = \
-              list(all_module | set(exclude_module[os_name]))
-    return os_dependency_module
+def get_os_dependent_module(dependent_module):
+    out_module = {}
+    for os_name in dependent_module.keys():
+        out_module[os_name] = list(dependent_module[os_name])
```
daeyeon: fixed

tools/build.py
```diff
@@ -425,7 +437,7 @@ def run_checktest(options):
     build_args = ['quiet=' + checktest_quiet]
     if options.iotjs_exclude_module:
         skip_module = ','.join(options.iotjs_exclude_module)
-        build_args.append('skip-module=' + skip_module)
+        skip_module and build_args.append('skip-module=' + skip_module)
```
daeyeon: I intended that `skip-module=` isn't appended if skip_module is empty.

tools/build.py
```diff
@@ -46,23 +54,27 @@ def init_options():
...
-            elif isinstance(opt_val, int):
-                argv.append('--%s=%s' % (opt_key, opt_val))
-            elif isinstance(opt_val, list):
-                for val in opt_val:
-                    argv.append('--%s=%s' % (opt_key, val))
+
+    config_option = config['build_option']
+    for opt_key in config_option:
+        opt_val = config_option[opt_key]
+        list_with_commas = ['iotjs-include-module','iotjs-exclude-module']
```
daeyeon: done.

tools/build.py
```diff
@@ -28,14 +28,22 @@
...
 platform = Platform()
 
+def get_config(build_option_path):
+    config_path_list = [path.BUILD_TARGET_CONFIG_PATH,
+                        build_option_path]
+    result = []
+    for cpath in config_path_list:
+        with open(cpath, 'rb') as f:
+            module = json.loads(f.read().decode('ascii'))
+            result += module.items()
```
daeyeon: done.

daeyeon:
@galpeter Thank you for the review. I fixed your findings.

tools/build.py
daeyeon: Your reading is reasonable, but the line was called. And it was because `options.iotjs_exclude_module` was not filtered when its list is empty. I fixed it. Thank you.

galpeter:
lgtm (informal)

yichoi:
LGTM

PR #1129, opened 2017-08-21
[Enhancement] [HTTPS] Added support for option.rejectUnauthorized in HTTPS.request

akhilkedia:
**What this patch adds -**  This patch implements the option rejectUnauthorize from Node's HTTPS.request  - [https://nodejs.org/api/tls.html#tls_tls_connect_options_callback](https://nodejs.org/api/tls.html#tls_tls_connect_options_callback)

**Why -** This option is by default true, meaning certificates are verified and checked to 
make sure that the connection is to a correct server.

Setting this option to false will let an HTTPS connection proceed even if server's
certificate is incorrect or cannot be verified. Sometimes it may be desirable to let the
connection proceed even if the certificate is invalid, and this option allows that.

**Test Cases -** While this option was tested manually against a Node server, unlike
other tests, I was unable to find a web service which offered incorrect
certs. Hence no test case was added for this enhancement.

### DIFF
--- a/docs/api/IoT.js-API-HTTPS.md
+++ b/docs/api/IoT.js-API-HTTPS.md
@@ -14,16 +14,17 @@ IoT.js provides HTTPS to support HTTPS clients enabling users to send HTTPS requ
 
 ### https.request(options[, callback])
 * `options` {Object}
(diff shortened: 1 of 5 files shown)


yichoi:
LGTM

PR #1133, opened 2017-08-22
jerry: Propagate ENABLE_STATIC_LINK cmake flags

rzr:
This can be helful for packagers,
who prefer to turn this variable off to use use system's libc.

Currently it will only be used for iotjs code,
but kept static in jerryscript module, this is not wanted.


LaszloLango:
typo in commit message: `helpful`

galpeter:
There are two jerry targets in the cmake file. You have modified the host jerry build which is only for snapshot generation during build. The other jerry build is the target jerry which will be used by the iotjs. Is this what you wanted?

Also I think there is a typo in the commit message: `... use use ...`

yichoi:
@rzr any further comments?

daeyeon:
Closed because of no further response.

PR #1134, opened 2017-08-22
iotjs: Install binaries to system's path

rzr:
This will be helpful for packaging,
for example on Tizen install using RPM macro: "%make_install".


yichoi:
LGTM

PR #1135, opened 2017-08-22
[WIP] SPI peripheral API introduced for Tizen

pmarcinkiew:
1. New platform module for Tizen was created using CAPI System Peripheral API.
2. The iotjs.spec was extended with peripheral dependencies.
3. The operating system flag for Tizen build was changed from Linux to Tizen.
   We use Tizen specific APIs for hardware so this change is reasonable.
4. The compiler flag __linux__ was used in many places to check for Linux
   specific features. This flag is set in compiler not in IoT.js build
   environment. We need a way to distinguish between pure Linux (e.g. Raspbian)
   and Tizen, which is very specific flavour. It is better to move platform
   and API specific structures to platform files so ifdefs are not necessary.

IoT.js-DCO-1.0-Signed-off-by: Marcin m.ziombski@samsung.com


yichoi:
rebase please

hs0225:
Please check test app.
I found segmentation fault on rpi-linux.
```
pi@raspberrypi:~/hs$ ./iotjs test_spi.js
Segmentation fault
```

pmarcinkiew:
I fixed issue for Linux but still some further testing is necessary for Nuttx.

yichoi:
please show the final result of NuttX with this patch and post here

hs0225:
@pmarcinkiew Did the test pass on tizen?

yichoi:
please response by today or I will close and make my own patch.

hs0225:
@tosyu This PR is SPI code for tizen poring. Did you tested it on rp3-tizen?

tosyu:
@hs0225 no, I have verified it only on TizenRT as there were some changes in the implementation

pmarcinkiew:
We found some issues with IoT version, which need fixes.

pmarcinkiew:
JS layer checks platforms and expects that Linux will use device file name not bus:

https://github.com/Samsung/iotjs/blob/master/src/js/spi.js#L50

In my opinion this change has to be modified to use solution similar to GPIO:

https://github.com/Samsung/iotjs/blob/master/src/js/gpio.js#L51

We should refrain from enforcing platform depend checks in JS if it is not necessary.

yichoi:
I am closing since this patch has not passed travis for a long time. That is I assume that you don't have enough time or intention to resolve this situation to meet one of our major milestone the second internal release. If the priory of this patch would be changed. I will relocate proper person to handle this.

Issue #1126, opened 2017-08-18
Precommit script produces broken STM32/Nuttx port


tadziopazur:
Recent iotjs repository (aa96e33310, Wed Aug 16 03:25:04 2017) built for STM32F4Discovery does not support serial console over USB (CN5).
It does not support USART1 nor USART2.

Not sure if it boots at all, nothing indicates it does.

On the other hand, upstream nuttx, with no IOTSJ and attached configuration works like a charm (well, after an include is added and the project actually builds).

https://github.com/Samsung/iotjs/pull/1125

How is Nuttx port with IoT.js supposed to be built and run on STM32F4 Discovery1?

Tests appear to be generating reports, so I guess the builds are generated and do run on target.
https://samsung.github.io/iotjs-test-results/

bzsolt:
@tadziopazur  based on [this guide](https://github.com/Samsung/iotjs/blob/master/docs/build/Build-for-STM32F4-NuttX.md) I can build the current IoT.js master (1f29e9f) from the beginning to stm32f4/NuttX, I can login to the board via `minicom`, can run tests, so everything seems to be fine.

I've tried to create a nuttx binary with the `precommit.py --test=nuttx`, it's also works, but I couldn't flash this binary onto the board, I guess further configuration is required, but as I know this way is intended to testing the build itself for NuttX, not for testing.

tadziopazur:
Oh, that is disconcerting, I'd much prefer that tests actually run what is build in automated process.
Alas, the instruction is unclear and didn't allow me to obtain a workable build.

So far I have had issues building and running IoT.js on STM32F4, as it does not build/boot with just random configuration. In my case, whatever I did, the build wouldn't boot with either:
 - FPU enabled
 - ethernet enabled
 - SDIO (=mmc) enabled

To add insult to injury, IoT.js has hardcoded FPU support for nuttx target. Is there a canonical, _working_ nuttx configuration for the IoT.js? Could you please share that config file?

tadziopazur:
All great and exciting, but there is three pages worth of manual changes applied to the mentioned config listed in the guide, so the file you mentioned is not the answer to my question.
Also, applying those changes is error-prone and AFAIR cannot be done in a single pass.
Thus I ask for the ACTUAL configuration (config the nuttx has been built with) file.
Can you dig it up please?

bzsolt:
I've used this config file without any manual configuration and it worked for me.

tadziopazur:
For me it didn't.
[...]

PR #1138, opened 2017-08-23
Nuttx patches for iotjs automatic build

tadziopazur:
TL;DR: precommit --test=nuttx produces nuttx.bin that works

This change makes the precommit test action on nuttx relevant in that the results can be uploaded into STM32F4 Discovery1 and be run with IoT.js interpreter.

What it does is to patch the nuttx os (works with nuttx-7.19 tag) to fix compilation errors.
During build stm32f4discovery/usbnsh defconfig is overwritten with a working one.


pmarcinkiew:
LGTM

config/nuttx/stm32f4dis/patches/0004-Provide-working-iotjs-config-for-7.19-with.patch
```diff
@@ -0,0 +1,1818 @@
...
+ -ROMFS, TMPFS  -USB/NSH  -ADC, I2C, SPI, UART, GPIO, TCP/IP, UDP
+
+---
+ configs/stm32f4discovery/usbnshiotjs/Make.defs |  113 ++
+ configs/stm32f4discovery/usbnshiotjs/defconfig | 1589 ++++++++++++++++++++++++
+ configs/stm32f4discovery/usbnshiotjs/setenv.sh |   80 ++
+ 3 files changed, 1782 insertions(+)
+ create mode 100644 configs/stm32f4discovery/usbnshiotjs/Make.defs
+ create mode 100644 configs/stm32f4discovery/usbnshiotjs/defconfig
+ create mode 100755 configs/stm32f4discovery/usbnshiotjs/setenv.sh
```
hs0225: How about creating files and copying it without applying a patch file like tizenrt?

tadziopazur: I prefer a patch, because the contents, location and validation are kept in a single place.
With file copy, one needs to know:
 - where to copy the file,
 - should there be no file, or overwrite is okay?
And we lose all the validation that patch applying does (I prefer to know the patch does not apply, than just blindly overwrite a file that might have changed).

I would also like to replace our current apps/system/iotjs directory copy with a patch. Maybe later, definitely not in this patch.

hs0225: There are too many things in your patch files and it is hard to maintain. If the config file will be changed, how do you manage it? If the version of the nuttx is updated and config file is changed, you will need to delete contents in the patch files and copy text.
The role of your patch will make the file new, not change files. I want to separate it into three files.
And we have copied config file so far. And you should update the documents about the changes.

config/nuttx/stm32f4dis/patches/0001-Only-call-up_netinitialize-if-network-is-configured.patch
```diff
@@ -0,0 +1,25 @@
+From 662e30417dae4bae722a3408deae930a7917d19d Mon Sep 17 00:00:00 2001
+From: Tomasz Wozniak <t.wozniak@samsung.com>
```
hs0225: I don't want your name to remain here.

LaszloLango:
Could you summarize what issues fix this PR in the commit message?

LaszloLango:
It this PR related to #1126?

pmarcinkiew:
This PR was created as response to https://github.com/Samsung/iotjs/issues/1126

config/nuttx/stm32f4dis/patches/0004-Provide-working-iotjs-config-for-7.19-with.patch
tadziopazur: Moved the config file to config/nuttx/stm32f4dis/dot_usbnshiotjs_defconfig.
Removed nuttx patches 1 (network buildbreak fix) and 4 (usbnshiotjs configuration for stm32f4).
Still two patches are left (missing include in string include chain, missing definition for PWM/Timer configuration). Without these nuttx would not build with the provided config.

config/nuttx/stm32f4dis/patches/0001-Only-call-up_netinitialize-if-network-is-configured.patch
tadziopazur: Why?
In case of this file it no longer matters, since I removed the patch.

tadziopazur:
PTAL.

LaszloLango:
@pmarcinkiew in this case the issue shouldn't be closed. Please re-open it and mention the issue number in the commit message. Issues must be closed when they are resolved.

yichoi:
reopen #1126 since I advocate @LaszloLango 's opinion

yichoi:
This is not match current (.) started environment. I am afraid to back to make another test environment.  so please solve #1126 first then think about reopening this.

PR #1139, opened 2017-08-23
Simplify C part of the http-parser module

galpeter:
* Removed a IOTJS_VALIDATED_STRUCT_METHOD macro usage.
* Extracted Resume/Pause common code parts.
* Extracted parse error generation code part to a single method.

Reduces binary code size a bit.


glistening:
@galpeter `IOTJS_VALIDATE_STRUCT_METHOD` is used almost all modules sources. Could you remove `IOTJS_VALIDATED_STRUCT_METHOD` changes from this PR? Basically we would like to keep the validation check although it reduces readability.

glistening:
@galpeter Did you run related unit tests on NuttX and RPI also?

yichoi:
LGTM if test succeed

robertsipka:
I've tested on NuttX and RPi, the tests were passed except the `test_net_http_request_response.js` test which currently fails on master as well.

PR #1141, opened 2017-08-24
Terminate the engine gracefully

bzsolt:


bzsolt:
https://github.com/Samsung/iotjs/pull/1142 fixes timer leak.

src/iotjs.c
```diff
@@ -66,6 +66,10 @@ static bool iotjs_jerry_initialize(const iotjs_environment_t* env) {
   // Set magic strings.
   iotjs_register_jerry_magic_string();
 
+  // Register VM execution stop callback.
+  IOTJS_VALIDATED_STRUCT_METHOD(iotjs_environment_t, env);
+  jerry_set_vm_exec_stop_callback(vm_exec_stop_callback, &(_this->state), 6);
```
daeyeon: IMHO, we can handle exceptions in iotjs_uncaught_exception. Do we truely need to register this callback? Or, do you have any plan to use it for a certain purpose?

bzsolt: Yes we do, this callback is required for the vm-exec-stop feature of JerryScript, and allow to terminate the script execution when we want to abort that.
You can find a more detailed description about this feature [here](http://jerryscript.net/api-reference/#jerry_set_vm_exec_stop_callback).
(I've added a more verbose commit message too.)

daeyeon: @bzsolt Thanks for your explanation. One more thing, do you believe that 6 for `frequncy` is appropriate?

bzsolt: Previously I did some experiment and this value seemed to be enough, but now I found a testcase which requires to decrease the frequency to 2. Later we could fine-tune this value further if needed.

daeyeon:
LGTM. I love this patch.

yichoi:
LGTM

PR #1144, opened 2017-08-25
Replace isNull(arg) call

tkeri:
Deleted isNull(arg) function calls from
JavaScript sources and replaced these with JavaScript
comparison operator.

### DIFF
--- a/src/js/gpio.js
+++ b/src/js/gpio.js
@@ -113,7 +113,7 @@ function gpioPinOpen(configuration, callback) {
 
     process.on('exit', (function(self) {
       return function() {
-        if (!util.isNull(_binding)) {
+        if (_binding !== null) {
           self.closeSync();
         }
       };
@@ -125,7 +125,7 @@ function gpioPinOpen(configuration, callback) {
   GpioPin.prototype.write = function(value, callback) {
     var self = this;
 
-    if (util.isNull(_binding)) {
+    if (_binding === null) {
       throw new Error('GPIO pin is not opened');
     }
 
@@ -139,7 +139,7 @@ function gpioPinOpen(configuration, callback) {
   };
 
   GpioPin.prototype.writeSync = function(value) {
-    if (util.isNull(_binding)) {
+    if (_binding === null) {
       throw new Error('GPIO pin is not opened');
     }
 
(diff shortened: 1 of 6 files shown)


tkeri:
When I measured these tests the result was hectic. So, I try to do something with the big standard derivation. The reason of this big stddev is that we didn't have any influence when the GC will be called, and when the network driver answer to our query. So, my approach was to limit the maximum heap memory, and the expectation was that the stddev will be better in way. As you can see in the table, limiting the heap memory can lead to have a much more consequent results. Of course this method didn't solve any of the original problem, but the measurement and the results will be more consistent.

https://gist.github.com/tkeri/3a40d859b064a0f4857941647f9ba711
https://gist.github.com/tkeri/b0fc2e9712a4a7881958cad2641faa3e

tkeri:
**Tester: freya, build: release arm-rpi2 7c886d1**
===================
| testfile                                   | total       | peak        | max rel stdev   |
|--------------------------------------------|-------------|-------------|-----------------|
|                                            | diff / byte | diff / byte |      %          |
||||
| run_pass/test_net_7.js                     |             |             |                 |
| Jerry Heapsize: 256                        | +382        | -34631      |     9.45        |
| Jerry Heapsize: 87                         | +11         | -9689       |     11.73       |
| run_pass/test_net_http_request_response.js |             |             |                 |
| Jerry Heapsize: 256                        | +86         | -7549       |     6.24        |
| Jerry Heapsize:  53                        | +95         | +24         |     0.04        |
| run_pass/test_net_httpserver.js            |             |             |                 |
| Jerry Heapsize: 256                        | +78         | -279        |     5.64        |
| Jerry Heapsize:   47                       | +401        | -58         |     9.82        |
| run_pass/test_net_http_status_codes.js     |             |             |                 |
| Jerry Heapsize: 256                        | +38         | -196        |     7.41        |
| Jerry Heapsize:  60                        | +155        | +157        |     9.82        |

**NOTE:** The '+' means memory saving in bytes.

yichoi:
https://travis-ci.org/Samsung/iotjs/jobs/268347180
Your DCO is not correct. please check it again

yichoi:
@tkeri rebase please

yichoi:
LGTM

PR #1147, opened 2017-08-28
Add a nuttx config including all options of Build-for-STM32F4-NuttX.md

haesik:


tadziopazur:
Please rename to something not starting with a dot (.)
Unix (which linux is) treat files starting with a dot as hidden files and that might not be what we want.

Consider alloptions_defconfig, iotjs_alloptions_defconfig or dot_alloptions_defconfig.
.config.alloptions

Has this config been tested with iotjs and discovery board?

yichoi:
let's remove . at next time #1148

PR #1153, opened 2017-08-29
Fix config and app copying for tizenrt

tosyu:
The config and app folders should always be copied if files are newer on the iotjs repo folder. When changing options for tizenrt whe changes are not made in deps/tizenrt as the `precommit.py` made those only when the folder did not exist. This creates confusion.

This fix resolves that with least impact (only newer files are copied)


yichoi:
LGTM

PR #1154, opened 2017-08-29
Add trouble shooting for TizenRT

daeyeon:
This also removes `$` from the instrunctions for quick-copy and use.


yichoi:
LGTM

PR #1161, opened 2017-08-31
UDP.setBroadCast is enabled on TizenRT

glistening:
TizenRT support udp broadcast, so it is enabled.

Enable SetBroadCast on TizenRT since TizenRT supports udp broadcast.


glistening:
With this patch and patch #90 on libtuv, `test_dgram_broadcast.js` succeed in sending broadcast message on TizenRT.

PR #1162, opened 2017-08-31
Optimize localhost lookups for Nuttx/TizenRT

tosyu:
This optimization helps when running dgram module code (and tests) on TizenRT/nuttx as "localhost" is not resolved on those platforms to IP address.

### DIFF
--- a/src/modules/iotjs_module_tcp.c
+++ b/src/modules/iotjs_module_tcp.c
@@ -118,6 +118,16 @@ const iotjs_jval_t* iotjs_connect_reqwrap_jcallback(THIS) {
 
 #undef THIS
 
+iotjs_string_t iotjs_create_ip(const iotjs_string_t* address) {
+  const IOTJS_VALIDATED_STRUCT_METHOD(iotjs_string_t, address);
+#if defined(__NUTTX__) || defined(__TIZENRT__)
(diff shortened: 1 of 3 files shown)


daeyeon:
Is this a temporary workaround to conquer the tests in fail? What if a local dns lookup like `/etc/hosts` is intentionally changed? I'm not sure that we could ignore such a case.

tosyu:
@daeyeon this is just for Nuttx/TizenRT.

This solution is the same like in dns module:

https://github.com/Samsung/iotjs/blob/master/src/modules/iotjs_module_dns.c#L146 from PR #491 done by Wonyong Kim

pmarcinkiew:
In my opinion we should implement consistent DNS lookup. If there is additional logic for loopback in DNS, why it is wrong in UDP?

zherczeg:
This should be a workaround, and should be improved later. But could go in now.

daeyeon:
@pmarcinkiew If so, why not removing the logic in DNS instead? 

I just want to know why we should accept this workaround now. Is there any tangible benefit we can have for now? e.g) a certain unit test passes.

daeyeon:
I'd like to close this because, as of no response, it seems not urgent to accept this workaround.

PR #1167, opened 2017-09-04
Fix jerry-debugger on IoT.js build

DanielBallaSZTE:
On 18th, July the Jerry Debugger API has changed, therefore the debugger-server couldn't start, because the patch introduced runtime port configuration, making the current jerry debugger call broken.
This patch fixes it.
Passing the environment value is needed, since now Jerry won't do the cleanup itself, it needs to be manually called from IoT.js.

### DIFF
--- a/src/iotjs.c
+++ b/src/iotjs.c
@@ -51,13 +51,12 @@ static bool iotjs_jerry_initialize(iotjs_environment_t* env) {
     jerry_port_default_set_log_level(JERRY_LOG_LEVEL_DEBUG);
 #endif
   }
-
-  if (iotjs_environment_config(env)->debugger) {
-    jerry_flags |= JERRY_INIT_DEBUGGER;
-  }
-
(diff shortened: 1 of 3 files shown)


DanielBallaSZTE:
I don't know about this failing job though.

pmarcinkiew:
TizenRT build was broken recently due to TizenRT changes. We are investigating issue.

pmarcinkiew:
See:
https://github.com/Samsung/iotjs/issues/1168

DanielBallaSZTE:
@pmarcinkiew Yes, now I see, thanks!

pmarcinkiew:
See: https://github.com/Samsung/iotjs/pull/1169

yichoi:
@pmarcinkiew You are supposed to stabilize the build environment of travis first

yichoi:
LGTM

PR #1170, opened 2017-09-04
Dgram tests enabled

pmarcinkiew:
The some data gram tests were enabled. They pass for standard Linux environment.


yichoi:
You are supposed to stabilize the build environment of travis first

pmarcinkiew:
Travis is not stable enough to enable more tests.

PR #1173, opened 2017-09-05
Fix module.require error throwing on Linux

glistening:
Currently module loading throw error below root when it should not.

$ iotjs ../some.js

uncaughtException: Error: Requested path is below root:


LaszloLango:
@glistening Is the comment still valid? I think you had to remove the comments too if it is not a temporary fix for TizenRT anymore.

glistening:
@LaszloLango I will remove the comments that is not valid any more next PR. Thank you!

PR #1174, opened 2017-09-05
Enable setTTL on TizenRT and refine setTTL test case

glistening:
- Enable setTTL on TizenRT
- Previously, we have to adjust setTTL, and run the test case again.
Now, it is done by one execution.


glistening:
Samsung/libtuv#93 has to be landed first.

yichoi:
LGTM

PR #1177, opened 2017-09-05
Fix test_module_require.js err on TizenRT

haesik:
- to support relative path parameter in require('../add2')

### DIFF
--- a/src/js/module.js
+++ b/src/js/module.js
@@ -67,12 +67,13 @@ iotjs_module_t.resolveFilepath = function(id, directories) {
     var dir = directories[i];
     var modulePath = dir + id;
 
-    // START: Temprorary fix for TizenRT
-    // See: https://github.com/Samsung/TizenRT/issues/320
     if (modulePath[0] !== '/') {
       modulePath = process.cwd() + '/' + modulePath;
     }
-    // END: Temprorary fix for TizenRT
+
+    if (process.platform === 'tizenrt' && modulePath.indexOf("..") != -1) {
+      modulePath = iotjs_module_t.normalizePath(modulePath);
+    }
 
     // 1. 'id'
     var filepath = iotjs_module_t.tryPath(modulePath);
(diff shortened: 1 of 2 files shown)


src/js/module.js
```diff
@@ -67,12 +67,14 @@ iotjs_module_t.resolveFilepath = function(id, directories) {
     var dir = directories[i];
     var modulePath = dir + id;
 
-    // START: Temprorary fix for TizenRT
     // See: https://github.com/Samsung/TizenRT/issues/320
```
LaszloLango: This line also must be removed if this isn't a temporary fix anymore.

src/js/module.js
```diff
@@ -67,12 +67,13 @@ iotjs_module_t.resolveFilepath = function(id, directories) {
...
     var modulePath = dir + id;
 
-    // START: Temprorary fix for TizenRT
-    // See: https://github.com/Samsung/TizenRT/issues/320
     if (modulePath[0] !== '/') {
       modulePath = process.cwd() + '/' + modulePath;
     }
-    // END: Temprorary fix for TizenRT
+
+    if (process.platform === 'tizenrt' && modulePath.indexOf("..") != -1) {
```
glistening: @haesik What about removing this condition and `normalizePath` in resolveModPath? `normalizePath` seems to be called for all platforms. For `tizenrt`, it is needed because it does not support `..`. For tizen (and perhaps linux), it should be normalized to avoid duplicated cache. We can call `normalizePath` in common entry point.

haesik: If you want to remove the condition we need to test for all system and I think it's better to use linux system call for linux.. but now I just want to test only TizenRT.

glistening: > If you want to remove the condition we need to test for all system
If you added `process.platform === tizenrt` because your PR is tested only in tizenrt, we can improve later PR.

> I think it's better to use linux system call for linux.
IMHO, I am not sure it is bottleneck. Furthermore, I am not sure how much linux system call would be faster.

Anyway, for this stage, LGTM

Issue #1182, opened 2017-09-07
Tizen GBS compilation fails with missing python, cmake


pmarcinkiew:
The current master for IoT.js fails to compile in GBS

Reproduction steps

1. Copy gbsbuild.sh:
    $ cp config/tizen/gbsbuild.sh .
2. run gbsbuild:
   $ ./gbsbuild.sh

Expected result:
RPM compiled for IoT.js

Result:
```
info: *** Build Status Summary ***
=== the following packages failed to build due to missing build dependencies (1) ===
iotjs:
  nothing provides python
  nothing provides cmake
  nothing provides glibc-static
```

Full list of missing packages:
```
piotr@bdell:~/src/iotjs_tizen_gbs$ gbs -c config/tizen/sample.gbs.conf build -A armv7l --include-all --clean
gbs:warning: Created a new config file /home/piotr/.gbs.conf. Please check and edit your authentication information.
info: generate repositories ...
warning: No local package repository for arch armv7l
info: build conf has been downloaded at:
      /var/tmp/piotr-gbs/tizen_unified.conf
info: start building packages from: /home/piotr/src/iotjs_tizen_gbs (git)
2017-09-07 12:54 +0200
gbs 0.24.6
info: prepare sources...
info: start export source from: /home/piotr/src/iotjs_tizen_gbs ...
info: Creating (native) source archive iotjs-1.0.0.tar.gz from 'cb4882675547da813ef0804bb518c969a7347e11'
info: package files have been exported to:
     /home/piotr/GBS-ROOT/local/sources/tizen_unified/iotjs-1.0.0-0
info: retrieving repo metadata...
info: parsing package data...
info: building repo metadata ...
info: resolving skipped packages ...
info: package dependency resolving ...
info: *** Build Status Summary ***
=== the following packages failed to build due to missing build dependencies (1) ===
iotjs:
  nothing provides python
  nothing provides cmake
  nothing provides glibc-static
  nothing provides ld-linux.so.3 needed by aul
  nothing provides ld-linux.so.3(GLIBC_2.4) needed by aul
  nothing provides libgcc_s.so.1 needed by aul
  nothing provides libgcc_s.so.1(GCC_3.5) needed by aul
  nothing provides libpthread.so.0 needed by aul
  nothing provides libpthread.so.0(GLIBC_2.4) needed by aul
  nothing provides libsqlite3.so.0 needed by aul
  nothing provides libxml2.so.2 needed by aul
[...]

Issue #1178, opened 2017-09-05
Broken links in Build-for-RPi3-Tizen doc


shulgaalexey:
Hi @sumin-lim 

Pls check hyperlinks in **Downloading TizenIoT Core Image for RPi3** section of docs https://github.com/Samsung/iotjs/blob/master/docs/build/Build-for-RPi3-Tizen.md#downloading-tizeniot-core-image-for-rpi3 
master branch

Seems both are out of date

Thanks

glistening:
@shulgaalexey Thank you for informing us. I've asked @sumin-lim. I hope http://download.tizen.org/snapshots/tizen/unified/latest/images/standard/iot-headless-2parts-armv7l-rpi3/ and http://download.tizen.org/snapshots/tizen/unified/latest/images/standard/iot-boot-arm64-rpi3/ work.

daeyeon:
It looks fixed. 👍

PR #1184, opened 2017-09-08
Fix the RPI3 Tizen image and build path

sumin-lim:
Changes
1. The path of build base has been changed. Developer need to change base path on .gbs.conf file.
2. Partitioning script (fusing) has been changed. 

IoT.js-DCO-1.0-Signed-off-by: Sumin Lim sumin.lim@samsung.com


sumin-lim:
Related to #1182 and  #1178.

PR #1186, opened 2017-09-08
Revert ADC API style

hs0225:
- tested on artik053, stm32f4dis


hs0225:
Currently, ADC has different API style from other system I/O modules. There are no updates since #1012, I revert to the previous style to identify style with other modules.

yichoi:
LGTM

PR #1187, opened 2017-09-08
Enable test_fs_readdir.js test file for NuttX and TizenRT.

rtakacs:
`readdir` is implemented for these targets: https://github.com/Samsung/libtuv/pull/95


yichoi:
LGTM

PR #1188, opened 2017-09-08
Update Jerry Submodule and fix build

DanielBallaSZTE:
Updating the jerry submodule plus fixing the IoT.js build, since there was an update that removed the `debugger_cleanup` function, which is now done automatically within `jerry_cleanup()`


DanielBallaSZTE:
@hs0225 Sadly, Travis has some issues with OSX builds right now.

PR #1189, opened 2017-09-11
Implement I2C module on TizenRT

hs0225:
- Add I2C feature on TizenRT
 - Fix simple bug
 - pass  on artik053

### DIFF
--- a/src/modules/iotjs_module_i2c.c
+++ b/src/modules/iotjs_module_i2c.c
@@ -21,7 +21,7 @@
 
 #define THIS iotjs_i2c_reqwrap_t* i2c_reqwrap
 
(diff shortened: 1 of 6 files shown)


src/modules/iotjs_module_i2c.c
```diff
@@ -21,7 +21,7 @@
 
 #define THIS iotjs_i2c_reqwrap_t* i2c_reqwrap
 
-IOTJS_DEFINE_NATIVE_HANDLE_INFO_THIS_MODULE(i2c);
+static JNativeInfoType this_module_native_info = {.free_cb = NULL };
```
glistening: I think we don't need explicitly set to NULL since it is static variable.

yichoi:
LGTM

PR #1190, opened 2017-09-11
Fix test_net_status_codes for TizenRT

daeyeon:
a) In order to run this test on platforms like TizenRT on which EMFILE error may occur, the logic to run tests is fixed to do each case one by one.
b) In the previous code, assuming a certain network delay occurs, the server could be closed before tests were not finished.
c) The status.code, 100, needs to be parted from this test. As of now, I have no idea to quickly make it.


yichoi:
LGTM

PR #1196, opened 2017-09-18
Use the Python based testrunner to measure JavaScript coverage

rtakacs:


rtakacs:
When using a Python based testrunner, every test execution has its own JS context. Since each JS context has its own `__coverity__` variable, this patch applies a `__coverity__` saving code to all the test files. The istanbuljs/nyc program can aggregate the result files.
<br/>

Notes:
  * nvm 6.11.0 doesn't work for me, but the 4.0 version seems ok
  * Should we keep the `check_architecture` function? I use `Ubuntu 16.04 64bit` OS and I can measure the coverage.

tools/testrunner.py
```diff
@@ -28,6 +28,47 @@
 from common_py.system.platform import Platform
 
 
+# Defines the folder that will contain the coverity info.
```
daeyeon: Using `coverity` here is confusing to me. Why don't you use `coverage`?

tools/testrunner.py
```diff
@@ -28,6 +28,47 @@
...
+
+# Append coverity source to the approptiate test.
+def append_coverity_code(testfile):
+    with open(testfile, 'a+') as file_p:
+        file_p.write(JS_COVERAGE_CODE.format(
+            folder=JS_COVERAGE_FOLDER, file=fs.basename(testfile)))
+
+
+# Remove coverity source from the approptiate test.
+def remove_coverity_code(testfile):
```
daeyeon: ditto

tools/testrunner.py
```diff
@@ -28,6 +28,47 @@
...
+
+  var filename = '{folder}/{file}';
+  fs.writeFileSync(filename, Buffer(JSON.stringify(__coverage__)));
+}})
+"""
+)
+
+
+# Append coverity source to the approptiate test.
+def append_coverity_code(testfile):
```
daeyeon: ditto

daeyeon:
@rtakacs I think we can remove `check_architecture`.

tools/testrunner.py
rtakacs: Thanks your notice. Fixed.

rtakacs:
@daeyeon I updated the patch. Need to mention that there are some tests (in the run_fail folder) where the coverage saving code is not executed for some reason:

* test_assert_equal.js
* test_assert_fail.js
* test_assert_notequal.js
* test_events_emit_error.js
* test_iotjs_runtime_error.js
* test_iotjs_syntax_error.js
* test_module_require_invalid_file.js
* test_process_exitcode_arg.js
* test_process_exitcode_var.js
* test_process_explicit_exit.js
* test_process_implicit_exit.js

I don't know the reason, but as I checked all `process.on('exit', function() { ...` should be executed, even if an uncaught exception happens.

robertsipka:
As I remember @bzsolt worked on this kind of problem. Does the `process.exit` function called properly in case of uncaught exception? @bzsolt: Do you know why these test cases can not reach the `process.exit`?

daeyeon:
@rtakacs Can we consider it as an issue of `testrunner.py`, which can be fixed in another commit? BTW, Travis failure needs to be investigated.

tools/measure_coverage.sh
```diff
@@ -145,8 +130,7 @@ mv src/cover_js src/js
 tools/build.py --jerry-cmake-param="-DFEATURE_SYSTEM_ALLOCATOR=ON" \
     --target-arch=x86 --compile-flag="-coverage" --no-snapshot --no-check-test
 
-# Run tests
-build/i686-linux/debug/bin/iotjs tools/check_test.js -- output-coverage=yes
+python tools/testrunner.py ${PWD}/build/i686-linux/debug/bin/iotjs --quiet
```
daeyeon: In order to get coverage information for peripheral I/O module such as gpio, we need to modify the above two lines manually. Do you have an idea on how to handle such a bit annoying stuff more easily?

rtakacs:
@daeyeon I don't think that the issue is from `testrunner.py` because it just executes the tests with `iotjs`. But we have been started to investigate that issue. (Maybe the solution can be a follow up patch.)

tools/measure_coverage.sh
rtakacs: What if we add a parameter to `measure_coverage.sh`?

e.g.
```bash
testrunner=$1

if [ $testrunner = 'jsdriver' ];
then
  build/i686-linux/debug/bin/iotjs tools/check_test.js -- output-coverage=yes
elif [ $testrunner = 'pydriver' ];
then
  python tools/testrunner.py ${PWD}/build/i686-linux/debug/bin/iotjs --quiet
else
  echo "wrong testrunner argument"
  exit 1
fi
```

bzsolt:
We found the cause @rtakacs will update the PR.

rtakacs:
@daeyeon  The problem is happened by the `testrunner.py`. To measure JavaScript coverage, it appends a `__coverity__` saving code to the end of the test files. Since some `run_fail` tests throw an exception by the `assert` module (e.g. assert.equal(1, 2)), the `process.on('exit', function() {{` is not executed, so our `exit` function is not subscribed to the `exit` event. So a possible solution is to apply the coverage saving code to the beginning of the test files.

rtakacs:
I've updated the patch. I added a `--coverage` flag to the testrunner. If that flag is not passed to the testrunner, the coverage measure code is not applied to the test files.

rtakacs:
@daeyeon I've modified the measure_coverage.sh script to support both JS and Python based testrunners. (Uploaded as an independent commit, but I can merge the two commits.)

daeyeon:
@rtakacs Thank you for the swift response. However, I should have mentioned my query more clearly. The thing was build path. We would need to build iotjs for other board like rpi2, and get it run on the board to get coverage information for peripheral I/O modules like gpio. That means the build path described in the script needs to be modified for the purpose. For example:
```bash
python tools/testrunner.py ${PWD}/build/i686-linux/debug/bin/iotjs --quiet
-> python tools/testrunner.py ${PWD}/build/arm-linux/debug/bin/iotjs --quiet

lcov -t "c_coverage" -o ".c-coverage.info" -c -d build/i686-linux/debug/
-> lcov -t "c_coverage" -o ".c-coverage.info" -c -d build/arm-linux/debug/
```
I think that we can handle this in another patch.

rtakacs:
@daeyeon: Thanks for your explanation. I've updated the second commit of this PR. I did not try it out (just a speculative development). I think we should keep the `--testdriver` option while we support both testrunners (check_test.js, testrunner.py). In this case we can compare their coverage results.

tools/measure_coverage.sh
```diff
@@ -127,11 +165,35 @@ mv src/cover_js src/js
...
+    build_path=${PWD}/build/arm-linux/debug
+else
+    echo "Not supported target-board: $target_board"
+    exit 1
+fi
+
+# Run the appropriate testrunner.
+if [ $test_driver = "jsdriver" ];
+then
+    ${build_path}/bin/iotjs tools/check_test.js -- output-coverage=yes
```
daeyeon: `--` needs to be removed.

bzsolt:
Hm, this PR would have needed some modification and commit squashing before merging.
(https://github.com/Samsung/iotjs/pull/1203 will fix the jsdriver option)

PR #1199, opened 2017-09-18
Fix test_net_connection

daeyeon:
This fix closes the server after the test sequence is done.


yichoi:
LGTM

PR #1202, opened 2017-09-20
Auto release feature

daeyeon:


daeyeon:
I applied this to `Buffer` and `Timer` only. This feature itself seems useful although `iotjs_jval_t` is being discussed.

glistening:
@daeyeon I would choose to use jerryscript-ext/autorelease.h directly after iotjs binding simplification is done.

daeyeon:
@glistening ok. let's use it.

Issue #1197, opened 2017-09-18
measuring coverage on rpi2


daeyeon:
`Error: ERR_REF_COUNT_LIMIT` occurs while check_test.js runs under `measure_coverage.sh` on rpi2. 

In the above observation, I modified the following items in measure_coverage.sh :
- check_architecture is blocked.
- use nvm install 4.0.0

rtakacs:
@daeyeon I also noticed that you wrote (#1196)

daeyeon:
This is fixed with https://github.com/Samsung/iotjs/pull/1205.

PR #1205, opened 2017-09-21
Fix coverage tool to easily run on the board

daeyeon:
This patch makes the below works.

a) Set `pydriver` as default since #1197 issue still occurs with `jsdriver`.
b) Any network connection isn't needed on the board once the appropriate node version is installed.
    `nvm install 4.0` requires network access.
c) The directories related are ignored by git.


yichoi:
LGTM

PR #1206, opened 2017-09-21
Parse testsets json into ordered structure

daeyeon:
This patch runs testsets in the order written in json. We can keep the test sequence we used so far.


yichoi:
LGTM

Issue #1073, opened 2017-07-21
Improving remote test runner to cover new os/boards


glistening:
_Note: We will not cover system I/O automation in this issue so that we can concentrate remote-test-running itself_

We are using [remote test runner](https://github.com/rtakacs/remote-testrunner) to run IoT.js unit tests automatically. It works greatly for NuttX/stm32f4-discovery and Linux/RPI. 

Now, we must support two more platforms (TizenRT/artik053 and TizenIoT/RPI) and the kind of boards may grow. Therefore I would like to improve remote test runner to cover several board/os platforms with minimal  cost.

We are thinking of using serial communication. Because

1) TizenRT does not support telnetd officially yet since it is under development. Serial communication is the only one that is verified and we can use safely. We don't need to enable and debug the features that the platform owner does not officially support.

2) We could use the serial communication for NuttX/stm32f4-discovery, Linux/RPI, etc..

@rtakacs Could you summarize the investigation of using serial communication here?

Any comments welcome.

kolipka:
As a continuation of https://groups.io/g/iotjs-dev/message/21:

A user story:
Now I get test_spi.js failed. I need to know how many tests/asserts was executed for this module, and what tests failed. That will help me to estimate condition of module: 1 of 10 failed is differend that 9/10 failed. Now to know we have to read sources and extract this information. This is pretty basic capability of UT frameworks.

I disagree with "_test-runner has no major weakness_" because:
- way of storing results is flawed and have to be fixed ASAP - this is critical flaw. No multiple sources of data, if script fail and amend empty subset of files, all data will be lost( and part of webpage).
- whole runner will have to be refactored since it was designed to store small number of devices. Because of that OS/device tweaks are scattered in many places. This will be rather heavy since requirements as you point it above can't be meet using current approach
- we are making JS engine / nodejs like runtime. We are not using it, we do not learn from other similar products. How can we learn what is missing?  We choose to use Python instead. 
- remote-testrunner workaround issues we have in our test system in overall. eg.: test skiplists. 


benefits from rewriting it in JS/nodejs as I see (iotjs in the future):
- promoting usage of JS
- learning about IoT.js ecosystem lacks
- the best in the world development ecosystem: nodejs + npm (vs python + pypi)
- https://www.npmjs.com/package/firebase maintained by google vs python-firebase last updated 3 years ago
- Since we target JS developers, we can expect contributions from JS developers


[...]

PR #1207, opened 2017-09-21
Add several options to the nuttx target in precommit.py

robertsipka:
As @rtakacs [mentioned](https://github.com/Samsung/iotjs/issues/1073#issuecomment-331081183) earlier, the `precommit.py` script should be extended with a flash mechanism.
It is a small modifications, and does not contain to many things to significantly influences the readability and the maintenance of the script.
The other hand, by adding the memstat patches and the testsuite to the binary, we finally able to measure the memory consumption of the IoT.js testsuite on the nuttx target.


hs0225:
@robertsipka Some code seem to duplicate with `js-remote-test`. Do you plan to merge them?

robertsipka:
@hs0225 : This is the first patch of the other [approach](https://github.com/Samsung/iotjs/issues/1073#issuecomment-331081183), where we could test our development on devices without downloading the js-remote-test project. Several related code parts are already implemented in the js-remote-test, which I have planned to use. The main goal is the testrunner itself had to be part of the IoT.js project.

hs0225:
@robertsipka Are there too many duplicate codes like `*.patch` in iot.js and jerryscript repository? I think it seems to be the same except the application build.

robertsipka:
@hs0225 : These patches only needed for IoT.js. The reason is that the jerryscript allocator could able to follow the libtuv and IoT.js memory allocations. The jerryscript itself does not have to contain them.

rtakacs:
@robertsipka I think, it would be a nicer solution if all your patch files (`tools/testrunner/patches`) could be in a `target/nuttx/stm32f4dis/patches` folder. The patches are belongs logically to the NuttX target, and not to the testrunner. Especially the `nuttx-7.19.diff`, because that has some bugfixes for the our used NuttX version.

Just a note:  the `target/nuttx/stm32f4dis/` folder is already has a patch, so all stm32f4dis-nuttx patches could be in one place.

hs0225:
@robertsipka OK got it. @rtakacs I agree with you.

robertsipka:
I think we will need these patches (3 of a 4) also in case of `artik05x` boards, which will require a `common` folder. On the other hand, since this patch only contain NuttX specific modifications, I agree with you guys and we should move it under a `common` folder later, if it needed (maybe not to under the `tools/testrunner` dir, maybe somewhere under the `config` dir). I've updated this patch accordingly.

rtakacs:
LGTM (informally)

By the way, could you give an example that shows the usage of your development?

robertsipka:
Adding the memstat patches and the testsuite to the binary for measure the memory consumption of the IoT.js testsuite on the nuttx target:
`./tools/precommit.py --test="nuttx" --buildtype=release --buildoptions=--jerry-memstat --enable-testsuite --flash`

config/nuttx/stm32f4dis/jerry-memstat.diff
```diff
@@ -0,0 +1,55 @@
+diff --git a/jerry-core/api/jerry.c b/jerry-core/api/jerry.c
+index 1032881..88db6a3 100644
+--- a/jerry-core/api/jerry.c
++++ b/jerry-core/api/jerry.c
+@@ -149,7 +149,7 @@ jerry_init (jerry_init_flag_t flags) /**< combination of Jerry flags */
+   }
+
+   /* Zero out all members. */
+-  memset (&JERRY_CONTEXT (JERRY_CONTEXT_FIRST_MEMBER), 0, sizeof (jerry_context_t));
++  // memset (&JERRY_CONTEXT (JERRY_CONTEXT_FIRST_MEMBER), 0, sizeof (jerry_context_t));
```
zherczeg: Why this is commented out?

LaszloLango: It need to prevent the overflow of the counter. When `jerry_init` called, the allocated memory is not zero, because the IoT.js did some allocation for the environment. So we should not zero out memory statistics here, because it will cause an overflow after the `free` function calls.

yichoi:
LGTM

PR #1214, opened 2017-09-25
Skip some filesystem tests on NuttX target.

rtakacs:
There are some filesystem tests that should be skipped by default because they depend on the type of the memory. When Read Only Memory is used, some fs tests cannot run. (Currently the js-remote-tests project also uses ROM for the test execution and it would be nice if the js-remote-test doesn't need to have a [local skiplist](https://github.com/Samsung/js-remote-test/blob/master/API/testrunner/iotjs-skiplist.json) to skip the filesystem tests. )


yichoi:
LGTM

PR #1218, opened 2017-09-26
Fix #1215 Add HTTPS link into API references

### LINKED ISSUE #1215: Missing HTTPS documentation in WIKI
Documentation [docs/api/IoT.js-API-HTTPS.md](https://github.com/Samsung/iotjs/blob/master/docs/api/IoT.js-API-HTTPS.md) is missing from [docs/api/IoT.js-API-reference.md](https://github.com/Samsung/iotjs/blob/master/docs/api/IoT.js-API-reference.md)

robertsipka:


daeyeon:
I'm neutral on this. We're likely not to need this for now since the module is available only on the platform, where we don't fully verify yet. There're already a few documents related though.

yichoi:
Current HTTPS module has Tizen IoT specific implementation. After we will make a more general Implementation then APIs and documents, I would like to land this kinds of patch.

PR #1219, opened 2017-09-26
Use fully-featured config file when the testsuite is enabled on nuttx target

robertsipka:
Using `.config.alloptions` since it contains all the features that IoT.js requires.
Restored `.config.travis` to the original state (modified in #1207).


rtakacs:
LGTM (informally)

PR #1221, opened 2017-09-27
Fix possible errors in codes

daeyeon:
This patches the following possible errors found:
- Accessing properties without null check
- Use global variable without declaring it
- Loading unused modules


yichoi:
LGTM

PR #1224, opened 2017-09-27
Add remote source sending feature to IoT.js

DanielBallaSZTE:
This patch makes it possible to send over client source to IoT.js through the debugger clients.
Note:
 - Currently only one source file is handled. In the future, there will be a follow-up patch, allowing the use of multiple source files through require. This requires them to be ordered properly, since importing a module which has not been compiled yet can cause complications.


src/modules/iotjs_module_process.c
```diff
@@ -290,7 +328,17 @@ iotjs_jval_t InitProcess() {
...
 
-  SetProcessArgv(process);
+  bool wait_source =
+      iotjs_environment_config(iotjs_environment_get())->debugger_wait_source;
+
+  if (!wait_source) {
+    SetProcessArgv(process);
+  }
+
+  iotjs_jval_t* wait_source_val = iotjs_jval_get_boolean(wait_source);
```
LaszloLango: iotjs_jval_t wait_source_val = *iotjs_jval_get_boolean(wait_source);

DanielBallaSZTE:
@LaszloLango Fixed it, thanks for the review!

glistening:
@DanielBallaSZTE Does debugger feature is enabled by default for both release and debug build?

DanielBallaSZTE:
@glistening You can turn it on with `--jerry-debugger` in both release and debug builds. However, by default it is disabled.

glistening:
@DanielBallaSZTE Okay. Debugger specific function and field are not enclosed with appropriate flags. Is it trivial for both size and performance?

DanielBallaSZTE:
@glistening I see your point, I'll add a check whether debugger is turned on, so the user won't be able to use other debugger flags without having the debugger turned on first.

Performance-wise there shouldn't be any changes at all, the only thing that changed is instead of processing the file argument from `argv` it is being processed after the debugger-client sent it over, but only when `debugger-wait-source` flag is turned on.

Considering size, memory consumption peaks for a moment, when the code is being passed to IoT.js itself, for a moment the code is stored twice in the memory, than it turns back to normal. Code size might be somewhat affected, if you wish, I could make some measurements for you.

src/modules/iotjs_module_process.c
```diff
@@ -290,7 +328,17 @@ iotjs_jval_t InitProcess() {
...
-  SetProcessArgv(process);
+  bool wait_source =
+      iotjs_environment_config(iotjs_environment_get())->debugger_wait_source;
+
+  if (!wait_source) {
+    SetProcessArgv(process);
+  }
+
+  iotjs_jval_t wait_source_val = *iotjs_jval_get_boolean(wait_source);
+  iotjs_jval_set_property_jval(&process,
```
LaszloLango: Note: This must be updated if #1228 will land first. #1228 will change the prototype of `iotjs_jval_set_property_jval`.

DanielBallaSZTE: Thanks, I'll definitely update if your patch lands.

yichoi:
@DanielBallaSZTE rebase please

glistening:
@DanielBallaSZTE Oh, I mentioned the size and performance for whole debugger feature, not this PR only. For example, `Config` structure contains several members for debugger, it could be saved if the member definition is enclosed with debugger build macro. If binary increase for debugger is trivial, it is okay, and LGTM.

DanielBallaSZTE:
@glistening Hmm..  In a different patch, if this one lands, I can rethink the structure how debugger features are being handled, and only allocate memory for them, if the debugger itself is turned on.

yichoi:
LGTM

PR #1253, opened 2017-10-09
Removed 'iotjs_jval_destroy'

LaszloLango:

### DIFF
--- a/src/iotjs.c
+++ b/src/iotjs.c
@@ -110,7 +110,7 @@ static bool iotjs_run(iotjs_environment_t* env) {
     iotjs_uncaught_exception(&jmain);
   }
 
-  iotjs_jval_destroy(&jmain);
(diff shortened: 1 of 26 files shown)


yichoi:
BTW, It is true that IoT.js is dependent with JerryScript however I am not sure that using ```jerry_release_value``` directly is good or not in the end. I am somewhat inclined to use ```iotjs_jval_destroy``` as a alias or definition of ```jerry_release_value```. Any challenges are welcomed :)

zherczeg:
My thinking is using jerry functions directly simplify iotjs, so it is easier to understand for everybody. Perhaps the question is whether we want to add extra functionality for this function in the future. If this is unlikely, I would remove this one-liner.

yichoi:
either way fine with me :)

yichoi:
LGTM

PR #1255, opened 2017-10-10
Avoid source string duplication during require

galpeter:
When loading an external js file via require
the source code was duplicated internally to wrap
file contents into a function.

Using the jerry_parse_function method it is possible
to directly create the function from the source
without wrapping the source into a function.


galpeter:
@zherczeg updated

PR #1259, opened 2017-10-11
Change the way debugger options are handled

DanielBallaSZTE:
Since debugger options are turned on after compiling the program, there's no proper way to handle them memory-wise.
However, creating a seperate structure for them makes it possible not to allocate memory at the beginning, if debugger is not turned on.
Moreover, later on there might be more debugger specific options (ie. context reset), therefore we can save more memory.

### DIFF
--- a/src/iotjs_env.h
+++ b/src/iotjs_env.h
@@ -18,13 +18,15 @@
 
 #include "uv.h"
 
+typedef struct {
+  bool wait_source;
+  int port;
+} DebuggerConfig;
 
 typedef struct {
   bool memstat;
   bool show_opcode;
-  bool debugger;
-  bool debugger_wait_source;
-  int debugger_port;
+  DebuggerConfig* debugger;
 } Config;
 
 typedef enum {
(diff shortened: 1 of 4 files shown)


src/iotjs_env.h
```diff
@@ -18,13 +18,15 @@
 
 #include "uv.h"
 
+typedef struct {
+  bool debugger_wait_source;
+  int debugger_port;
```
daeyeon: It looks better to remove `debugger_` prefix here.

LaszloLango: I agree. No need to prefix variables inside a struct.

DanielBallaSZTE: Yeah, that's actually right, I didn't change them after moving them to the structure.

DanielBallaSZTE:
@daeyeon Thanks for the review, I changed the PR according to it.

daeyeon:
@DanielBallaSZTE I confirmed it. I have one more query just found. We need to free the buffer of `DebuggerConfig` somewhere, but I can't find where it is.

DanielBallaSZTE:
@daeyeon Yeah, I accidentally forgot to free it, thanks for that!

yichoi:
LGTM

PR #1261, opened 2017-10-12
Removed 'iotjs_jval_t' pointer from 'iotjs_jval_get_undefined'

LaszloLango:
, 'iotjs_jval_get_null' and 'iotjs_jval_get_boolean'.

### DIFF
--- a/src/iotjs_binding.c
+++ b/src/iotjs_binding.c
@@ -20,12 +20,7 @@
 #include <string.h>
 
 
-static iotjs_jval_t jundefined;
-static iotjs_jval_t jnull;
-static iotjs_jval_t jtrue;
-static iotjs_jval_t jfalse;
 static iotjs_jval_t jglobal;
-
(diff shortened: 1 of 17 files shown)


src/iotjs_binding.c
```diff
@@ -129,18 +129,18 @@ static iotjs_jval_t iotjs_jval_create_raw(jerry_value_t val) {
 }
 
 
-iotjs_jval_t* iotjs_jval_get_undefined() {
-  return &jundefined;
+iotjs_jval_t iotjs_jval_get_undefined() {
```
zherczeg: Cant we remove this function? Or is it a follow-up patch? I think the change will be something similar, so I would just remove it here.

LaszloLango: Yes, we can. I'll update the PR, if no one has objection.

LaszloLango:
@zherczeg I've updated the PR. Please check.

yichoi:
LGTM

PR #1273, opened 2017-10-19
Add filtering disabled modules

daeyeon:
This patch refers to target-board along with target-os to build available modules for each travis task.

### DIFF
--- a/build.module
+++ b/build.module
@@ -13,7 +13,8 @@
     },
     "disabled": {
       "board": {
-        "rp2": ["adc"],
+        "x86-linux": ["adc", "ble", "dgram", "gpio", "i2c", "pwm", "spi", "uart"],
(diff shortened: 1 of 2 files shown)


build.module
```diff
@@ -13,7 +13,8 @@
     },
     "disabled": {
       "board": {
-        "rp2": ["adc"],
+        "host-linux": ["adc", "ble", "dgram", "gpio", "i2c", "pwm", "spi", "uart"],
```
glistening: This section is `board`. I think you mean x86-linux by `host-linux`. If so, what about using `x86-linux`?

daeyeon: Thanks for the review. It looks better so I changed it. I used the term because it's been commonly used through our build and travis configurations.

yichoi:
LGTM

PR #1277, opened 2017-10-20
Add a build step for mbedtls

daeyeon:
This commit builds and links mbedtls when there is no real target device.


glistening:
@daeyeon There was a similar PR(#630) about 8 months ago. It was closed without conclusion. Please see [comment](https://github.com/Samsung/iotjs/pull/630#issuecomment-280549825). 

Also, `mbedtls` seems to be downloaded for every platform, and is built when there is no real target device. It is not clear for me. What is the first target you are going to use `mbedtls`?

daeyeon:
@glistening I understand your concerns. I wasn't fully aware of such a discussion. Now I'd also like to know if `https` deserve to be a common module. Plus, it should be clarified which library to use for each platform.

yichoi:
Don't worry about my comments any more

cmake/mbedtls.cmake
```diff
@@ -0,0 +1,71 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+cmake_minimum_required(VERSION 2.8)
+
+if(NOT "${TARGET_BOARD}" STREQUAL "None")
```
glistening: It seems that `mbedtls` is built if no target board is specified. What about introducing explicit cmake option `mbedtls`like `BUILD_MBEDTLS` (default:off) ? This option would be eliminated if HTTPS module policy and implementation is completed. You can use prefix like `EXPERIMENTAL_BUILD_MBEDTLS` to inform the user the option could be removed later.

daeyeon: @glistening I'm ok to selectively build it with `EXPERIMENTAL_BUILD_MBEDTLS`. However, cloning it as submodule is required. Does it work with you?

glistening: @daeyeon I hope to download `mbedtls` module selectively if possible. However, I cannot know and find the way to handle submodule conditionally. If no one has objection, I am ok.

daeyeon: I've updated this patch as suggested.

Issue #1276, opened 2017-10-20
Rework the module loading system (proposal)


LaszloLango:
First of all there are several problems with the current approach.
 * Does not support external/third-party modules. Although it can read package.json (of NPM) and load JS files, but it is only a limited functionality.
 * Handles the native (implemented in C) modules and JS modules differently
   * `process.binding` and `require`
 * Include and exclude lists are confusing

## Proposal

 - [x] Use only `require` for module loading. Remove `process.binding`
 - [x] Use a new JSON descriptor for modules.
 - [x] Use a new "profile" system to describe the enabled modules
   - [x] `build.modules` can be dropped
 - [x] Hide native (written in C) parts and handle modules as a whole.
 - [x] Configure module dependecies
 - [x] Compile only the enabled modules
 - [x] Support external modules, by loading every new `modules.json` from the given directories (configurable as a CMake parameter)
 - [x] Use `ENABLE_MODULE_[NAME]` CMake variables to override the default settings (written in the `package.json`)

### Suggested JSON format

```json
{
  "modules": {
    "iotjs_core_modules": {
      "require": ["assert", "buffer", "console", "process", "testdriver",
                  "timers"]
    },
    "adc": {
      "platforms": {
        "linux": {
          "native_files": ["modules/linux/iotjs_module_adc-linux.c"]
        },
        "nuttx": {
          "native_files": ["modules/nuttx/iotjs_module_adc-nuttx.c"]
        },
        "tizenrt": {
          "native_files": ["modules/tizenrt/iotjs_module_adc-tizenrt.c"]
        },
        "undefined": {
          "native_files": ["modules/iotjs_module_adc.c"]
        }
      },
      "native_files": ["modules/iotjs_module_adc.c"],
      "init": "InitAdc",
      "js_file": "js/adc.js",
      "require": ["iotjs_core_modules"]
    },
   ...
}
```
 
### Introduce profiles for defining the enabled modules.

<pre>
iotjs
 |-- cmake
 |-- config
 |-- deps
 |-- docs
 |-- include<b>
 |-- profiles
      |-- minimal.profile
      |-- normal.profile
      |-- full.profile</b>
 |-- samples
 |-- src
 |-- test<b>
       |-- profiles
            |-- host-linux.profile
            |-- nuttx.profile
            |-- rpi-linux.profile
            |-- tizen.profile
            |-- tizenrt.profile</b>
 |-- tools
</pre>

[...]

PR #1278, opened 2017-10-20
Rework the module system

LaszloLango:
Related issue: #1276


glistening:
@LaszloLango I got following error.

```
CMake Error at cmake/JSONParser.cmake:217 (cmake_policy):
  Policy "CMP0054" is not known to this version of CMake.
```

`CMP0054` seems to be introduced in CMake version 3.1. (https://cmake.org/cmake/help/v3.1/policy/CMP0054.html). I use CMake version 2.8.x on Ubuntu 14.04 LTS.

```
$ cmake --version
cmake version 2.8.12.2
```

glistening:
@LaszloLango For your information, I've succeed to install CMake 3.2.2 on Ubuntu 14.04 LTS with those commands:

```
$ sudo add-apt-repository ppa:george-edison55/cmake-3.x
$ sudo apt-get update
$ sudo apt-get install cmake
```

But second command ( $ sudo apt-get update ) would not finish. So I had to escape from blocking by Ctrl-C. But I continued to execute 3rd command and I got following:

```
$ cmake --version
cmake version 3.2.2
```

LaszloLango:
@glistening thanks for your comments. Will fix the patch to work with version 2.8.x of CMake. We are using Ubunut 16.04 LTS and the default CMake version is 3.5.1 on that.

LaszloLango:
@glistening it is fixed. I've updated the PR. It worked for me with CMake 2.8.12.2.

You can load an extarnal module with the following command:
`tools/build.py --external-modules=my_module_dir` where `my:module_dir` contains a `modules.json` descriptor. For example you can move an existing module (e.g.: PWM) to a new directory and load it as an external module.

glistening:
@LaszloLango I checked the updated PR is built well on Ubuntu 14.04 and cmake 2.8.12.2. Also, I checked it works well with existing module option `--iotjs-include-module`. But I did not tested `--external_modules` since I don't understand fully how to use it. Could you explain more in detail with example? What path should I put for my_module_dir? Does it accept relative path from iotjs root or absolute path? What content should I put for `modules.json`? I think we can make another PR for documentation since this PR seems not break previous options.

LaszloLango:
@glistening, I've just updated the proposal (#1276), but the profile part is not finished yet. That's the final step. I hope we can finish it today, so you can try it. I'll let you know when it is ready for testing.

bzsolt:
@glistening we've finished the profile part and the description of the new module system is also added, so this PR is ready for testing.

LaszloLango:
I forgot to add a documentation file yesterday (`docs/devs/Writing-New-Module.md`). I've updated the PR.

glistening:
@LaszloLango I'd like to merge this PR after we reached conclusion about https://github.com/Samsung/iotjs/issues/1276#issuecomment-339395523 since it is a kind of (build) API.

yichoi:
@LaszloLango please rebase it to the master

daeyeon:
While building the exact module described in `Writing-New-Module.md`, the following build errors occur. Please check if my observation is same as yours.

- cmake version 2.8.12.2
> I wrote mymodule.js and modules.json according to Writing-New-Module.md#writing-javascript-module.
> Then, do ./tools/build.py --external-modules=./my-module --cmake-param=-DENABLE_MODULE_MYMODULE=ON

- error:
CMake Error at cmake/iotjs.cmake:71 (string):
string no output variable specified

---

In case of writing-native-module, 
- iotjs/my-module/my_module.c:1:1: error: unknown type name ‘iotjs_jval_t’

I'd like to know how building a module works when the 3rd party module is located outside `src` folder.

LaszloLango:
@daeyeon I can confirm, there is a bug in the resolving of the require field. The example works without it, because buffer and console modules are part of the core modules. I don't know yet what we messed up with that part, because it worked fine before. I'll check it tomorrow morning. At the native module I forgot to add the include at the beginning of the file, so just add `#include "iotjs_def.h"` to `my_module.c` and the build will find the missing symbol. I'll update the docs tomorrow too.

> I'd like to know how building a module works when the 3rd party module is located outside `src` folder.

With the `--external-modules` (in `build.py`) or `-DEXTERNAL_MODULES` (in `CMake`) you add a list of directories and the build will handle it as it would be part of the `src` dir. In the current approach the external modules directory could contain multiple modules, but only one `modules.json` in its root. For every module an uppercase `ENABLE_MODULE_[NAME]` CMake variable will be generated, where `NAME` is the name of the module in the `modules.json`. So you can enable or disable the default values by setting `ENABLE_MODULE_[NAME]` to `ON` or `OFF`. The purpose of the "`profile`s" is to describe the default settings of the build. You must use `--cmake-param=-DENABLE_MODULE_MYMODULE=ON` oprtion in case of external module, because the `default.profile` enables only the basic and core modules. The other way is to use your own profile file and add it the `build.py` like
`./tools/build.py --profile=my.profile`

I hope I could answer your questions and made things clearer.

LaszloLango:
@daeyeon I've updated the PR. The dependency resolving must work from now. I tested the "mymodule" examples, both js and native examples worked for me. Please check.

daeyeon:
@LaszloLango a) In my observation, `js_file` should be same as `module_name`. Can you check if it's intended?

```
{
    <"modules">: {
    <module_name>: {
      ["js_file"]: <js_entry>
    }
  }
}
```

b) 
> an uppercase ENABLE_MODULE_[NAME] CMake variable will be generated, where NAME is the name of the module in the modules.json

Only with `Writing-New-Module.md`, it's a bit unclear how `ENABLE_MODULE_[NAME]` will be defined.

- Please add some explanations about how ENABLE_MODULE_[NAME] is generated from `modules.json` into `Writing-New-Module.md`.
  - What will be ENABLE_MODULE_[NAME] in case that characters such as hyphen, dash, or underscore and so on, in `<module_name>`.

zherczeg:
I think the point of saparating the name of the js file and the module name is platform support. So if you have `adc_linux.js` and `adc_tizenrt.js` as the source files you can still call these `adc`. This is usually better if you have 5 platforms 4 uses the same file (eg `adc_generic.js`) and the fifth has its own variant. You could put these files into separate folders, but you would have a lot of duplications in that case. That is bad for maintenance.

LaszloLango:
@daeyeon 
a) Your observation is correct, but it is not intended. It is a bug. I'll fix it.
b) I'll add more details to the docs.

LaszloLango:
@daeyeon a) and b) are both fixed.

> What will be ENABLE_MODULE_[NAME] in case that characters such as hyphen, dash, underscore and so on in <module_name>?

Currently there is no character replacing, so the `ENABLE_MODULE_[NAME]` will preserve the special chars in the module name, but this could be changed if you want. :)

docs/devs/Writing-New-Module.md
```diff
@@ -172,4 +269,46 @@ And for asynchronous callbacks, after `libtuv` calls your native function, if yo
...
+
+### Using native module in JavaScript module
+
+Logging to console needs native functionality, so `console` JavaScript module in `src/js/console.js` passes its arguments into native handler like:
+
+```javascript
+Console.prototype.log = native.stdout(util.format.apply(this, arguments) + '\n');
+```
+
+Where `native` is the JS object returned by the native `InitConsole` function in `iotjs_module_console.c`.
```
hs0225: Is it better to specify a native variable name when using a native module in the JS module? I am confused. For example, in `gpio.js`, the native binding variable is `gpio`(not native or native.gpio), but `dns.js` is `native`. 
And how do I use more than two native module in JS module?

tools/travis_script.py
```diff
@@ -100,6 +82,4 @@ def get_include_module_option(target_os, target_board):
             if buildtype == 'release':
                 set_release_config_tizenrt()
             exec_docker(DOCKER_TIZENRT_OS_PATH,
-                        ['make', 'IOTJS_ROOT_DIR=../../iotjs',
-                        'IOTJS_BUILD_OPTION=' +
-                        get_include_module_option(target_os, 'artik05x')])
+                        ['make', 'IOTJS_ROOT_DIR=../../iotjs'])
```
hs0225: To test all module, please add `'IOTJS_BUILD_OPTION=--profile=test/profiles/tizenrt.profile'`

daeyeon:
The proposal says :

> Use only require for module loading. Remove process.binding

However, I found the following different cases also.

a) In `net.js`, `var tcp = require('tcp');` It's fine.
b) In `console.js`, `native` exists as its native part. It's documented but a bit confusing.
c) In `gpio.js`,  no native module loading, but `gpio` looks available as its native part.

Can you clarify what b) and c) are?

src/modules.json
```diff
@@ -0,0 +1,344 @@
...
+    },
+    "ble": {
+      "js_file": "js/ble.js",
+      "require": ["blehcisocket", "ble_characteristic", "ble_descriptor",
+                  "ble_hci_socket_acl_stream", "ble_hci_socket_bindings",
+                  "ble_hci_socket_crypto", "ble_hci_socket_gap",
+                  "ble_hci_socket_gatt", "ble_hci_socket_hci",
+                  "ble_hci_socket_hci_status", "ble_hci_socket",
+                  "ble_hci_socket_mgmt", "ble_hci_socket_smp",
+                  "ble_primary_service", "ble_uuid_util"]
```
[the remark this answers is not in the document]
LaszloLango: @daeyeon good idea. @zherczeg had a similar suggestion. We wanted to do it in a followup work, beacause it is not trivial. The idea is to use `"require": [ "auto" ]` or `"require": [ "*" ]` for automatic require detection. But missing require field could trigger it too. That is also a good idea. Should we do it in this PR?

docs/devs/Writing-New-Module.md
LaszloLango: We changed the function wrapper to use an extra `native` parameter. If `native` is not undefined than the module has parts written in C. The `native` variable contains the return value of the `init` function.

> And how do I use more than two native module in JS module?

You can use only the `require()` and use what a module exports. This is intentional. Our idea was to make it possible to hide implementation details from module users. If you write a module with both js and native files, then it is the javascript file's responsibility to export the public objects. The native files of that module is only available for that module. For such use cases where a native part must be shared, you can create a module that only contains native files. If there is no JS file presented then the return value of the init function will be exported.

```
{
  "modules": {
    "common_native_module": {
      "native_files": ["common.c"],
      "init": "InitCommonThings"
    }
  }
}
```

my_module1.js:
```js
var common = require('common_native_module');
if (native) {
 // private native part
}
...
```

my_module2.js:
```js
var common = require('common_native_module');
if (native) {
 // private native part
}
...
```

LaszloLango:
a) It is definitely a bug. We haven't tested those tests which needs a test environment.
b) In short: `native` represents the private part of a module written in C.

src/modules.json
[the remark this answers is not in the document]
daeyeon: @LaszloLango Great to hear that you already discussed on this. I also thought we can do this in another PR.

tools/iotjs_build_info.js
```diff
@@ -1,31 +0,0 @@
-/* Copyright 2017-present Samsung Electronics Co., Ltd. and other contributors
```
rtakacs: This file should not be removed because [testrunner.py:156](https://github.com/Samsung/iotjs/blob/master/tools/testrunner.py#L156) uses that to determine the built-in modules.

LaszloLango:
@daeyeon, @hs0225 I've updated the PR.

src/js/i2c.js
```diff
@@ -44,7 +44,6 @@
  */
 
 var util = require('util');
-var i2c = process.binding(process.binding.i2c);
```
daeyeon: L:92 `new i2c(...` -> `new native(...`

src/js/dgram.js
```diff
@@ -67,7 +66,7 @@ function Socket(type, listener) {
   this.type = type;
   this.fd = null; // compatibility hack
 
-  // If true - UV_UDP_REUSEADDR flag will be set
+  // If true - UV_udp_REUSEADDR flag will be set
```
daeyeon: `UV_UDP_REUSEADDR`

src/js/i2c.js
LaszloLango: fixed

LaszloLango:
@daeyeon thanks for your review. I've updated the PR.

yichoi:
LGTM

PR #1282, opened 2017-10-24
Sample: GPIO control panel

tosyu:
This sample shows to how to create a http server, serve content and supply a simple REST api to allow change/monitor pin state


haesik:
I have tested it on TizenRT. I think it's not work correctly..
when I connect board using browser(I tried on mobile phone and Safari browser), 
I can see "serving static: index.html" message on console..
On the browser, however,  I could only see "404 not found" message.

tosyu:
@haesik the 404 error should be fixed, TizenRT has a problem with relative paths so I added process.cwd() to fix this

haesik:
I made another PR #1471 to fix an error

PR #1283, opened 2017-10-24
Sample: Light fade

tosyu:
Turns LED light on/off at the press of the button with a fade effect using PWM


yichoi:
LGTM

PR #1284, opened 2017-10-24
Sample: UART JavaScript console

tosyu:
Simple example that evaluates JS code through UART interface, the code runs when CTRL+R is pressed


haesik:
I made another PR #1472 to fix an error

PR #1288, opened 2017-10-25
Remove `--jerry-debugger-port` build option.

robertsipka:
The port number of the debugger-server can be specify via the `iotjs` binary options.


yichoi:
LGTM

PR #1289, opened 2017-10-26
Remove 'DJHANDLER_CHECK_ARGS(0)' calls.

LaszloLango:
This check does not make any sense, because it will always be true,
due to using unsigned integer types.

### DIFF
--- a/src/iotjs_binding.h
+++ b/src/iotjs_binding.h
@@ -252,13 +252,8 @@ iotjs_jval_t iotjs_jval_create_function_with_dispatch(
   JHANDLER_CHECK_ARGS_4(type0, type1, type2, type3);             \
   JHANDLER_CHECK_ARG(4, type4);
 
-// Workaround for GCC type-limits warning
-static inline bool ge(uint16_t a, uint16_t b) {
-  return a >= b;
-}
-
-#define JHANDLER_CHECK_ARGS(argc, ...)                               \
-  JHANDLER_CHECK(ge(iotjs_jhandler_get_arg_length(jhandler), argc)); \
+#define JHANDLER_CHECK_ARGS(argc, ...)                             \
+  JHANDLER_CHECK(iotjs_jhandler_get_arg_length(jhandler) >= argc); \
   JHANDLER_CHECK_ARGS_##argc(__VA_ARGS__)
 
 #define JHANDLER_CHECK_THIS(type) \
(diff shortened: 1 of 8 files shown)


LaszloLango:
The gain of the patch is only a few byte reduction in the `.text` section, but the code is much cleaner.

glistening:
@LaszloLango The previous code explains the code is for GCC type-limits warning

```
-// Workaround for GCC type-limits warning
-static inline bool ge(uint16_t a, uint16_t b) {
-  return a >= b;
-}
```

I don't know what gcc version make problem, but I would like to keep this code.

LaszloLango:
@glistening, the problem is that an unsigned value never will be less than zero. With that workaround we hide the problem because the compiler will use implicit type conversion. The GCC does not warn now, but the `DJHANDLER_CHECK_ARGS(0)` never could be false.

LaszloLango:
@glistening, @daeyeon 
A bit longer explanation. The GCC warning is the following without the `ge` function:
`comparison is always true due to limited range of data type`

Full trace:
```
iotjs/src/iotjs_binding.h:261:58: error: comparison is always true due to limited range of data type [-Werror=type-limits]
   JHANDLER_CHECK(iotjs_jhandler_get_arg_length(jhandler) >= argc); \
                                                          ^
iotjs/src/iotjs_binding.h:217:9: note: in definition of macro ‘JHANDLER_CHECK’
   if (!(predicate)) {                         \
         ^
iotjs/src/iotjs_binding.h:289:41: note: in expansion of macro ‘JHANDLER_CHECK_ARGS’
 #define DJHANDLER_CHECK_ARGS(argc, ...) JHANDLER_CHECK_ARGS(argc, __VA_ARGS__)
                                         ^
iotjs/src/modules/iotjs_module_process.c:170:3: note: in expansion of macro ‘DJHANDLER_CHECK_ARGS’
   DJHANDLER_CHECK_ARGS(0);
```
This happens at every occurrence of `DJHANDLER_CHECK_ARGS(0)`. The expansion of `DJHANDLER_CHECK_ARGS` ends in a condition that will always be true: `unsigned int >= 0`.

Using the `ge` function is only hide this problem and makes the IoT.js slower, because of the superfluous checks. I hope this explanation helped you to understand the PR better. Do not hesitate  to ask if something is still not clear for you. :)

glistening:
@LaszloLango I checked that this PR does not show warning on Ubuntu 14.04 with gcc 4.9.4, and Ubuntu 16.04. I was going to check Ubuntu 14.04 with gcc 4.8.4 (default gcc), but meanwhile you explained in detail. I am ok for this PR.

PR #1290, opened 2017-10-26
Simplify the binding layer.

LaszloLango:
* Removed 'iotjs_jval_is_*' functions
 * Removed 'iotjs_error_t'
 * Removed 'iotjs_binding_initialize' and 'iotjs_binding_finalize'

### DIFF
--- a/src/iotjs.c
+++ b/src/iotjs.c
@@ -114,11 +114,8 @@ static bool iotjs_run(iotjs_environment_t* env) {
 
 
 static int iotjs_start(iotjs_environment_t* env) {
-  // Initialize commonly used jerry values.
-  iotjs_binding_initialize();
-
   // Bind environment to global object.
-  const iotjs_jval_t global = iotjs_jval_get_global_object();
+  const iotjs_jval_t global = jerry_get_global_object();
   jerry_set_object_native_pointer(global, env, NULL);
 
   // Initialize builtin modules.
@@ -127,6 +124,9 @@ static int iotjs_start(iotjs_environment_t* env) {
   // Initialize builtin process module.
   const iotjs_jval_t process = iotjs_init_process_module();
   iotjs_jval_set_property_jval(global, "process", process);
+
+  // Release the global object
+  jerry_release_value(global);
 
   // Set running state.
   iotjs_environment_go_state_running_main(env);
@@ -215,9 +215,6 @@ int iotjs_entry(int argc, char** argv) {
 
   int res = uv_loop_close(iotjs_environment_loop(env));
   IOTJS_ASSERT(res == 0);
-
-  // Release commonly used jerry values.
-  iotjs_binding_finalize();
 
   // Release JerryScript engine.
   iotjs_jerry_release(env);
--- a/src/iotjs_binding.c
+++ b/src/iotjs_binding.c
@@ -21,9 +21,11 @@
 #include <string.h>
 
 
-static iotjs_jval_t jglobal;
-static iotjs_jargs_t jargs_empty;
-
+static iotjs_jargs_t jargs_empty = {.unsafe = { 0, 0, NULL },
+#ifndef NDEBUG
+                                    .flag_create = IOTJS_VALID_MAGIC_SEQUENCE
+#endif /* !NDEBUG */
+};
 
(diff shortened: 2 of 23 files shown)


src/iotjs_binding.c
```diff
@@ -21,9 +21,11 @@
 #include <string.h>
 
 
-static iotjs_jval_t jglobal;
-static iotjs_jargs_t jargs_empty;
-
+static iotjs_jargs_t jargs_empty = {.unsafe = { 0, 0, NULL },
+#ifndef NDEBUG
+                                    .flag_create = IOTJS_VALID_MAGIC_SEQUENCE
```
daeyeon: `.unsafe` and `.flag_create` looks indentation wrong.

src/iotjs.c
```diff
@@ -128,6 +125,9 @@ static int iotjs_start(iotjs_environment_t* env) {
   const iotjs_jval_t process = iotjs_init_process_module();
   iotjs_jval_set_property_jval(global, "process", process);
 
+  // Release the global object
+  jerry_release_value(global);
```
daeyeon: I guessed releasing `global` at the termination is right and safe. Can you kindly clarify my query?

src/iotjs_binding.c
LaszloLango: `tools/check_tidy --autoedit` made this indentation. What would be the correct indentation?

src/iotjs.c
LaszloLango: Global is not used in the code below. It is only needed for registering the "process" object. No need to hold the reference until the IoT.js terminates.

glistening: As of my understanding, reference count for `global` is increased by `jerry_get_global_object` so calling `jerry_release_value` does not make the code unsafe. Succeeding `jerry_release_value` (-1) seems more natural for me as a pair of `jerry_get_global_object` (+1)

LaszloLango: @glistening yes, that's right.

yichoi:
@LaszloLango please rebase it to the master

src/iotjs.c
daeyeon: @LaszloLango Understood. We didn't need to hold `jglobal` at the very beginning. Thanks.

@glistening As wrote, my query wasn't why calling the release function as the pair. I know the reference counting internally works. I wondered what the thoughts of someone who wrote `jglobal`, which is released at the termination, was. It turns out unnecessary.

src/iotjs_binding.c
daeyeon: I see. It's right.

daeyeon:
While building the exact module described in `Writing-New-Module.md`, the following build errors occur. Please check if my observation is same as yours.

- cmake version 2.8.12.2
> write mymodule.js and modules.json according to Writing-New-Module.md#writing-javascript-module
> ./tools/build.py --external-modules=./my-module --cmake-param=-DENABLE_MODULE_MYMODULE=ON

- error:
CMake Error at cmake/iotjs.cmake:71 (string):
string no output variable specified

---

In case of writing-native-module, 
- iotjs/my-module/my_module.c:1:1: error: unknown type name ‘iotjs_jval_t’

I'd like to know how this works when the 3rd party module is located outside `src` folder.

LaszloLango:
@daeyeon, I'll check it tomorrow morning, but you commented on the wrong PR. :)

LaszloLango:
@daeyeon, add `#include "iotjs_def.h"` to my_module.c. I left it out from the example.

daeyeon:
@LaszloLango Right. I selected the wrong RP. Sorry! I will move it.

src/iotjs.c
hs0225: @LaszloLango In `uv_loop_close`, isn't it possible to use `global` in the callback function?

LaszloLango: @hs0225 you can get the gobal object anytime before the jerry_cleanup just like before this patch.

hs0225: @LaszloLango I understand. There is no problem because `jerry_cleanup` is called later.

LaszloLango:
@hs0225 may I land this PR or do you have other comments?

hs0225:
@LaszloLango No more comments. LGTM

PR #1291, opened 2017-10-26
Remove excess value acquiring in iotjs_jhandler_return_jval/iotjs_jha…

bzsolt:
…ndler_throw


glistening:
@bzsolt Now, the caller of `iotjs_jhandler_return_jval` and `iotjs_jhandler_throw` must make sure the value argument will not be used any more. If they want to use the value argument, they must call acquire explicitly. Is it right? If so, I would like to inform other members.

glistening:
I am ok for this change. Also, I checked the occurrence of the changed protocol applied well.

PR #1298, opened 2017-11-08
Stringify object in console module

daeyeon:
This patch stringifies the given object in console. 

For instance, `console.log({ a: '123', b: 123, c: [1, 2, 3]});`

- Current:
`[object Object]`

- This patch:
`{"a":"123","b":123,"c":[1,2,3]}`

### DIFF
--- a/src/js/util.js
+++ b/src/js/util.js
@@ -146,6 +146,10 @@ function formatValue(v) {
     return 'undefined';
   } else if (v === null) {
     return 'null';
+  } else if (Array.isArray(v) || v instanceof Error) {
+    return v.toString();
+  } else if (typeof v === 'object') {
+    return JSON.stringify(v, null, 2);
   } else {
     return v.toString();
   }
--- a/test/run_pass/test_console.js
+++ b/test/run_pass/test_console.js
@@ -24,6 +24,7 @@ console.log([1, 2, 3]);
 console.log(1, 2, 3);
 console.log('a', 1, 'b', 2, 'c', 3);
 console.log("test", null, undefined);
+console.log({ a: '123', b: 123, c: [1, 2, 3]});
 
 console.error("Hello IoT.js!!");
 console.error(1);


glistening:
@daeyeon What if the array size is quite big? (e.g., length =1000) I tested this scenario on FireFox console.

```
var o = [ 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10,1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10,]
undefined
console.debug(o)
Array [ 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 200 more… ]
```

I would like to choose this style. You may use another style. Anyway, I would like to handle big size of array.

daeyeon:
@glistening The style looks fine but not the scope this commit covers. I changed nothing related to Array.

glistening:
@daeyeon Attempting print big size array would make the device hang, I don't want to introduce the maybe hanging operation without safe guard.

daeyeon:
@glistening Again. This commit is intended not to show `[object Object]` when a common object is given. The style and the safe guide should be in another commit after more discussion. E.g) What should be the style? How big is the big you mentioned?

glistening:
@daeyeon I don't think we need to survey (nodejs, chrome, ...). It is not complex question. I think adding several line for length condition would be enough. Let me assume you're right. it requires all the discussions, then we must hold this PR. Before the fix PR is landed, the risk of hanging exists.

daeyeon:
@glistening 
> Because before the fix PR, the risk of hanging will be exposed to users.

Regarding the above example which you tested on the browser, the latest source in the master branch is already exposing the risk of hanging regardless of this patch. Thus, it doesn't make sense that the risk will occur if not fix this PR. Why should a patch for a single purpose include any fix not related?

At least, I can agree not to enumerate a value in case that the type of any property within an object is `Array`.

daeyeon:
@glistening 

>  It is not complex question. Adding several lines for length condition would be enough.

I'm afraid I can't agree that. Can you simply say what the length condition is proper in user's perspective? And, can you easily answer how big size of `Array` makes the system hang? How about its displayed style? Everyone can have a different preference. If you have a clear opinion, please suggest it in another PR.

yichoi:
LGTM for now. If needed , please make another issue then discuss again.

glistening:
@daeyeon 

> Regarding the above example which you tested on the browser, the latest source in the master branch is already exposing the risk of hanging regardless of this patch.

I can't understand the sentence above. The risk of hanging is introduced by this your PR. Before this PR, it just printed `[object Object]`. After this PR, it will create very large string and try to print. Anyway, unfortunately it is already merged.

glistening:
@daeyeon 
> I'm afraid I can't agree that. Can you simply say what the length condition is proper in user's perspective? 

Again, I cannot understand why this is complex problem. Almost all IDE ( Visual Studio, ... ) or browser limits the amount of data for inspecting data. It is a simple problem. There may be various preference for how it will print for limited data. But it can be changed after safe guard for size is added. But it is already merged.

daeyeon:
@glistening You didn't clearly answer how large the length is reasonable yet. That's my complex point. Until it's defined, our strategy will be printing large strings or not depending on user's intention.

daeyeon:
@glistening You didn't clearly answer how large the length is reasonable yet. That's my complex point. Any wrong limit could hurt its usability. Until it's defined, our strategy would be printing large strings or not depending on the user's intention. 

I will stop answering other misunderstandings since that's not much important for us.

daeyeon:
@glistening You didn't clearly answer how large the length is reasonable yet. That's my complex point. Any wrong limit could hurt its usability. Until it's defined, our strategy would be printing large strings or not depending on the user's intention.

I'll stop answering other miscomprehensions to focus on other importance.

daeyeon:
@glistening You didn't clearly answer how large the length is reasonable yet. That's what I asked. You told just several lines of code will be enough not to make the system hang without hurting its usability, but it's not suggested. Until it's defined, our strategy would be printing large strings or not depending on the user's intention.

I'll stop answering other miscomprehensions to focus on other importance.

daeyeon:
@glistening You didn't clearly answer how large the length is reasonable yet. That's what I asked. You seemly told just several lines of code will be enough not to make the system hang without hurting its usability, but it's not suggested. Until it's defined, our strategy would be printing large strings or not depending on the user's intention.

I'll stop answering other miscomprehensions to focus on other importance.

daeyeon:
@glistening You didn't clearly answer how large the length is reasonable yet. That's what I asked. You seemingly told just several lines of code will be enough not to make the system hang without hurting its usability, but it's not suggested. Until it's defined, our strategy would be printing large strings or not depending on the user's intention.

I'll stop answering other miscomprehensions to focus on other importance.

daeyeon:
@glistening You didn't yet answer how large the length is reasonable. That's what I asked. You told that several lines with it won't make the system hang you concerned. However, the codes to prove that wasn't suggested. It can be proposed in another issue. Until then, our strategy would be printing large strings or not depending on the user's intention.

I'll stop answering other miscomprehensions to focus on other importance.

Issue #1059, opened 2017-07-18
Random travis-ci test fails


tosyu:
Hello

I have a problem with TravisCI tests :scream:  Sometime it raindomly fails on asserts spilling out errors in valgrind environment. Sometimes it fails on coverity, sometimes it fails on host-linux tests. The solution is to force restart the tests hoping that the next run will pass :confused: 

Example https://travis-ci.org/Samsung/iotjs/builds/254446216?utm_source=github_status&utm_medium=notification

IMHO this error has nothing to do with the PR :cry:

tadziopazur:
I have seen this behavior by Travis -- seemingly irrelevant tests fail at uv error handling.
Sometimes they fail directly at timeout handling (net tests IIRC).

I think the real cause is the invalid handling of timeouts that are due to underpowered Travis worker instances.

pmarcinkiew:
Today it happen two times already for me. I just change commit message and trigger Travis to get 100% pass. It was test_net_2 two times. In my opinion it is unstable. We should disable it and investigate it separately. It is not efficient to debug some network instability during each pull request.

daeyeon:
No further update for a long time. Some tests seriously flaky on Travis are now skipped. https://github.com/Samsung/iotjs/blob/master/test/testsets.json

PR #1299, opened 2017-11-08
Fix failing tests after the module system rework.

LaszloLango:
Three tests (test_events_uncaught_error.js, test_process_uncaught_order.js,
test_process_uncaught_simple.js) were failing with 'tools/testrunner.py' in
snapshot mode after the module system rework.

### DIFF
--- a/src/iotjs.c
+++ b/src/iotjs.c
@@ -104,6 +104,7 @@ static bool iotjs_run(iotjs_environment_t* env) {
       jerry_exec_snapshot_at((const void*)iotjs_js_modules_s,
                              iotjs_js_modules_l, module_iotjs_idx, false);
   if (jerry_value_has_error_flag(jmain)) {
+    jerry_value_clear_error_flag(&jmain);
     throws = true;
   }
 #endif


glistening:
@LaszloLango Could you explain more for this issue and PR? 

i) These three tests seems ok on x86_64 since travis test was ok when you pushed #1278, and I checked `test_events_uncaught_error.js` is ok on my Ubuntu 14.04. The errors occur on RPI and artik053. What is the difference? RPI also uses linux based OS.

ii) Clearing error flag was not needed before module rework. Why this explicit clearing is required?

LaszloLango:
i) The JS tesrunner overrides the `exit()` function and this is somehow hide the error when we use it instead of `tools/testrunner.py`. The JS testrunner is not reliable (see also: #1059).

**Command**:
`build/x86_64-linux/debug/bin/iotjs test/run_pass/test_events_uncaught_error.js ; echo $?`
**Output**:
`2`

ii) Yes, it was needed before it too. The rework did not change that logic. The binding function did that step, like it still does in no-snapshot builds. See: 
https://github.com/Samsung/iotjs/blob/master/src/iotjs.c#L100 and https://github.com/Samsung/iotjs/blob/master/src/iotjs_binding.c#L393

Anyway, this is a hotfix for those tests. I'd like to revise the error handling of IoT.js in the near future.

yichoi:
LGTM

PR #1304, opened 2017-11-13
Improve built-in native c module generation

galpeter:
The change is split into two things:

1) Simplify the module array generation in the iotjs.cmake.
Replace the module initializer function
with an already initialized module array.

2) Split built-in native C modules into two array.

The `iotjs_modules` array is composed of `iotjs_module_t`
elments where only the `jmodule` member is not initialized
statically. Thus the array can not be placed into the rodata
section (if needed).

By extracting the only dynamic member (`jmodule`) into its own
array it is possible to move the original module array into the
rodata section.


daeyeon:
LGTM, but the two commits need to be combined into single one.

galpeter:
Merged the two commits into one.

PR #919, opened 2017-06-09
Change the way to search custom modules

bzsolt:
- Search modules under `iotjs_modules/` instead of `node_modules/`.
 - Use `IOTJS_PATH` environment variable instead of `NODE_PATH`.

IoT.js-DCO-1.0-Signed-off-by: Zsolt Borbély zsborbely.u-szeged@partner.samsung.com

PR #902, opened 2017-06-06
Update Module documentation

bzsolt:
IoT.js-DCO-1.0-Signed-off-by: Zsolt Borbély zsborbely.u-szeged@partner.samsung.com

PR #1305, opened 2017-11-14
Fix finding a package entry

daeyeon:
This patch contains the followings:

i) fix: If a file, which is named the main property in `id/package.json`, doesn't contain its extension, try loading it with '.js'.
~~ii) change: revert `iotjs_modules` to `node_modules` to leverage node modules which can be used in iotjs.~~

### DIFF
--- a/src/js/module.js
+++ b/src/js/module.js
@@ -49,6 +49,10 @@ if (process.env.IOTJS_PATH) {
   moduledirs.push(process.env.IOTJS_PATH + '/iotjs_modules/');
 }
 
+function tryPath(modulePath, ext) {
+  return iotjs_module_t.tryPath(modulePath) ||
+         iotjs_module_t.tryPath(modulePath + ext);
+}
 
 iotjs_module_t.resolveDirectories = function(id, parent) {
   var dirs = moduledirs;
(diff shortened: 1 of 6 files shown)


glistening:
@daeyeon Are you trying to revert #919? We used the prefix `nodejs`, but we decided to use the prefix `iotjs`. We need to discuss about this rollback.

glistening:
@daeyeon  As I remember, we used the prefix `iotjsjs` because we do not ensure the nodejs compatability. we don't want to make iotjs users confused as we support nodejs modules. @bzsolt Do you remember the background for your PR #919?

daeyeon:
Using `iotjs_modules` is what I suggested actually and we asked helps to USZ. On my second thought, it seems better to reopen using `node_modules`. It can, at least, gives users a way to select any module which can run in iotjs. Our current usability to leverage other modules is poor. Why don't we put some caution about node module compatibility in the document?

glistening:
@daeyeon I vote for keeping `iotjs_modules`. For me, it is strange a little that uses `node_modules` folder for iotjs modules. For example, it will search $HOME/node_modules implicitly for iotjs modules. I don't think it is natural.

bzsolt:
@glistening #919 origins from https://github.com/Samsung/iotjs/pull/902#discussion_r121045943

glistening:
@daeyeon What about introducing environment variable like `IOTJS_SEARCH_NODEJS_MODULES`? When the user explicitly set this environment variable, iotjs will search nodejs_modules additionally after searching iotjs_modules first.

daeyeon:
According to the internal discussion, ii) is separated in low priority and later handled. Thank you.

src/js/module.js
```diff
@@ -49,6 +49,19 @@ if (process.env.IOTJS_PATH) {
...
+  if (filepath = iotjs_module_t.tryPath(modulePath)) {
+    return filepath;
+  }
+
+  if (filepath = iotjs_module_t.tryPath(modulePath + ext)) {
+    return filepath;
+  }
+
+  return filepath;
+}
```
glistening: I think the code above is same to following:

```js
function tryPath(modulePath, ext) {
  return iotjs_module_t.tryPath(modulePath) ||
         iotjs_module_t.tryPath(modulePath + ext)
}
```

daeyeon: @glistening Thanks, it looks better. I've changed it.

PR #1306, opened 2017-11-16
Correct wrong descriptions

daeyeon:


glistening:
@daeyeon I also find that the description for `IOTJS_BUILD_OPTION` is missing in this document. Could you include the description in this PR? Or if you are not pleased to do, I will make another PR.

PR #1307, opened 2017-11-16
Fixed chain calls of Promise objects which contains async calls.

### LINKED ISSUE #1292: Chain calls cannot be implemented
I want to implement exact chain calls in iotjs. Here are the test cases. Just read several files in chained form. The file name can be modified by itself. This code works normally in nodejs. I opened the es2015-subset. So Promise is available. Parallel code is normally called and serial is not.

```javascript
var fs = require('fs')
function readfile(fileName){
  return new Promise(function (resolve, reject){
    fs.readFile(fileName, function(error, data){
      if (error) reject(error);
        console.log(data.toString());      
        resolve(data);      
    });
  });
};
code ERR:
readfile('test.csv').then(function() {
    readfile('NMap.ts').then(function() {
        readfile('NMap.js').then(function() {
             readfile('test.csv').then(function() {
             console.log(Date.now());
           })
        })
    })
})
code OK:
var tasks=[];
tasks.push(readfile('test.csv'));
tasks.push(readfile('NMap.ts'));
tasks.push(readfile('NMap.js'));
tasks.push(readfile('test.csv'));
Promise.all(tasks).then(function () {
      console.log(new Date());
});
```

LaszloLango:
Fixes #1292 

### DIFF
--- a/test/run_pass/test_iotjs_promise_chain_calls.js
+++ b/test/run_pass/test_iotjs_promise_chain_calls.js
@@ -1,0 +1,58 @@
+/* Copyright 2017-present Samsung Electronics Co., Ltd. and other contributors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+var fs = require('fs')
+var assert = require('assert');
+
+function readfile(fileName) {
+  return new Promise(function (resolve, reject) {
+    fs.readFile(fileName, function(error, data) {
+      if (error) {
+        reject(new Error('Cannot open file!'));
+      } else {
+        resolve(data.toString());
+      }
+    });
+  });
+};
+
+function loadfi(files, idx) {
+  var fileName = process.cwd()
+                 + '/resources/promise_chain_calls/'
+                 + files[idx];
+
+  readfile(fileName).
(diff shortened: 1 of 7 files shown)


test/run_pass/test_iotjs_promise_chain_calls.js
```diff
@@ -0,0 +1,58 @@
...
+                 + files[idx];
+
+  readfile(fileName).
+    then(function(value) {
+      assert.equal(value, 'Content of ' + files[idx] + '\n');
+      idx++;
+      loadfi(files, idx);
+    }).catch(function (e) {
+     // Note: assert cannot be used here, because exception cannot be thrown
+     //       from Promise.prototype.catch
```
galpeter: The indentation is off by one in the comments. Is that intentional?

test/run_pass/test_iotjs_promise_chain_calls.js
```diff
@@ -0,0 +1,58 @@
...
+      if (e.message !== "Cannot open file!") {
+        console.log('Error message does not match. Expected '
+                    + '"Cannot open file!", but got "'
+                    + e.message + '".');
+        process.emitExit(1);
+      } else if (idx !== 4) {
+        console.log('Did not read every file. The expected number of files '
+                    + 'read is "4", but got "' + idx + '".');
+        process.emitExit(1);
+      }
```
galpeter: Two questions:
* Should we also check any other reasons for failure?
* Can we hoist the `emitExit` call out of the if-s to the function end? Thus if any error occurs we'll exit.

test/run_pass/test_iotjs_promise_chain_calls.js
LaszloLango: No, it is accidental.

test/run_pass/test_iotjs_promise_chain_calls.js
LaszloLango: Yes, we should check and the test does it. If any error happens different than the `reject("Cannot  open file.")` it will exit with 1. This will happen, when the test tries to open "`undefined`". I wanted to keep the original behavior of the test.

yichoi:
LGTM

PR #1317, opened 2017-11-21
Remove 'iotjs_jval_set_prototype' to reduce binary size.

LaszloLango:

### DIFF
--- a/src/modules/iotjs_module_fs.c
+++ b/src/modules/iotjs_module_fs.c
@@ -313,7 +313,7 @@ iotjs_jval_t MakeStatObject(uv_stat_t* statbuf) {
   IOTJS_ASSERT(jerry_value_is_object(stat_prototype));
 
   iotjs_jval_t jstat = iotjs_jval_create_object();
-  iotjs_jval_set_prototype(jstat, stat_prototype);
+  jerry_set_prototype(jstat, stat_prototype);
 
   jerry_release_value(stat_prototype);
 
--- a/src/iotjs_binding.c
+++ b/src/iotjs_binding.c
@@ -173,15 +173,6 @@ iotjs_jval_t iotjs_jval_as_function(iotjs_jval_t jval) {
 }
 
 
-bool iotjs_jval_set_prototype(const iotjs_jval_t jobj, iotjs_jval_t jproto) {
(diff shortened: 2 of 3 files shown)


LaszloLango:
The `.text` section is reduced by 156 bytes.

src/modules/iotjs_module_fs.c
```diff
@@ -41,7 +41,61 @@ static void iotjs_fs_reqwrap_destroy(iotjs_fs_reqwrap_t* fs_reqwrap) {
...
+JS_FUNCTION(StatsIsFile) {
+  DJS_CHECK_THIS(object);
+  iotjs_jval_t stats = JS_GET_THIS(object);
+  return StatsIsTypeOf(stats, S_IFREG);
+}
+
+
+static iotjs_jval_t MakeStatObject(uv_stat_t* statbuf) {
+  iotjs_jval_t jstat = iotjs_jval_create_object();
+  iotjs_jval_t stat_prototype = iotjs_jval_create_object();
```
galpeter: With this approach we always recreate the Stat object prototype and each time it is a different one. So we don't get anything with using prototypes as two stat objects prototype will be different thus there isn't any sense to have a prototype object at all.

The old way was to ensure that every Stat object will have the same prototype.

LaszloLango:
@galpeter thanks for your review. I've update the PR.

glistening:
@LaszloLango I would like to keep `iotjs_jval_set_prototype`.

i) The binary size gain came from the fact that the only usage of `iotjs_jval_set_prototype` does not handle error case. If we add the checking error (it should be there), perhaps there are no binary gain.

ii) Although it may be transient, for setting method, I have to use `iotjs_jval_set_method`. For property, I have to use `jerry_set_prototype`. If we will keep `iotjs_jval_set_method`, `iotjs_jval_set_property_jval`, ....  ( I am voting for keep since they are handy to use.), I would like to keep `iotjs_jval_set_prototype`.

LaszloLango:
@glistening, OK. I don't insist on this PR. Unfortunately my initial idea did not work well, so I had to reduce the patch to this version. The PR does not have a significant gain, so let's close it.

PR #1318, opened 2017-11-21
Specify the profile in measure_coverage.sh

bzsolt:
Additionally fix identation and abort the script when the build fails.


yichoi:
LGTM

Issue #1245, opened 2017-10-05
Assertion `!uv__io_active(&stream->io_watcher, POLLIN | POLLOUT)' failed in uv__stream_destroy


renatahodovan:
###### IoT.js version:

```
Checked revision: 9c1e57c
Build command: ./tools/build.py --buildtype=debug --target-arch=i686
```
###### OS:

```
Ubuntu 16.04.3 LTS
```
###### Test case:

```javascript
var net = require('net');
var http = require('http');

net.Server.prototype.listen(3982, Boolean.prototype.toString)
http.request({host: "localhost", host: "127.0.0.1", method: "GET", path: "/", port: 5471})
```

###### Backtrace:

```
iotjs: iotjs/deps/libtuv/src/unix/stream.c:461: uv__stream_destroy: Assertion `!uv__io_active(&stream->io_watcher, POLLIN | POLLOUT)' failed.

Thread 1 "iotjs" received signal SIGABRT, Aborted.
0xf7fd8be9 in __kernel_vsyscall ()
(gdb) bt
#0  0xf7fd8be9 in __kernel_vsyscall ()
#1  0xf7d98ea9 in raise () from /lib/i386-linux-gnu/libc.so.6
#2  0xf7d9a407 in abort () from /lib/i386-linux-gnu/libc.so.6
#3  0xf7d91d07 in ?? () from /lib/i386-linux-gnu/libc.so.6
#4  0xf7d91d8b in __assert_fail () from /lib/i386-linux-gnu/libc.so.6
#5  0x080de2c5 in uv__stream_destroy (stream=0x81691f4) at iotjs/deps/libtuv/src/unix/stream.c:461
#6  0x080d8dc4 in uv__finish_close (handle=0x81691f4) at iotjs/deps/libtuv/src/unix/core.c:208
#7  0x080d8eae in uv__run_closing_handles (loop=0x81276a0 <default_loop_struct>) at iotjs/deps/libtuv/src/unix/core.c:238
#8  0x080d906e in uv_run (loop=0x81276a0 <default_loop_struct>, mode=UV_RUN_DEFAULT)
    at iotjs/deps/libtuv/src/unix/core.c:306
#9  0x08059e57 in iotjs_entry (argc=2, argv=0xffffc414) at iotjs/src/iotjs.c:217
#10 0x080596f6 in main (argc=2, argv=0xffffc414) at iotjs/iotjs_linux.c:19
```

[...]

PR #1323, opened 2017-11-23
Do not call uncaught exception handlers at exiting state.

LaszloLango:

### DIFF
--- a/src/iotjs.c
+++ b/src/iotjs.c
@@ -88,7 +88,7 @@ static bool iotjs_jerry_init(iotjs_environment_t* env) {
 }
 
 
-static void iotjs_run() {
+static void iotjs_run(iotjs_environment_t* env) {
 // Evaluating 'iotjs.js' returns a function.
 #ifndef ENABLE_SNAPSHOT
   jerry_value_t jmain = iotjs_jhelper_eval("iotjs.js", strlen("iotjs.js"),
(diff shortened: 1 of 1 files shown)


LaszloLango:
It did cause some confusion and misunderstanding that the 'process.exit' prints `uncaughtException: Abort script`. This patch fixes this. No need to call `iotjs_uncaught_exception`, when we call `process.exit()`.

zherczeg:
I think this is still a wrong fix for this issue :)

The proper fix is implementing a watchdog. The vm timeout callback should throw an error when the timeout limit is reached, not when the engine is in exit status. The two things has nothing to do with each other. We also need to add several watchdog resets to the code. The timeout default (in seconds) should be defined in compile time, but runtime set should be available as well.

LaszloLango:
@zherczeg I do not agree. This patch has nothing to do with watchdog. They are two different things.

src/iotjs.c
```diff
@@ -109,7 +109,8 @@ static bool iotjs_run(iotjs_environment_t* env) {
   }
 #endif
 
-  if (throws) {
+  IOTJS_VALIDATED_STRUCT_METHOD(iotjs_environment_t, env);
+  if (throws && (_this->state) != kExiting) {
```
daeyeon: `iotjs_environment_is_exiting` can simplify this.

zherczeg:
What you do: ignore exception when the engine exits. But this is not correct since even during exit an error can be thrown.

The problem is: when the engine exits, the current watchdog implementation aborts all js execution, including cleanup code. Hence if we do any cleanup in JS, an error is thrown by watchdog, and the cleanup does not run (bad for network cleanup), and we return early.

glistening:
@zherczeg @LaszloLango #1245 might be also related to this issue. Nodejs can exit early because it works on Linux or Windows, and so on. But in our cases (NuttX and TizenRT), we have to take care of clean up by ourselves. I will make a issue for problematic cases.

PR #1325, opened 2017-11-23
Fixed CMake file of JerryScript after the recent update.

LaszloLango:
'JERRY_CMDLINE_MINIMAL' was renamed to 'JERRY_CMDLINE_TEST' in JerryScript.


cmake/jerry.cmake
```diff
@@ -119,7 +119,7 @@ ExternalProject_Add(libjerry
     -DCMAKE_C_FLAGS=${CMAKE_C_FLAGS}
     -DENABLE_ALL_IN_ONE=ON
     -DJERRY_CMDLINE=OFF
-    -DJERRY_CMDLINE_MINIMAL=OFF
+    -DJERRY_CMDLINE_TEST=OFF
```
akosthekiss: Why was/is JERRY_CMDLINE_{MINIMAL,TEST} even mentioned when IoT.js does not use it anyhow? The default is OFF, so why repeat here? Removing the repetition of defaults would lessen the maintenance burden on IoT.js.

LaszloLango: IMHO the original reason was to ensure it is turned off even if the default changes on a Jerry update.

glistening: +1 for @akosthekiss's comment. What about removing `DJERRY_COMDLINE` and `DJERRY_CMDLINE_TEST`?

LaszloLango: OK, I'll update this, but we should keep `-DJERRY_CMDLINE=OFF`, because it is set to `ON` by default and IoT.js does not need it, so it only slows down the build.

yichoi:
LGTM after @akosthekiss confirmed

yichoi:
@akosthekiss ?

PR #1326, opened 2017-11-23
Rework error handling.

LaszloLango:


src/iotjs_binding_helper.c
```diff
@@ -28,24 +28,24 @@ void iotjs_uncaught_exception(iotjs_jval_t jexception) {
   IOTJS_ASSERT(jerry_value_is_function(jonuncaughtexception));
 
   iotjs_jargs_t args = iotjs_jargs_create(1);
+  jerry_value_clear_error_flag(&jexception);
```
daeyeon: I guessed `jexception` should be released somewhere when this function is invoked like `iotjs_uncaught_exception(jerry_acquire_value(jres));` Can you clarify my query?

src/iotjs_binding.c
```diff
@@ -105,16 +105,9 @@ iotjs_jval_t iotjs_jval_create_function(JHandlerType handler) {
...
-  return iotjs_jval_create_error_type(JERRY_ERROR_COMMON, msg);
-}
-
-
-iotjs_jval_t iotjs_jval_create_error_type(jerry_error_t type, const char* msg) {
-  iotjs_jval_t jval;
-
-  const jerry_char_t* jmsg = (const jerry_char_t*)(msg);
-  jval = jerry_create_error(type, jmsg);
+iotjs_jval_t iotjs_jval_create_error_without_error_flag(const char* msg) {
```
glistening: @LaszloLango Do we need to rename `iotjs_jval_create_error` to `iotjs_jval_create_error_without_error_flag`? We always create error that has cleared error flag. We may introduce `iotjs_jval_create_error_without_error_flag` later when we need `iotjs_jval_create_error_with_error_flag`.

src/iotjs_binding_helper.c
LaszloLango: No, `jerry_value_clear_error_flag` will release it and gives a new value.

src/iotjs_binding.c
LaszloLango: IMHO using errors without error flags in the whole code base is a bad practice. My motivation was to clarify this part. Do you think we should keep to current logic and never use errors with error flags?

src/iotjs_binding_helper.c
daeyeon: The api reference, https://github.com/jerryscript-project/jerryscript/blob/master/docs/02.API-REFERENCE.md#jerry_value_clear_error_flag, says it clears the error flag. And the example releases the value after calling `jerry_value_clear_error_flag`. Can you kindly tell me what I am still misunderstanding?

src/iotjs_binding.c
glistening: I think there is no change in logic for creating error. Previously it created the error without error flag, currently it is same. The only change is name (`iotjs_jval_create_error` to `iotjs_jval_create_error_without_error_flag`). Do you prefer `iotjs_jval_create_error_without_error_flag` since it is more descriptive name for clarification?

zherczeg: I think Laszlo prefers to use the error flag logic built-into the JerryScript. The motivation of having a separate throws flag was supporting multiple engines. However now the flag is obsolete, and we can use the error management system of JerryScript.

src/iotjs_binding_helper.c
zherczeg: Maybe we could add more description to the help. The idea is we provide one reference by releasing the old reference and creating a new one. From user perspective: "I still have one reference, just don't use the old value from now on".

src/iotjs_binding.c
glistening: @zherczeg @LaszloLango Still, I don't understand. Another question raises from @zherczeg's comment. `iotjs_jval_create_error_without_error_flag` will create error object using `jerry_create_error`, then decrease reference count by `jerry_value_clear_error_flag`. Is it correct?

src/iotjs_binding_helper.c
LaszloLango: @daeyeon we investigated this and found a bug in JerryScript. Let's hold this PR until it is fixed.

LaszloLango:
We can reopen this PR after https://github.com/jerryscript-project/jerryscript/pull/2131 is merged and pulled into IoT.js

src/iotjs_binding.c
LaszloLango: @glistening there are three ways to get error values.
i) It comes from javascript (`jerry_run` return with an error value)
ii) Thrown by `jerry_create_error` when `JS_CREATE_ERROR` is used
iii) Not thrown, but only created by `iotjs_jval_create_error[_without_error_flag]`.

The third case is only needed (if i see this correctly), when module callback must be called with an error to handle.

src/iotjs_binding_helper.c
LaszloLango: Fixed.

LaszloLango:
@daeyeon @glistening I've reopened and updated the PR, please check again.

zherczeg:
Looks reasonable.

src/iotjs_binding_helper.c
```diff
@@ -28,24 +28,25 @@ void iotjs_uncaught_exception(jerry_value_t jexception) {
...
+  if (jerry_value_has_error_flag(jres)) {
     iotjs_environment_t* env = iotjs_environment_get();
 
     if (!iotjs_environment_is_exiting(env)) {
       iotjs_set_process_exitcode(2);
       iotjs_environment_go_state_exiting(env);
     }
   }
+
+  jerry_release_value(jexception);
```
glistening: @LaszloLango Before this PR, it was `iotjs_uncaught_exception(jexception)` caller's role to release jexception. After this PR, the 'iotjs_uncaught_exception` will always release `jexception` parameter internally. So the user have to acquire the jexception before call. (jerry_acquire_value is added in that reason) Usually I don't expect my parameter will be released implicitly by calling some function. Could you explain why do you want to change the behavior? 

I prefer not to release the parameter, if there is compelling reason. Because if some functions releases the parameters, while others not, it will make us confused. And it would be better to emphasize the function user. "It will release your parameter." by comment or by using descriptive function name. (In that case, I would choose latter. Because users don't read function comment, and comment is to be out of date.)

LaszloLango: AFAIK `iotjs_uncaught_exception` is an internal function an shouldn't be used by a user. The user must register a callback, but the IoT.js must call the callback. So I do not understand your concerns.

Technical details:
When you clear the error flag it will give back a new `jerrry_value_t` and if there where no other reference, than the original value is freed. So the `jmain` will contain an invalid (freed) value at the call site of `iotjs_uncaught_exception`  (`iotjs_run`). This is why we use `jerry_acquire_value` before calling `iotjs_uncaught_exception`.

The other way would be to give back the return value of `jerry_value_clear_error_flag`.

`jerry_value_t iotjs_uncaught_exception(jerry_value_t jexception)`
or
`void iotjs_uncaught_exception(jerry_value_t* jexception)`

glistening: @LaszloLango 
When I mentioned a user, I mean the iotjs core developer like me. There were many changes in iotjs by many developers. I found several different style and convention. So I think it is important to keep the code consistent, and readable. I would like to give effort in our internal code, not only for external api.

This change makes it difficult to read. Imagine my string is release after I call function `foo(myStr)`, and `foo` don't provide comment, and cannot guess it will release my string from `foo`.

Could you please answer to my original question?
Still I cannot find what is the benefit of moving release into  `iotjs_uncaught_exception`.

LaszloLango: Sorry I thought I could answer that question. :) The JerryScript API is intended to use in this way. What about the suggested alternatives?

>The other way would be to give back the return value of jerry_value_clear_error_flag.
>
>jerry_value_t iotjs_uncaught_exception(jerry_value_t jexception)
>or
>void iotjs_uncaught_exception(jerry_value_t* jexception)

Would it be more readable for you?

glistening: @LaszloLango Hmm, my confusion comes from the modified semantics of `jerry_value_clear_error_flag`. It introduced several issues(https://github.com/jerryscript-project/jerryscript/issues/2080, https://github.com/jerryscript-project/jerryscript/pull/2131) and added an API(https://github.com/jerryscript-project/jerryscript/pull/2096). I need to take a closer look for this API and implementation to review this PR.

glistening: @LaszloLango The comment `jerry_get_value_without_error_flag` (https://github.com/jerryscript-project/jerryscript/issues/2080#issuecomment-343445688) was very helpful. What about using `jerry_get_value_without_error_flag` on caller of `iotjs_uncaught_exception` like code below?

```C
jerry_value_t errval = jerry_get_value_without_error_flag(jres);
iotjs_uncaught_exception(errval);
jerry_release_value(errval);
```

yichoi:
@LaszloLango rebase please

src/iotjs_binding_helper.c
zherczeg: If I understand you correctly and you wish to keep "errval" alive, then iotjs_uncaught_exception() could do the reference increase itself. So you can do the usual "call free for any jres".

```
val = something()
if (jerry_is_value_error(val))
  iotjs_uncaught_exception(val);
jerry_release_value(val);
```

glistening: @zherczeg Right.

I feel odds when I read the code because:

`err_val` (which is passed to `iotjs_uncaught_exception` by value) is released implicitly. Therefore we have to increase reference count for `err_val` before calling `iotjs_uncaught_exception`. 

I think there are two approach to explain the code more clearly.

(1) increase `err_val` in `iotjs_uncaught_exception()` (@zherczeg's https://github.com/Samsung/iotjs/pull/1326#discussion_r160096745)
(2) use `jerry_value_get_without_error_flag` for getting a copy of err_val without error flag, and passing it to `iotjs_uncaught_exception` (my https://github.com/Samsung/iotjs/pull/1326#discussion_r159368404)

Either works for me.

@LaszloLango I don't saying your implementation is wrong. But I would like to write the code differently so that others who are not accustomed for jerry could understand more easily. Because most don't know the details that you explained on https://github.com/Samsung/iotjs/pull/1326#discussion_r156869477

LaszloLango: Sorry for the late answer. I missed your previous comment. I'll update the PR as you suggested.

LaszloLango:
@glistening I've updated the PR. Please check.

yichoi:
LGTM

Issue #1328, opened 2017-11-24
`./tools/build.py --jerry-debugger` fails


daeyeon:
Running `./tools/build.py --jerry-debugger` fails with the following errors.

```
error: implicit declaration of function ‘nanosleep’ [-Werror=implicit-function-declaration]
   nanosleep (&(const struct timespec)
error: invalid use of undefined type ‘struct timespec’
```
It's seemingly occurred since #1315. Do we need to consider a travis check for this?

LaszloLango:
@daeyeon We also recognized that the build fails with `--jerry-debugger` since the "all in one" build was enabled. AFAIK, @galpeter is working on it.

galpeter:
Currently I don't have any good/clean solution. So I think we should revert the all-in-one build for now until we have a good resolution. I'll create a PR to revert it.

PR #1330, opened 2017-11-24
Revert the all in one build for JerryScript

galpeter:
With the all in one build enabled for the
JerryScript the debugger build fails.

Until a good solution is found revert the change to
fix the build.


galpeter:
Related:  #1328

yichoi:
LGTM

PR #1334, opened 2017-11-28
Fix an invalid type conversion

hs0225:
'device' is cast to integer pointer in i2c_create_platform_data function


yichoi:
LGTM

PR #1341, opened 2017-12-01
Improve precommit testing

LaszloLango:
Add a new option which can modify or append the default list of tests,
so the 'precommit.py' can run different test sets based on the build
options.


LaszloLango:
Changes must adapt also to `testrunner.py`, but I'd like hear your option about this PR at first. :) The motivation of this work was to catch the errors early. Unfortunately we realized some bugs only several days after it was merged and revealed that our current precommit testing is not thorough enough. I think precommit testing can be more accurate with additional test set files. What do you think?

LaszloLango:
This patch revealed a new issue: #1353

LaszloLango:
I've updated the PR. The docker image must be updated to work properly on travis. @hs0225 could you add [jsonmerge](https://pypi.python.org/pypi/jsonmerge) to the docker image? (`pip install jsonmerge`)

yichoi:
LGTM

PR #1346, opened 2017-12-05
Fix creating http request

daeyeon:
This patch mainly adds `Host` field into the http request header.


yichoi:
LGTM

PR #1357, opened 2017-12-06
Check the given path in fs.exists

### LINKED ISSUE #1355: Assertion 'JERRY_CONTEXT (jmem_heap_allocated_size) == 0' failed in jmem_heap_finalize
###### IoT.js version:

```
Checked revision: dbd52a7
Build command: tools/build.py --buildtype=debug
```
###### OS:

```
Ubuntu 17.10
```
###### Test case:

```javascript
var fs = require('fs');

try { fs.exists(function() {}, Date.prototype.getUTCMinutes) } catch (err) {  }
try { fs.exists("test/resources/tobeornottobe.txt") } catch (err) {  }

```

###### Backtrace:

```
uncaughtException: TypeError: Date object expected
uncaughtException: Abort script
ICE: Assertion 'JERRY_CONTEXT (jmem_heap_allocated_size) == 0' failed at iotjs//deps/jerry/jerry-core/jmem/jmem-heap.c(jmem_heap_finalize):181.
Error: ERR_FAILED_INTERNAL_ASSERTION

(gdb) bt
#0  jmem_heap_finalize () at iotjs/deps/jerry/jerry-core/jmem/jmem-heap.c:181
#1  0x00005555555a79a6 in jmem_finalize () at iotjs/deps/jerry/jerry-core/jmem/jmem-allocator.c:51
#2  0x00005555555830f4 in jerry_cleanup () at iotjs/deps/jerry/jerry-core/api/jerry.c:220
#3  0x000055555556f421 in iotjs_jerry_release (env=0x555555841440 <current_env>) at iotjs/src/iotjs.c:92
#4  0x000055555556f9f0 in iotjs_entry (argc=2, argv=0x7fffffffdec8) at iotjs/src/iotjs.c:242
#5  0x000055555556f0da in main (argc=2, argv=0x7fffffffdec8) at iotjs/iotjs_linux.c:19

```

<sup>Found by [Fuzzinator](http://fuzzinator.readthedocs.io/)</sup>

DanielBallaSZTE:
The path argument passed to fs.exists must be either a String or a Buffer
fixes #1355 

### DIFF
--- a/src/js/fs.js
+++ b/src/js/fs.js
@@ -20,6 +20,9 @@ var util = require('util');
 var fsBuiltin = native;
 
 fs.exists = function(path, callback) {
+  if (!(util.isString(path)) && !(util.isBuffer(path))) {
+    throw new TypeError('Path should be a string or a buffer');
+  }
   if (!path || !path.length) {
     process.nextTick(function() {
       if (callback) callback(false);


src/js/fs.js
```diff
@@ -20,6 +20,11 @@ var util = require('util');
 var fsBuiltin = native;
 
 fs.exists = function(path, callback) {
+  if ((!(util.isString(path)) && !(util.isBuffer(path)))
```
galpeter: I'm not sure this is a good solution: shouldn't the error be reported via the callback?

glistening: @galpeter I think we must consider the case when user does not pass callback ( It is strange, but it is possible). In this case, we have no way to inform the user there is an error since there is no callback. Node.js seems throwing TypeError for this case. Here is my test.

```sh
$ node -e 'fs = require('fs');fs.exists(3.14159);'

assert.js:81
  throw new assert.AssertionError({
  ^
AssertionError: path must be a string
    at Module.require (module.js:496:3)
    at require (internal/module.js:20:19)
    at [eval]:1:6
    at ContextifyScript.Script.runInThisContext (vm.js:25:33)
    at Object.runInThisContext (vm.js:97:38)
    at Object.<anonymous> ([eval]-wrapper:6:22)
    at Module._compile (module.js:570:32)
    at evalScript (bootstrap_node.js:354:27)
    at run (bootstrap_node.js:123:11)
    at run (bootstrap_node.js:390:7)
```
If I misunderstand what you mean, could you explain in code what do you mean by the error reported via the callback?

galpeter: @glistening I was unsure how the node works. That is for example: ` fs.exists(3, function(er) { .. });` will the error be reported via callback or not. But your example clarifies that such error is always reported in a sync fashion.

So this solution looks good.

src/js/fs.js
```diff
@@ -20,6 +20,11 @@ var util = require('util');
...
+      || (util.isFunction(path))) {
```
galpeter: ohh another thing: Do we need to explicitly check fo function here? Shouldn't the previous check be enough?

DanielBallaSZTE: Hmm, yes, you are definitely right, thanks for highlighting that.

DanielBallaSZTE:
@glistening Hmm, the mentioned issue still seems flaky, I think only one half is solved with this PR, so I'm going to look further into it. It sometimes still throws an assertion error, if a valid string is given as the path.

src/js/fs.js
```diff
@@ -20,6 +20,10 @@ var util = require('util');
 var fsBuiltin = native;
 
 fs.exists = function(path, callback) {
+  if (!(util.isString(path)) && !(util.isBuffer(path))) {
+    throw new TypeError("Path should be a string or a buffer");
```
yorkie: Should we follow single-quoted for string

src/js/fs.js
```diff
@@ -20,6 +20,10 @@ var util = require('util');
...
+    return;
```
yorkie: `return` should not be after the `throw` statement :(

src/js/fs.js
glistening: @yorkie Good point. In fact, there had been no rule for quote for string. However, we are preparing the rule like https://github.com/Samsung/iotjs/blob/master/.eslintrc.js#L56. Perhaps we can fix them all at once with eslint.

src/js/fs.js
glistening: @yorkie You're right! @DanielBallaSZTE Please fix it.

src/js/fs.js
yorkie: Okay, I'm fine with your proposal :)

DanielBallaSZTE:
@yorkie Thanks for the review, sorry for the late fixes, I've been on holiday. :)

PR #1358, opened 2017-12-07
Add an error explanation on http client

daeyeon:
This patch mainly adds and improves the error handling for dns.lookup and net.socket errors. Users can find the basic clue for errors instead of `socket hang up`.

### DIFF
--- a/src/js/http_client.js
+++ b/src/js/http_client.js
@@ -71,21 +71,23 @@ exports.ClientRequest = ClientRequest;
 function setupConnection(req, socket) {
   var parser = common.createHTTPParser();
   parser.reinitialize(HTTPParser.RESPONSE);
-  req.socket = socket;
-  req.connection = socket;
-  parser.socket = socket;
-  parser.incoming = null;
-  parser._headers = [];
-  req.parser = parser;
-
   socket.parser = parser;
   socket._httpMessage = req;
 
+  parser.socket = socket;
+  parser.incoming = null;
+  parser._headers = [];
   parser.onIncoming = parserOnIncomingClient;
+
+  req.socket = socket;
+  req.connection = socket;
+  req.parser = parser;
+
   socket.on('error', socketOnError);
   socket.on('data', socketOnData);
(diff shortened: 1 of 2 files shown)


src/js/http_client.js
```diff
@@ -56,8 +56,8 @@ function ClientRequest(options, cb) {
   // Create socket.
   var socket = new net.Socket();
 
-  // connect server.
-  socket.connect(port, host);
+    // connect server.
+    socket.connect(port, host);
```
LaszloLango: accidental change?

daeyeon: Right, it's fixed. Thanks.

test/run_pass/test_net_httpclient_parse_error.js
```diff
@@ -39,7 +39,6 @@ request.on('error', function(err) {
 request.end();
 
 process.on('exit', function() {
-  // The first error is a Parse Error.
-  // The second error is the socket hang up.
-  assert.equal(errors, 2);
+  // The error is a Parse Error.
+  assert.equal(errors, 1);
```
glistening: @daeyeon It seems the API behavior is changed by this PR. Could you explain what is changed?

daeyeon: @glistening We roughly throw an `Error` instance with the `socket hang up` message in case that a socket is closed with no response. For instance, that cover cases such as `dns look up error` , `the connection is blocked by server`, and so on. https://github.com/Samsung/iotjs/blob/84600faea6a30352a764c120db8d9bd2965deccd/src/js/http_client.js#L115

However, in the case of `http parsing error`, we have a place throwing the exact error. Thus, in the normal manner, throwing the `Parse Error` only was right. I also checked this test using nodejs, and the result is also `1`, `Parse Error`.

glistening: @daeyeon Thank you for clear explanation.

yichoi:
LGTM

PR #1366, opened 2017-12-13
Make magic string definition consistent

glistening:


glistening:
I found some inconsistent usages in iotjs magic string. I fixed it.

A question occurred: 

"Why we use iotjs_magic_string.h?"

I don't think it is part of [jerryscript magic string](https://github.com/jerryscript-project/jerryscript/blob/master/docs/04.INTERNALS.md#string). Please let me know there is an mechanism that makes iotjs magic string a part of jerryscript magic string.

If not, for binary size? readability? maintenance?
I cannot find benefit from iotjs magic string.
Could anyone explain it?

LaszloLango:
They are part of the registered magic stings. Related PR: #694
Related code from the mentioned PR:
```diff --git a/tools/js2c.py b/tools/js2c.py
index 938aced..e1139c9 100755
--- a/tools/js2c.py
+++ b/tools/js2c.py
@@ -210,7 +210,14 @@ def get_js_contents(name, is_debug_mode=False):
 
 def js2c(buildtype, no_snapshot, js_modules, js_dumper, verbose=False):
     is_debug_mode = buildtype == "debug"
-    magic_string_set = { b'process' }
+    magic_string_set = set()
+
+    str_const_regex = re.compile('^#define IOTJS_MAGIC_STRING_\w+\s+"(\w+)"$')
+    with open(fs.join(path.SRC_ROOT, 'iotjs_magic_strings.h'), 'r') as fin_h:
+        for line in fin_h:
+            result = str_const_regex.search(line)
+            if result:
+                magic_string_set.add(result.group(1))
 
     # generate the code for the modules
     with open(fs.join(path.SRC_ROOT, 'iotjs_js.h'), 'w') as fout_h, \
```

glistening:
@LaszloLango Thank you for reminding and referencing. A memory occurred again. :)

glistening:
@LaszloLango 

Currently we defines all strings in `iotjs_magic_strings.h`. For example, there are several module specific strings in `iotjs_magic_strings.h`, none of them are enclosed with corresponding macros. Strings may be included in binary when the corresponding module is not included at all. If it is true, I would like to enhance this situation by one of followings:

1) Remove module specific strings from magic string header, only keep the basic modules' strings
2) Keep all strings in magic string, but enclose with module enabler macro

What do you think about it?

glistening:
@LaszloLango 

I compared the `nuttx.bin` binary size with current and previous one.
Reducing magic string seems to reduce binary size.

#### build command (build all modules)
``
$ ./tools/precommit.sh --test=nuttx --buildtype=release
``

#### nuttx.bin size in byte

|  | |
| --- | --- |
| current | 395,380 |
| revert js2c.py change | 394,700 |

LaszloLango:
@glistening when you revert the change, then the binary gain will show in the runtime memory usage. The magic strings are stored in the ROM instead of RAM. Usage of magic strings reduces the runtime memory consumption and improves the performance.

LaszloLango:
> Keep all strings in magic string, but enclose with module enabler macro
This is a good idea.

glistening:
@LaszloLango Yes, I know magic string sacrifices the binary size for runtime memory and performance. We may need tuning for this when the requirement is clear. Thank you for review. I will make another PR for enclosing magic strings to reduce binary size. It is out of scope of this PR.

yichoi:
LGTM

PR #1368, opened 2017-12-13
Move iotjs_linux.c from root directory into linux directory

glistening:
iotjs_linux.c is the only source code in root directory.
It should be located under src.


yichoi:
LGTM

PR #1373, opened 2017-12-15
Add auto-release feature

daeyeon:
This suggests auto-release feature for `jerry_value_t`, `iotjs_jargs_t`,
and `iotjs_string_t`. Although jerryscript has the same for `jerry_value_t`,
it seems better for us to have our own one. Because the jerry feature is
as small pieces of code as we can easily handle. Also, we need to add
more types into the implementation.

This is applied only to the `buffer` module to listen to opinions.

### DIFF
--- a/src/iotjs_binding.h
+++ b/src/iotjs_binding.h
@@ -224,4 +224,22 @@ jerry_value_t iotjs_jhelper_eval(const char* name, size_t name_len,
 
 jerry_value_t vm_exec_stop_callback(void* user_p);
 
+#ifdef __GNUC__
+static inline void jerry_autorelease_value(const jerry_value_t* value) {
+  jerry_release_value(*value);
+}
+
+#define jerry_value_auto_t \
+  jerry_value_t __attribute__((__cleanup__(jerry_autorelease_value)))
+
+#define iotjs_jargs_auto_t \
+  iotjs_jargs_t __attribute__((__cleanup__(iotjs_jargs_destroy)))
+
+#define iotjs_string_auto_t \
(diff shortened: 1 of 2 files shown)


src/iotjs_binding.h
```diff
@@ -224,4 +224,22 @@ jerry_value_t iotjs_jhelper_eval(const char* name, size_t name_len,
 
 jerry_value_t vm_exec_stop_callback(void* user_p);
 
+#ifdef __GNUC__
```
glistening: @daeyeon Does clang support `__cleanup__`? If not, it will break build support for macOS.

glistening: @daeyeon Currently macOS build is ok (Travis `host-darwin` build passes.) because clang defines `__GNUC__` for GCC compatibility. But it does not guarantee it will work correctly. The `__cleanup__` may be ignored. We need to check it.

glistening:
This PR uses `gcc (clang)` specific attributes. Merging this PR means we will not support some devices that does not have `gcc (clang)` any more. I would like to get other's opinion.

yichoi:
+1 for @glistening

zherczeg:
While this change improves maintainability, it will limit the spreading of iotjs by moving away from standard C. Perhaps the question is: how much time is spent on tracking unfreed values? And that is also a question how much this improves readability.

An alternative could be: would it be possible to create a GCC specific warning system, which warns you if a value is not freed?

daeyeon:
I understand other's concerns. So I'd like to close this.

PR #1375, opened 2017-12-18
module: support node_modules and fix index loading

yorkie:
Add the following 2 features:

- load modules from the `node_modules` directories
- load this kind of module `foobar/index.js` by `./foobar`

### DIFF
--- a/src/js/module.js
+++ b/src/js/module.js
@@ -41,9 +41,14 @@ var moduledirs = [''];
 if (cwd) {
   moduledirs.push(cwd + '/');
   moduledirs.push(cwd + '/iotjs_modules/');
+  moduledirs.push(cwd + '/node_modules/');
 }
 if (process.env.HOME) {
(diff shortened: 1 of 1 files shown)


glistening:
@yorkie We have discussed several times about this policy. Please see https://github.com/Samsung/iotjs/pull/1305. We decided not to load `node_modules` by default.

yorkie:
> We decided not to load node_modules by default.

So do we have a plugin way to add the directories to load modules?

BTW, how about the another bugfix in this PR, when we have a directory: 

```
|-foo
   |- index.js
```

Then the following requires is working on Node.js:

```js
require('./foo');
```

glistening:
@yorkie Let's handle the two items separately. 

1. Providing additional module load directory
We have discussed internally several times. We need to talk about it more on separate issue

2. search index.js under specified directory by default
I think we could add this feature. But I would like to other's opinions.

yorkie:
Okay, I will submit separately later :)

PR #1376, opened 2017-12-18
Export remotely received source as a module

DanielBallaSZTE:
Previously remotely received sources were directly called.
This solution presented an issue, where requiring other modules inside the remote source was not possible.
This patch fixes it, and also opens the door to a follow up patch, handling multiple sources to be sent.


DanielBallaSZTE:
Note: Currently, there's an issue in iotjscode not trimming the end of the text properly, sending some corrupted string at the end of the source, so the patch will fail in some cases until I make a fix in iotjscode.

src/js/module.js
```diff
@@ -170,54 +170,77 @@ iotjs_module_t.tryPath = function(path) {
...
+  }
 };
 
 
 iotjs_module_t.runMain = function() {
   if (process.debuggerWaitSource) {
-    var fn = process.debuggerSourceCompile();
-    fn.call();
+    var fn = process.getRemoteSource();
+    iotjs_module_t.load(fn, null, true);
```
galpeter: I think it would be better to have a separate function for "remote load" so then we don't need to pass around the third parameter.

src/js/module.js
```diff
@@ -170,54 +170,77 @@ iotjs_module_t.tryPath = function(path) {
...
+    return module.exports;
+  }
 };
 
 
-iotjs_module_t.prototype.compile = function() {
-  var source = process.readSource(this.filename);
-  var fn = process.compile(this.filename, source);
-  fn.call(this.exports, this.exports, this.require.bind(this), this);
+iotjs_module_t.prototype.compile = function(remote, _filename, _source) {
```
galpeter: This could be rewritten to receive the filename and source which should be compiled and not the remote information. Then the caller function should read the source if required and thus there is no need for the if-else block here.

src/iotjs_magic_strings.h
```diff
@@ -65,7 +65,7 @@
 #define IOTJS_MAGIC_STRING_CREATETCP "createTCP"
 #define IOTJS_MAGIC_STRING_CWD "cwd"
 #define IOTJS_MAGIC_STRING_DATABITS "dataBits"
-#define IOTJS_MAGIC_STRING_DEBUGGERSOURCECOMPILE "debuggerSourceCompile"
+#define IOTJS_MAGIC_STRING_GETREMOTESOURCE "getRemoteSource"
```
galpeter: I would like to keep the debugger prefix, maybe name it `debuggerGetSource`?

DanielBallaSZTE:
@galpeter Nice ideas, thanks for the review. I've updated the Pull Request.

src/js/module.js
```diff
@@ -204,18 +202,36 @@ iotjs_module_t.load = function(id, parent) {
   return module.exports;
 };
 
+iotjs_module_t.loadRemote = function(id, parent) {
```
galpeter: Two questions:
* Do we need the parent argument at all?
* Can we have parameters like this: "function(filePath, source)"?
  * Maybe we'll need to rename this then to something else.

src/js/module.js
```diff
@@ -204,18 +202,36 @@ iotjs_module_t.load = function(id, parent) {
...
+  if (process.builtin_modules[id[0]]) {
```
galpeter: Can this case occur at all?

src/js/module.js
DanielBallaSZTE: Good point, parent would be `null` anyways.

DanielBallaSZTE: I don't think `filePath` would be a proper name, `fileName` would fit this usecase more, since it never really exists, so doesn't have a path.

src/js/module.js
DanielBallaSZTE: Honestly, I neither do I think it would happen.

PR #1377, opened 2017-12-19
Enclose magic strings with module enabler macros

glistening:
This PR is for reducing binary size.
For example, it reduced about 1,016 bytes for minimal profile binary on NuttX.


glistening:
It is the implementation of https://github.com/Samsung/iotjs/pull/1366#issuecomment-351613877.

#### nuttx.bin size (in bytes)
| | |
|---|---|
|Before|316,688|
|After|315,672|

PR #1381, opened 2017-12-22
Group the magic strings by module

galpeter:
This change serves two things:
1) Fix the -Wexpansion-to-defined warnings.
2) Prepare the IoT.js to support per module magic strings.


glistening:
@galpeter 
First, thank you for fixing warning that I was not aware of.
For grouping magic string, I also thought about it when I prepare #1377. But I decided not to grouping for reasons below:
1. I found some strings belongs to multiple modules. Where these strings should be?
2. I prefer alphabetically sorted way, because it makes me find the string fast. After grouping, we have to search the candidate string whether it exists or not.

Anyway, it is my preference. I am ok if others also like grouping.

yichoi:
@galpeter rebase please

galpeter:
Yes, I'm using a newer GCC (7+) thus the error message. I we don't want grouping we could also use the following solution:
```
#if ENABLE_MODULE_GPIO
```
that is to remove the `defined` part as currently we define all module as 0 if they are not enabled.

Sadly I don't see a better way to resolve the compiler warning/error (maybe we could suppress it but that could be dangerous).

galpeter:
@yichoi @glistening I've created an alternative solution: https://github.com/Samsung/iotjs/pull/1411 which could be used instead of this.

glistening:
@galpeter I like #1411 which solves all. Thank you.

glistening:
@galpeter #1411 landed. Could you please close this PR?

yichoi:
Closing. I think all including @galpeter agree that #1411 is better.

PR #1385, opened 2018-01-02
Add Cmake files to support Tizen gbs build

haesik:
- Adding new target rpi3 for TizenIoT
- Fix Tizen version for GBS build to TizenIoT preview1


CMakeLists.txt
```diff
@@ -79,6 +79,8 @@ elseif("${TARGET_ARCH}" STREQUAL "i686")
   iotjs_add_compile_flags(-D__i686__ -D__x86__ -march=i686 -m32)
 elseif("${TARGET_ARCH}" STREQUAL "x86_64")
   iotjs_add_compile_flags(-D__x86_64__)
+elseif("${TARGET_ARCH}" STREQUAL "gbs")
+  iotjs_add_compile_flags(-D__gbs__)
```
glistening: @haesik Could you explain why `__gbs__` is defined here? I cannot find __gbs__ enclosed code in this PR. Furthermore identifier with leading double underscores is reserved by standard.

haesik: I Just add for the future use..  Ok, I will remove it now..

cmake/config/gbs-tizen.cmake
```diff
@@ -0,0 +1,18 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+include(CMakeForceCompiler)
+
+set(CMAKE_SYSTEM_NAME Tizen)
+set(CMAKE_SYSTEM_PROCESSOR armv7l)
```
glistening: @haesik What about `arm-tizen-gbs.conf` to conform naming convention? I prefer to use `arm-tizen.cmake` and handle both of gbs and native build in same file. I would remove native build code if we don't support it any more.

glistening: @haesik I prefer to use `arm-tizen.cmake` that contains gbs build only. Because gbs is the official way to build tizen package. But if there is reason to support both of gbs and non-gbs (ordinary arm linux binary), `arm-tizen-gbs.conf` would be better to specify that it is for gbs build.

Next, when we prepared to support x86 in future, I still prefer to put x86-tizen and arm-tizen in `x86-tizen.conf` and `arm-tizen.conf` respectively. Because it is more clear and other existing config follow the convention (e.g., `x86_64-linux.conf` and `arm-linux.conf`) in same directory. You may merge two files into one file under whatever you name for it. But for now, it is only for arm/tizen (with gbs).

haesik: I don't agree with your opinion of specifying arm or x86.. if we use ar-tizen.conf x86-tizen.conf, we should make arm-tizen.conf, x86-tizen.conf, x86_64-tizen, etc.. for the same gbs build. But I agree to use noarch.. so, ok I will use noarch instead of gbs.

CMakeLists.txt
```diff
@@ -79,6 +79,7 @@ elseif("${TARGET_ARCH}" STREQUAL "i686")
   iotjs_add_compile_flags(-D__i686__ -D__x86__ -march=i686 -m32)
 elseif("${TARGET_ARCH}" STREQUAL "x86_64")
   iotjs_add_compile_flags(-D__x86_64__)
+elseif("${TARGET_ARCH}" STREQUAL "noarch")
```
LaszloLango: `noarch` is quite odd for me. Can't we define a valid architecture for Rpi3? Do you want to use it for `aarch64`  or `armv8` systems?

glistening: @LaszloLango 

AFAIK, we should not pass architecture specific options directly when we use `gbs`. `gbs` handles target specific options in its own control. If we pass architecture specific option, it would make conflicts. It's the reason @haesik stated he would like to use a common file for all architecture(https://github.com/Samsung/iotjs/pull/1385#discussion_r159364015), and I suggested `noarch` (https://github.com/Samsung/iotjs/pull/1385#discussion_r159333676) for this purpose.

If you were like me, you may feel odd about several `arm` specific options like

```cmake
set(CMAKE_SYSTEM_PROCESSOR armv7l)
```

```cmake
if("${TARGET_OS}" STREQUAL "tizen")
  iotjs_add_compile_flags(-mfpu=neon-vfpv4 -mfloat-abi=softfp)
endif()
```
The former is removed by my offline comment, but the latter still exists.

glistening: @LaszloLango Do you need more explanation or have any siggestion?

LaszloLango: @glistening thanks for the explanation.

PR #1387, opened 2018-01-02
Multiple source file sending

DanielBallaSZTE:
This patch makes iot.js capable of receiving multiple sources from the debugger client.
These sources can also require each other.

### DIFF
--- a/src/js/module.js
+++ b/src/js/module.js
@@ -28,6 +28,7 @@ module.exports = iotjs_module_t;
 
 
 iotjs_module_t.cache = {};
+iotjs_module_t.remoteCache = {};
 
 
 var cwd;
@@ -210,7 +211,7 @@ iotjs_module_t.loadRemote = function(filename, source) {
 
   module.filename = filename;
   module.compile(filename, source);
-  iotjs_module_t.cache[filename] = module;
+  iotjs_module_t.remoteCache[filename] = module;
 
   return module.exports;
 };
@@ -225,7 +226,9 @@ iotjs_module_t.prototype.compile = function(filename, source) {
 iotjs_module_t.runMain = function() {
   if (process.debuggerWaitSource) {
     var fn = process.debuggerGetSource();
-    iotjs_module_t.loadRemote(fn[0], fn[1]);
+    fn.forEach(function (e) {
+      iotjs_module_t.loadRemote(e[0], e[1]);
+    });
   } else {
     iotjs_module_t.load(process.argv[1], null);
   }
--- a/src/modules/iotjs_module_process.c
+++ b/src/modules/iotjs_module_process.c
@@ -78,9 +78,28 @@ static jerry_value_t wait_for_source_callback(
 
 // Export JS module received from the debugger client
 JS_FUNCTION(DebuggerGetSource) {
-  jerry_value_t res;
-  jerry_debugger_wait_for_client_source(wait_for_source_callback, NULL, &res);
-  return res;
+  jerry_debugger_wait_for_source_status_t receive_status;
+  jerry_value_t ret_val = jerry_create_array(0);
+  uint8_t counter = 0;
+  do {
+    jerry_value_t res;
+    receive_status =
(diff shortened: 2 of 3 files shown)


src/js/module.js
```diff
@@ -174,6 +175,9 @@ iotjs_module_t.load = function(id, parent) {
   if (process.builtin_modules[id]) {
     return Native.require(id);
   }
+  if (iotjs_module_t.remoteCache[id]) {
+    return iotjs_module_t.remoteCache[id].exports;
+  }
```
glistening: @DanielBallaSZTE `loadRemote` seems be for remote loading, and `load` is for both of local and remote. I think all logic related to remote loading could be placed in `loadRemote`. Is it possible?

DanielBallaSZTE: Yes, merging the two functions is definitely possible.
On the one hand the source loading would be handled within one function, which would be nice indeed.
On the other hand though, the function would require a couple more parameters, and several if-else cases, which -in my opinion- wouldn't look that great.

The lines you highlighted are not really for loading both remote and local modules.
Remote modules are handled only in `loadRemote`, the only thing it checks, is whether the module that would be loaded locally is already present in the remoteCache.

glistening: @DanielBallaSZTE Why do we need to check the remoteCache exists before loading local module?

yichoi:
@DanielBallaSZTE rebase please

src/js/module.js
```diff
@@ -28,6 +28,7 @@ module.exports = iotjs_module_t;
 
 
 iotjs_module_t.cache = {};
+iotjs_module_t.remoteCache = {};
```
zherczeg: Do we need a separate remote cache? Can't we just put the sources into the normal cache?

src/modules/iotjs_module_process.c
```diff
@@ -76,11 +76,50 @@ static jerry_value_t wait_for_source_callback(
...
+  jerry_value_t ret_val = jerry_create_array(counter);
+  counter = 0;
+  Sources_t* previous;
+  while (head->next != NULL) {
+    jerry_set_property_by_index(ret_val, counter++, head->file_array);
+    jerry_release_value(head->file_array);
+    previous = head;
+    head = head->next;
+    iotjs_buffer_release((char*)previous);
+  }
```
zherczeg: Could we just put the sources into the array when they are received? I mean have two steps: we create a chain list, then we copy the chain list into an array. I think we can do it in one step.

DanielBallaSZTE: Yes, you are actually right, I just realised the size of `jerry arrays` can be adjusted dynamically.

src/js/module.js
DanielBallaSZTE: The only reason there is a separate cache is for better visibility.
Since there is a separate function for loading them as well, I thought it would clearer to read. But that's just my opinion.

src/js/module.js
DanielBallaSZTE: @glistening Oh, you are right. Local modules are loaded before remotes would be, therefore there's no need to check whether a remote modules exists with the same name. Thanks for that! Somehow I missed that part.

src/modules/iotjs_module_process.c
```diff
@@ -78,9 +78,28 @@ static jerry_value_t wait_for_source_callback(
 
 // Export JS module received from the debugger client
 JS_FUNCTION(DebuggerGetSource) {
-  jerry_value_t res;
-  jerry_debugger_wait_for_client_source(wait_for_source_callback, NULL, &res);
-  return res;
+  jerry_debugger_wait_for_source_status_t receive_status;
+  jerry_value_t ret_val = jerry_create_array(1);
```
zherczeg: Default size is 1?

DanielBallaSZTE: @zherczeg You are right, there are actually some cases where even 1 element initialization is unnecessary. I've fixed it. Thanks for the tip.

yichoi:
LGTM

Issue #1229, opened 2017-09-28
Remove Travis checks in precommit.py


daeyeon:
All the travis checks in `precommit.py` are being moved to `travis_script.py`. Once that's done, `precommit.py` will not be used for travis and be renamed for another possible purpose.

LaszloLango:
@daeyeon can you clarify for me what should be done by `precommit.py` and `travis_script.py`? Why do we need two sparate scripts?

daeyeon:
@LaszloLango Please take a look at #1392 newly PRed. Some functionalities such as `flash images`, which is not related to travis tests, have been dropped from `precommit.py`. It's no longer needed for travis tests, but not removed yet. We could revise it to easily set up build environments.

LaszloLango:
What is the meaning of the precommit script? I still don't understand why do we need to maintain two script for testing which are basically do the same testing, so there are a lot of code duplication.

daeyeon:
@LaszloLango There is no intention to maintain `precommit.py` for testing. It can be removed. Do you mind if we remove it from #1392?

LaszloLango:
@daeyeon yes please. If nobody should use the `precommit.py`, then we should remove it in #1392

bzsolt:
Is there anything else what wasn't covered by #1392?

LaszloLango:
AFAIK nothing is left. We can close this issue now. @daeyeon feel free to reopen if necessary.

PR #1392, opened 2018-01-04
Move all Travis script to travis_script.py

hs0225:
'precommit.py' contains trvais script and building environment, which is confusing.
The next task is to rename 'precommit.py appropriately, and this should only be used to set the build environment.


hs0225:
related issue #1229

.travis.yml
```diff
@@ -8,38 +8,31 @@ services:
...
-  if [[ "$RUN_DOCKER" == "yes" ]]; then tools/travis_script.py;
-  else tools/precommit.py $OPTS;
-  fi
+  tools/travis_script.py
 
 env:
   global:
     - secure: "lUGzoKK/Yn4/OmpqLQALrIgfY9mQWE51deUawPrCO87UQ2GknfQ4BvwY3UT5QY0XnztPBP1+vRQ2qxbiAU7VWicp280sXDnh0FeuZD14FcE9l0FczraL12reoLu+gY5HWFfbkZncmcBsZkxDEYxhkM14FJU8fxyqGQW2ypJNz+gUGP+8r40Re5J3WjcddCQNe5IG8U+M9B4YeDHhN2QspLdN5pkgn56XtdGa3+qbecO2NpjJG5ltM9j1tTuo/Dg22DxrIFVfeFSFKUj4nfMrgPo5LevRsC/lfaBSCsj751eqrxRcQRh2hkpiIJ7mEBs2LL1EH9O6Mbj+eRh8BvIYqTB85VPNFc43sLWk14apcSVBrxJE5j3kP9sAsOD9Y5JynnkeuxYyISrkywwoX2uxsmCzIfGbwsv5VLToQzrqWlGYrHOAmVXNi8561dLfsWwxxFUjdqkZr1Kgc8UfnBEcBUtSiKCHS86/YUUbBJGkEkjDUS0GiqhFY4bXLQCR7EX4qDX3m6p7Mnh4NVUolpnSmyeYE/MjmqQ+7PJsPLL3EcIYmJ7dtW3mZ3yE2NyaFD0Pym9+TiuCCXRtrNVK1M3Kya64KNv+HbhjT/fTCgXLSeyDmJOKVAqugRlDo3b1KGR1LI0AfegzSA6mEC4e9JLjYiSnHPMUahzgLt8oU0hNFRY="
   matrix:
     - OPTS="host-linux" TARGET_OS="linux" RUN_DOCKER=yes
```
LaszloLango: `OPTS` is set but not used

tools/travis_script.py
```diff
@@ -46,44 +52,107 @@ def run_docker():
...
+                '--clean',
+                '--buildtype=' + buildtype] + args)
 
 if __name__ == '__main__':
-    run_docker()
+    if os.getenv('RUN_DOCKER') == 'yes':
+        run_docker()
 
-    test = os.environ['OPTS']
+    test = os.getenv('OPTS')
```
LaszloLango: Why aren't we using cmdline arguments instead? IMHO this is a bad practice.

LaszloLango:
The title is misleading. The patch does not move the checks, but only duplicates them.

tools/travis_script.py
hs0225: In `.travis.yml`, environment variables have been used when running scripts.

Before code,
```
`tools/precommit.py $OPTS;
```

Travis branches test with environment variables.
```
env:
   matrix:
```
If we use cmdline arguments in python script, we need to add code that handle arguments. But it also use environment variables. Do you want to put environment variables as arguments in python script as before?

LaszloLango: Yes, I think it is better, but it is only my personal preference.

daeyeon: IMHO, this script is named for travis test. So I can accept environment dependencies from travis. I assume that this script will never be locally used like how `precommit.py` was wrongly used.

.travis.yml
hs0225: `OPTS` is used at [this](https://github.com/hs0225/iotjs/blob/279ec1e309d34fe969ef2aa70b66868432f4f846/tools/travis_script.py#L74).

tools/travis_script.py
LaszloLango: @daeyeon in that case we can keep it.

yichoi:
LGTM

PR #1394, opened 2018-01-05
Rename iotjs_module_objects_t to iotjs_module_rw_data_t

glistening:
Previously, module was composed of iotjs_module_t (const data)
and iotjs_module_objects_t (read/write data). iotjs_module_objects_t
was introduced to put more data into readonly partition.

I change the names to iotjs_module_ro_data (read-only) and
iotjs_module_rw_data (read-write) to specify it more clearly.


yichoi:
LGTM

PR #1395, opened 2018-01-05
Fix mismatching member type in DebuggerConfig

mmatyas:
`jerry_debugger_init` expects `uint16` as port, but it is defined
as a regular `int` in `DebuggerConfig`.

IoT.js-DCO-1.0-Signed-off-by: Mátyás Mustoha mmatyas@inf.u-szeged.hu

### DIFF
--- a/src/iotjs_env.c
+++ b/src/iotjs_env.c
@@ -122,7 +122,7 @@ bool iotjs_environment_parse_command_line_arguments(iotjs_environment_t* env,
       size_t port_length = sizeof(strlen(argv[i] - port_arg_len - 1));
       char port[port_length];
       memcpy(&port, argv[i] + port_arg_len, port_length);
-      sscanf(port, "%d", &(_this->config.debugger->port));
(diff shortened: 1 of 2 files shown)


mmatyas:
I've just fixed the `scanf` format string causing problems on some platforms.

LaszloLango:
still lgtm

src/iotjs_env.c
```diff
@@ -122,7 +122,7 @@ bool iotjs_environment_parse_command_line_arguments(iotjs_environment_t* env,
       size_t port_length = sizeof(strlen(argv[i] - port_arg_len - 1));
       char port[port_length];
       memcpy(&port, argv[i] + port_arg_len, port_length);
-      sscanf(port, "%d", &(_this->config.debugger->port));
+      sscanf(port, "%hd", &(_this->config.debugger->port));
```
glistening: What about %hu?

mmatyas: Yeah, that'd be even better, thanks. Patch updated.

PR #1396, opened 2018-01-05
tools: fix magic_text quote problems in js2c

yorkie:
The internal JavaScript files for js2c will be crashed for:

```js
"string"/foo/bar"
```

Just replace the double-quoted(").

### DIFF
--- a/tools/js2c.py
+++ b/tools/js2c.py
@@ -353,6 +353,7 @@ def js2c(buildtype, js_modules, snapshot_tool=None, verbose=False):
         sorted_strings = sorted(magic_string_set, key=lambda x: (len(x), x))
         for idx, magic_string in enumerate(sorted_strings):
             magic_text = repr(magic_string)[1:-1]
+            magic_text = magic_text.replace('"', '\"')
 
             fout_magic_str.write('  MAGICSTR_EX_DEF(MAGIC_STR_%d, "%s") \\\n'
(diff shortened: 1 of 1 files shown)


LaszloLango:
Good catch. Please fix the DCO in the commit message.
> Signed-off-by message is incorrect. The following line should be at the end of the db01b71 commit's message: 'IoT.js-DCO-1.0-Signed-off-by: Yorkie Liu yorkiefixer@gmail.com'.

yorkie:
Fixed @LaszloLango :)

tools/js2c.py
```diff
@@ -353,6 +354,7 @@ def js2c(buildtype, js_modules, snapshot_tool=None, verbose=False):
         sorted_strings = sorted(magic_string_set, key=lambda x: (len(x), x))
         for idx, magic_string in enumerate(sorted_strings):
             magic_text = repr(magic_string)[1:-1]
+            magic_text = string.replace(magic_text, "\"", "\\\"")
```
galpeter: There is no need to import the string module, all strings have the `replace` method on them and also you can mix the `'` and `"` characters to remove a few backslashes.

example:
`magic_text.replace('"', '\"')`

yorkie:
Fixed @galpeter

yichoi:
LGTM

Issue #1389, opened 2018-01-02
WISH: Support raspbian for raspberrypi0


rzr:
Hi,
Here is a low priority wish about supporting raspbian, the issue can be closed anytime since it's more a downstream task than upstream work.

But to help community I will use this URL to share my work in progress.

More to come soon.

Note that Pi0 can also supported be by Tizen too, PM me if any interest.

Relate-to: https://github.com/Samsung/iotjs/pull/1145
Cc: @ziransun
Cc: @nherriot

rzr:
If anyone want to try on raspbian (or debian):
git clone https://github.com/tizenteam/iotjs && cd iotjs && ./debian/rules
Once verified, extra optional modules could be added.

nherriot:
I think it would be a very good 'strategic' move to support raspberry pi. Especially now IoTJS is second place in rankings to cylonjs.com - it's now the 3rd biggest selling personal compute platform 'ever' !
https://www.raspberrypi.org/magpi/raspberry-pi-sales/

rzr:
Yea, I agrea it would be a good target to extend support, I have a plan to get it available directly from debian and derivates (raspbian etc), It's on my list I will keep you updated stay tunned at:
https://blogs.s-osg.org/author/pcoval

yichoi:
IoT.js supports RPi2 officially and I don't have a plan to support other variants of RPis. I think everyone who can handle RPis can easily port to each variants of RPis.

nherriot:
yicho - are you saying that 'everyone' who can handle RPis know how to port a Javascript framework from one Linux variant to another? (That's more than 10 million people!!!) I personally think that is a skilled job and certainly not one that even experienced Java Scrip developers want to be bothered with. I hope that in the near future this is something the community can find here at the IoTJS home! :-)

rzr:
To be tracked at:
https://bugs.debian.org/889824

PR #1400, opened 2018-01-08
iotjs: Add full.profile

rzr:
More features might be added later.
This is also helpful for raspbian.

Thanks-to: @nherriot <Nicholas.Herriot@gmail.com>
Bug: https://github.com/Samsung/iotjs/issues/1389


glistening:
+1 for @LaszloLango. It is not clear what is the full profile since there is no platform that support all modules currently, and there are profiles for each platform under `test/profiles/` that enables maximum modules for each platform.

rzr:
OK, for feedback now I think this item can be closed, 

BTW I am wondering if there is any attempt for https module to support tizenrt (with a libcurl port?)

To be tracked at:
https://developer.artik.io/forums/t/where-do-the-header-files-exist/3064

yichoi:
I will keep this closing until you guys understand developments guides of IoT.js

Issue #1367, opened 2017-12-13
Remove JavaScript layer and rework API in System I/O module


hs0225:
To simplify the code like ADC, remove the JavaScript layer that only performs type checking.

Currently, some APIs are confusing to us. For example, the close function of GPIO works asynchronously, but, close function of I2C is performed in synchronization. This needs to be fixed.
All API provide sync and async functions depending on the presence of the callback function.
The following APIs are changed.

Before,
```js
var Gpio = require('gpio');
var gpio = new Gpio();

// only must open asynchronously
var gpio1 = gpio.open({pin: 1}, function(err) {
    var value = gpio1.readSync(); // sync
    gpio1.read(function(err, value) { // async
    });
});
```

After,
```js
var Gpio = require('gpio');

var gpio1 = new Gpio({pin: 1}); // sync
var gpio2 = new Gpio({pin: 2}, function(err) {  // async
   var value = gpio1.read(); // sync
   gpio2.read(function(err, value) { // async
      
   });
});
```
The modules below are involved in this task.

- [x] ADC
- [x] GPIO
- [x] I2C
- [x] PWM
- [x] SPI
- [x] UART

galpeter:
I've been thinking a while now about the optional callback parameter and it's effect to make the call async. I think it can be a bit confusing that a given call is sync or not simply on the number of arguments. In my opinion we should provide a sync and async methods just like our fs api to reduce the possible programming errors.

hs0225:
Considering the opinion of many people, I would like to confirm the following.
```js
var Gpio = require('gpio');

// async API
gpio.open({pin: 1}, function(err, gpioPin) {
    gpioPin.read(function(err, value) {
    });
});

// sync API
var gpioPin = gpio.openSync({pin: 1});
var value = gpioPin.readSync();
```

hs0225:
Done.

PR #1405, opened 2018-01-10
Fix function names in I2C module.

LaszloLango:

### DIFF
--- a/src/modules/iotjs_module_i2c.c
+++ b/src/modules/iotjs_module_i2c.c
@@ -20,18 +20,18 @@
 
 IOTJS_DEFINE_NATIVE_HANDLE_INFO_THIS_MODULE(i2c);
 
-static iotjs_i2c_t* iotjs_i2c_create(const jerry_value_t ji2c) {
+static iotjs_i2c_t* i2c_create(const jerry_value_t ji2c) {
   iotjs_i2c_t* i2c = IOTJS_ALLOC(iotjs_i2c_t);
   IOTJS_VALIDATED_STRUCT_CONSTRUCTOR(iotjs_i2c_t, i2c);
-  i2c_create_platform_data(i2c);
+  iotjs_i2c_create_platform_data(i2c);
   _this->jobject = ji2c;
   jerry_set_object_native_pointer(ji2c, i2c, &this_module_native_info);
 
   return i2c;
 }
 
-static iotjs_i2c_reqwrap_t* iotjs_i2c_reqwrap_create(
-    const jerry_value_t jcallback, iotjs_i2c_t* i2c, I2cOp op) {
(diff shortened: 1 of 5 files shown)


src/modules/iotjs_module_i2c.c
```diff
@@ -79,11 +79,11 @@ iotjs_i2c_t* iotjs_i2c_instance_from_reqwrap(iotjs_i2c_reqwrap_t* i2c_reqwrap) {
 
 static void iotjs_i2c_destroy(iotjs_i2c_t* i2c) {
   IOTJS_VALIDATED_STRUCT_DESTRUCTOR(iotjs_i2c_t, i2c);
-  i2c_destroy_platform_data(_this->platform_data);
+  iotjs_i2c_destroy_platform_data(_this->platform_data);
   IOTJS_RELEASE(i2c);
 }
 
-static void i2c_worker(uv_work_t* work_req) {
+static void iotjs_i2c_worker(uv_work_t* work_req) {
```
glistening: @LaszloLango We used `i2c_worker` here instead of `iotjs_i2c_worker` because it is a static function, and prefix `iotjs_` is used for global names.

LaszloLango: There are some inconsistency here. Other static functions that use the `iotjs_` prefix in this file.
```
static iotjs_i2c_t* iotjs_i2c_create
static iotjs_i2c_reqwrap_t* iotjs_i2c_reqwrap_create
static void iotjs_i2c_reqwrap_destroy
static void iotjs_i2c_destroy
```

Other modules also uses `iotjs_` prefix for those functions (and also for `_worker` functions).
```
static void iotjs_gpio_write_worker
static void iotjs_gpio_read_worker
static void iotjs_gpio_close_worker
static void iotjs_gpio_after_worker
static void iotjs_uart_write_worker
static void iotjs_uart_close_worker
static void iotjs_uart_after_worker
static void iotjs_spi_transfer_worker
static void iotjs_spi_close_worker
static void iotjs_pwm_common_worker
static void iotjs_pwm_after_worker
static void iotjs_adc_read_worker
static void iotjs_adc_close_worker
```

Which one should we use here and in other modules? I think we should use the same coding conversion and we have to touch these function while we are working on #1367, so it is a good opportunity to do it now. :)

glistening: @LaszloLango I found the inconsistencies also. I think it is good opportunity to change. Could you use `iotjs_` for global names only during wokring on #1367? If you find other inconsistencies, I would be happy for removing inconsistent style.

LaszloLango:
@glistening I've updated the PR. Please check.

yichoi:
LGTM

PR #1408, opened 2018-01-11
Fix the measure_coverage.sh script

rtakacs:
Removed the old `--no-check-test` option from the build flags.


robertsipka:
Could you please also remove from other places (e.g. Build-Script.md)

rtakacs:
@robertsipka Nice catch! Of course :)

rtakacs:
@robertsipka I've updated the files to use the new `run-test` option instead of the old `no-check-test` one.

docs/build/Build-Script.md
```diff
@@ -235,6 +227,14 @@ To build for nuttx os, nuttx home directory must be given.
 ./tools/build.py --target-os=nuttx --target-arch=arm --target-board=stm32f4dis --nuttx-home="..."
 ```
 
+--
+#### `--run-test`
+With given this option, unit test checking will be performed.
```
glistening: @rtakacs Although it is not a part of this PR, what about "run unit tests after build"? I would remove all occurrences of `With given this option` in explanation for each option. Without this phrases, it is still clear the each sentence says about each option.

PR #1409, opened 2018-01-11
module: add __filename and __dirname

yorkie:
Add the `__filename` & `__dirname` to module file context.

### DIFF
--- a/src/js/module.js
+++ b/src/js/module.js
@@ -218,8 +218,10 @@ iotjs_module_t.loadRemote = function(filename, source) {
 
 
 iotjs_module_t.prototype.compile = function(filename, source) {
-    var fn = process.compile(filename, source);
-    fn.call(this.exports, this.exports, this.require.bind(this), this);
+  var dirname = filename.replace(/\/[^\/]*$/, '');
+  var fn = process.compile(filename, source);
+  fn.call(this.exports, this.exports, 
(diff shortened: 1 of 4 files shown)


zherczeg:
Is this a non-standard extension?

yorkie:
Nope, it's standard in Node.js, see https://nodejs.org/dist/latest-v9.x/docs/api/globals.html#globals_dirname

glistening:
@yorkie I am not sure whether we should provide them in iotjs just because they exists in node.js. I would like to know the specific reason: Are you trying to reuse node.js module? Are you writing your own module that requires `__filename` or`__dirname`?

yorkie:
@glistening there are Node.js modules which uses `__dirname`, and it should could be used directly in IoT.js!

LaszloLango:
@yorkie what modules use `__dirname`? We have to consider the runtime memory usage. This change will increase the memory usage for every module, not just for those which use it. A few bytes does not matter in case of Nodejs. Our usecase is different. If some module really need this feature then we might consider a different approach even if it could be slower.

yorkie:
In Node.js, developer uses `__dirname` in usual to search filename based on the current module file, could we finish the same function in IoT.js?

yichoi:
Closing since I think above answer of @glistening quite enough for you.

PR #1411, opened 2018-01-12
Fix portable build warning in magic strings

galpeter:
Usage of 'defined' in macro is not portable as reported
by the GCC compiler.

Simplifying the module checks to resolve the warning. This
is done by removing the 'defined' part. This works
because all ENABLE_MODULE_<module_name> defines are
set to at least a 0.


yichoi:
LGTM

PR #1417, opened 2018-01-18
Sandbox/nherriot/devel

nherriot:
Adding demo scripts which:
1) Test out each light individually.
2) Test out each colour (red, green, blue) at 10% 50% and 100%
3) Test out switching all LED's off.
4) Test out white at 10% 50% 100%

Adding correct latch sequences to the driver.
Tidy up code, fix comments, add links and formatting changes.


glistening:
@nherriot

First, we don't have plan to support and maintain for making debian (raspbian) package for now.

Second, this PR needs update:
- `debian` should not be in root directory. I recommend you to find a way to put it under `config/debian` like tizen port does.
-  `demo/*' would be better to put under `sample/*'. We want to simplify the fonder structure.
- `debian/*' should not be git-ignore'd,  it must be under source control.
- merge multiple commits into one commit

yichoi:
Closing with above reasons

PR #1418, opened 2018-01-18
Remove Validated Struct from the buffer module.

knightburton:


yichoi:
LGTM

PR #1424, opened 2018-01-22
Add printing out debug logs using external output

daeyeon:

### DIFF
--- a/tools/build.py
+++ b/tools/build.py
@@ -224,6 +224,9 @@ def adjust_options(options):
     options.host_tuple = '%s-%s' % (platform.arch(), platform.os())
     options.target_tuple = '%s-%s' % (options.target_arch, options.target_os)
 
+    # Normalize the path of build directory.
+    options.builddir = fs.normpath(options.builddir)
+
     options.build_root = fs.join(path.PROJECT_ROOT,
                                  options.builddir,
                                  options.target_tuple,
@@ -444,9 +447,9 @@ if __name__ == '__main__':
         print("\n%sTo run tests use '--run-test' "
               "or one of the folowing commands:%s"
               % (ex._TERM_BLUE, ex._TERM_EMPTY))
-        print("\n    tools/testrunner.py build/%s/%s/bin/iotjs"
-              % (options.target_tuple, options.buildtype))
-        print("OR\n    %s(deprecated)%s build/%s/%s/bin/iotjs "
+        print("\n    tools/testrunner.py %s/%s/%s/bin/iotjs"
+              % (options.builddir, options.target_tuple, options.buildtype))
+        print("OR\n    %s(deprecated)%s %s/%s/%s/bin/iotjs "
               "tools/check_test.js\n"
-              % (ex._TERM_RED, ex._TERM_EMPTY, options.target_tuple,
-                 options.buildtype))
+              % (ex._TERM_RED, ex._TERM_EMPTY, options.builddir,
+                 options.target_tuple, options.buildtype))


glistening:
@daeyeon What about adding your implementation for debug log in existing  `src/iotjs_debuglog.c` and call the function in console module?

src/iotjs_env.c
```diff
@@ -152,6 +153,15 @@ uv_loop_t* iotjs_environment_loop(const iotjs_environment_t* env) {
   return _this->loop;
 }
 
+void iotjs_environment_set_dbgout(iotjs_environment_t* env, dbgout_t dbgout) {
+  IOTJS_VALIDATED_STRUCT_METHOD(iotjs_environment_t, env);
```
glistening: @daeyeon Currently we are removing `IOTJS_VALIDATED_STRUCT_XXX`. We can remove it.

src/iotjs_env.c
```diff
@@ -152,6 +153,15 @@ uv_loop_t* iotjs_environment_loop(const iotjs_environment_t* env) {
...
+  _this->dbgout = dbgout;
+}
+
+dbgout_t iotjs_environment_dbgout(const iotjs_environment_t* env) {
+  const IOTJS_VALIDATED_STRUCT_METHOD(iotjs_environment_t, env);
```
glistening: @daeyeon Ditto

src/iotjs_env.c
glistening: @daeyeon I don't want to mix 'validated struct' removal patch with your PR. So I pushed #1432.

daeyeon:
@glistening `iotjs_debuglog.c` is updated.

src/iotjs_env.c
daeyeon: I agree. It should be handled in another pr.

src/iotjs_debuglog.h
```diff
@@ -36,6 +37,8 @@ extern const char* iotjs_debug_prefix[4];
       fprintf(iotjs_log_stream, __VA_ARGS__);                       \
       fprintf(iotjs_log_stream, "\n");                              \
       fflush(iotjs_log_stream);                                     \
+      if(iotjs_debug_extern_out)                                    \
```
glistening: @daeyeon Please add a space between if and opening parenthesis.

src/modules/iotjs_module_console.c
```diff
@@ -15,6 +15,10 @@
 
 #include "iotjs_def.h"
 
+#define DBG_BUF_SIZE 1024
```
glistening: @daeyeon Too big.

src/iotjs_debuglog.h
daeyeon: fixed.

src/modules/iotjs_module_console.c
```diff
@@ -25,13 +29,26 @@ static jerry_value_t Print(const jerry_value_t* jargv,
...
-    } else {
-      fprintf(out_fd, "\\u0000");
+  iotjs_environment_t* env = iotjs_environment_get();
+  dbgout_t dbgout = iotjs_environment_dbgout(env);
+
+  if (dbgout) {
+    char tmpbuf[DBG_BUF_SIZE];
+    int level = (out_fd == stdout) ? DBG_LV_NORMAL : DBG_LV_ERR;
+
+    strncpy(tmpbuf, str, DBG_BUF_SIZE - 1);
```
glistening: @daeyeon I think we don't need `strncpy`. Therefore we can get rid of `tmpbuf`.

src/iotjs_debuglog.h
```diff
@@ -36,6 +37,8 @@ extern const char* iotjs_debug_prefix[4];
       fprintf(iotjs_log_stream, __VA_ARGS__);                       \
       fprintf(iotjs_log_stream, "\n");                              \
       fflush(iotjs_log_stream);                                     \
+      if (iotjs_debug_extern_out)                                   \
+        iotjs_debug_extern_out(lvl, __VA_ARGS__);                   \
```
glistening: @daeyeon What is the difference between `iotjs_log_stream` and `iotjs_debug_extern_out`? 

First, I cannot understand why do we need two separate logging systems.

Second, implementation are scattered in inconsistent way. `iotjs_log_stream` is defined as global variable in iotjs_debuglog.c and `iotjs_debug_extern_out` is defined in iotjs_env as a member. It makes me understand and follow up the code.

PR #1431, opened 2018-01-22
Module ADC rework

bzsolt:
Related issue: #1367

### DIFF
--- a/src/js/adc.js
+++ b/src/js/adc.js
@@ -13,14 +13,16 @@
  * limitations under the License.
  */
 
-function Adc() {
-  if (!(this instanceof Adc)) {
-    return new Adc();
-  }
-}
-
-Adc.prototype.open = function(configuration, callback) {
-  return new native.Adc(configuration, callback);
+var adc = {
+  open: function(config, callback) {
+    var adcPin = new native(config, function(err) {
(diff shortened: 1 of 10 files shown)


src/js/adc.js
```diff
@@ -13,14 +13,16 @@
...
-    return new Adc();
-  }
-}
-
-Adc.prototype.open = function(configuration, callback) {
-  return new native.Adc(configuration, callback);
+var adc = {
+  open: function(config, callback) {
+    var adcPin = new native(config, function(err) {
+      callback(err, adcPin);
```
glistening: @bzsolt What happens when user calls `open` without `callback` parameter?

bzsolt: It will throw a TypeError.

glistening: @bzsolt Although the behavior is changed from previous one, it seems reasonable. Also, I confirmed node also throws `TypeError`.

```
$ node -e "require('fs').open('/etc/hosts','r')"
fs.js:131
    throw new TypeError('"callback" argument must be a function');
    ^
```

yichoi:
@bzsolt rebase please

bzsolt:
@yichoi The PR is rebased.

PR #1432, opened 2018-01-23
Remove validated struct form from iotjs_env

glistening:

### DIFF
--- a/src/iotjs_env.c
+++ b/src/iotjs_env.c
@@ -44,26 +44,23 @@ void iotjs_environment_release() {
   if (!initialized)
     return;
 
-  IOTJS_VALIDATED_STRUCT_DESTRUCTOR(iotjs_environment_t,
-                                    iotjs_environment_get());
-  if (_this->config.debugger)
-    iotjs_buffer_release((char*)(_this->config.debugger));
-  if (_this->argv)
-    iotjs_buffer_release((char*)_this->argv);
+  iotjs_environment_t* env = iotjs_environment_get();
+  if (env->config.debugger)
+    iotjs_buffer_release((char*)(env->config.debugger));
+  if (env->argv)
+    iotjs_buffer_release((char*)env->argv);
   initialized = false;
 }
 
 
 static void initialize(iotjs_environment_t* env) {
-  IOTJS_VALIDATED_STRUCT_CONSTRUCTOR(iotjs_environment_t, env);
-
-  _this->argc = 0;
-  _this->argv = NULL;
-  _this->loop = NULL;
-  _this->state = kInitializing;
-  _this->config.memstat = false;
-  _this->config.show_opcode = false;
-  _this->config.debugger = NULL;
+  env->argc = 0;
+  env->argv = NULL;
+  env->loop = NULL;
+  env->state = kInitializing;
+  env->config.memstat = false;
+  env->config.show_opcode = false;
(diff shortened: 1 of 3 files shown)


daeyeon:
Please wait that #1424 is landed.

glistening:
@daeyeon I pushed it because I don't want to see new 'validated struct ' that is under removal. But if you would like to land your PR first, I will wait.

src/iotjs_env.c
```diff
@@ -118,74 +113,67 @@ bool iotjs_environment_parse_command_line_arguments(iotjs_environment_t* env,
...
   // 2) Standard C guarantees that strings pointed by the argv array shall
   //    retain between program startup and program termination
   while (i < argc)
-    _this->argv[_this->argc++] = argv[i++];
+    env->argv[env->argc++] = argv[i++];
 
   return true;
 }
 
 uint32_t iotjs_environment_argc(const iotjs_environment_t* env) {
```
LaszloLango: This function is not needed after the PR. (The same is true for `iotjs_environment_argv`, `iotjs_environment_loop`, `iotjs_environment_set_loop`, `iotjs_environment_is_exiting` and `iotjs_environment_config`).

glistening: @LaszloLango I don't think this getter/setter is introduced because of `IOTJS_VALIDATED_XXX`. When iotjs_env was implemented in C++, these members were `private`. The getter/setter was existed. The author don't want to access these members directly. See https://github.com/Samsung/iotjs/commit/dd65f4a5aeeea85063a2578b3424c28fcc5093e5#diff-69640a26fee27c1633d7c73f55cccdc0L67. However, you think it would be good to remove them, I will do in this PR.

glistening: @LaszloLango I found it is not the problem of this PR only. For example, do you want to remove [iotjs_string_size](https://github.com/Samsung/iotjs/pull/1445/files#diff-b5f407736578fed92b54ecf056559f99R117) also? There are several functions that are not needed after removing `IOTJS_VALIDATED_xxx`. For consistency, we must remove all of them and change all the usages. I think it is better to decide it later with other's opinions.

LaszloLango: > I don't think these getter/setter are introduced because of IOTJS_VALIDATED_XXX.
I think that was the only reason (checking every "unsafe" access).

> Do you want to remove iotjs_string_size also?
Yes, I do.

> There are several functions that are not needed after removing IOTJS_VALIDATED_xxx
IMHO, it is better to remove all of them if it won't make hurt the readability and maintenance of code. We can reduce the binary size if we reduce the number of such small function. 

> I think it is better to decide it later with other's opinions.
Good idea. Please raise an issue for the topic.

glistening: @LaszloLango 

> I think that was the only reason (checking every "unsafe" access).
Could you explain why do you think so? I told you why I don't think so with previous C++ source.

> Yes, I do
Do you want to remove all simple getter/setter? If so:

First, the issue is not relevant to this PR. This PR is about removing IOTJS_VALIDATED_xxx, not removing simple getter/setter that existed from old C++ source.

Second, could you raise an issue by yourself with the benefit? What benefit do we get from this change? I could not find out any reason to do so.

Third, what about adding requesting change on #1445 for same reason?

yichoi:
LGTM

PR #1436, opened 2018-01-23
Add blinkt module.

ziransun:
[1] Blinkt is a Eight super-bright RGB LED indicator can be mounted
    on rpi 3/2/B+/A+/Zero/Zero w. See https://thepihut.com/products/blinkt
    for more hardware information.
    [2] The module uses existing GPIO interfaces to access data line
    and clock line.
    [3] Blinkt module provides interfaces for accessing the eight pixel
    lights.

IoT.js-DCO-1.0-Signed-off-by: Ziran Sun <ziran.sun@samsung.com>

### DIFF
--- a/src/js/blinkt.js
+++ b/src/js/blinkt.js
@@ -1,0 +1,200 @@
+/* Copyright 2015-present Samsung Electronics Co., Ltd. and other contributors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+
+"use strict";
+
+var Gpio = require('gpio');
+var gpio = new Gpio();
+
+/*
+ Define the DATA pin and the CLK pin for the board.
+ This targets the Pimoroni Blinkt Board.
+*/
+var DAT = 23;
+var CLK = 24;
+var Blinkt;
+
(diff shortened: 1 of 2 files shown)


glistening:
@ziransun Till now, we don't have a specific H/W module as a module. We have functionality modules. (GPIO, I2C, HTTP, ...). Therefore we may need to hear opinion on this PR. If you would like to register blinkt as a module, we need an API document and an example.

yichoi:
Thanks for your contribution @ziransun however I thinks It's not fundamental APIs for general H/W.

LaszloLango:
IMHO we should create a new repository for such modules (eg.: `iotjs-extensions`). IoT.js supports external modules, so no need to put such things into the `src` directory. However modules like this could be useful for somebody.

yichoi:
@LaszloLango Sounds reasonable. I will try by myself to create a new repository

yichoi:
@LaszloLango Sounds reasonable, however we could not guarantee the compatibility such modules all the time. It requires additional resources. doesn't it ?

LaszloLango:
Good question. It might work if the additional resources would be added as submodules with a fixed hash. Maybe we should raise an issue and discuss this with the community.

yichoi:
@LaszloLango why not

ziransun:
Thanks all for your comments. Discussing this in the community sounds good. Please keep me informed and I'd be very interested in joining the discussions.

rzr:
+1 on extra git repo,
I am wondering, are there any plans for installing 3rd party modules (even in compiled form) from iotjs, or should they just need to be copied at the right place (eventually packaged as rpm or deb if we don't support npm or something similar).

nherriot:
It would be better for a developer to go to the one place to find all the lib's and extensions they need. Many of the other communities do similar things. So for example, the blinkt library was first developed and published on github - it was then grabbed and included into the nodejs npm portal.

In this instance we have a contributor directly offering a lib for a very popular device directly to the portal and it getting rejected. That sounds odd to me?

Look here:
https://cylonjs.com/
They already support quite a lot of devices!

LaszloLango:
@nherriot thanks for joining into this conversation, but please write your comments on the recently opened issue for the topic: #1441

PR #1438, opened 2018-01-23
Add curl into link libraries for iotjs build.

ziransun:
curl library is not automatically added while building
iotjs for rpi zero natively. This library is required
for http module.

IoT.js-DCO-1.0-Signed-off-by:Ziran Sun <ziran.sun@samsung.com>

### DIFF
--- a/cmake/iotjs.cmake
+++ b/cmake/iotjs.cmake
@@ -427,6 +427,7 @@ target_include_directories(${TARGET_LIB_IOTJS} PRIVATE ${IOTJS_INCLUDE_DIRS})
 target_link_libraries(${TARGET_LIB_IOTJS}
   ${JERRY_LIBS}
   ${TUV_LIBS}
+  curl
   libhttp-parser
   ${MBEDTLS_LIBS}
   ${EXTERNAL_LIBS}


glistening:
@ziransun curl is not linked for all configuration. It is linked when it is necessary. (currently when https is enabled). Please see https://github.com/Samsung/iotjs/blob/master/src/modules.json#L197.

ziransun:
@glistening With https module enabled in build, curl doesn't seem linked in for rpi zero w native build. For other systems I tried so far it is linked in as you described.

glistening:
@ziransun Do you mean your are building iotjs on raspberry pi zero target (i.e. not cross compilation)? Could you give us your build environment (os, compiler, build command), and logs?

rzr:
I added https to full profile (#1400) used to build deb package, 

I think it worked without changing the cmake file.

For creating raspbian deb package, you can double check on your host by runninng:

      docker build https://github.com/tizenteam/iotjs.git

glistening:
@rzr We don't support debian packaging and RPI zero officially. I cannot accept the global change for unsupported configuration.

About full.profile, please see @LaszloLango's comment https://github.com/Samsung/iotjs/pull/1400#pullrequestreview-87471935 and mine https://github.com/Samsung/iotjs/pull/1400#issuecomment-356799295.

rzr:
OK, for feedback now I think this item can be closed, 

BTW I am wondering if there is any attempt for https module to support tizenrt (with a libcurl port?)

To be tracked at:
https://developer.artik.io/forums/t/where-do-the-header-files-exist/3064

PR #1439, opened 2018-01-24
Add printing out debug logs using external output

daeyeon:


daeyeon:
I made a different approach discussed from https://github.com/Samsung/iotjs/pull/1424. In the context, we assume `iotjs_debuglog.c` will be renamed to `iotjs_logger.c` in another commit, which is used both release and debug mode. A `console output` in `iotjs_logger` is accessible with iotjs modules and logger itself directly through a global variable.

PR #1443, opened 2018-01-24
Fixing remote source sending

DanielBallaSZTE:
Rework remote module loading, so that modules get put into `remoteCache` with their sources, which get compiled later, if `require(modulename')` is called.
That way remoteCache has its own purpose, not only storing modules just as cache does.

### DIFF
--- a/src/js/module.js
+++ b/src/js/module.js
@@ -28,6 +28,7 @@ module.exports = iotjs_module_t;
 
 
 iotjs_module_t.cache = {};
+// Cache to store not yet compiled remote modules
 iotjs_module_t.remoteCache = {};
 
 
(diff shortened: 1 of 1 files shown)


src/js/module.js
```diff
@@ -173,6 +173,10 @@ iotjs_module_t.load = function(id, parent) {
   if (process.builtin_modules[id]) {
     return Native.require(id);
   }
+  if (iotjs_module_t.remoteCache[id]) {
```
glistening: @DanielBallaSZTE If someone try to load module A and A happens to be in remoteCache, then it will use remote cache result. I don't think it is correct.

DanielBallaSZTE: @glistening I think a module is a module, no matter which cache it belongs to.
Remote source sending is there, to send files remotely to the client. It's basically the same as local loading a module. Also, `native` modules won't be overwritten, since remote loading happens after them. In my opinion it is a hundred percent correct to have it like this.

Anyways, I don't see a use case, where the user would like to `require('modulename.js')` which happens to be somewhere on the local storage, and then sends a file with the same name remotely through the debugger client.

glistening: @DanielBallaSZTE If so, why do we need `remoteCache`?

DanielBallaSZTE: @glistening I've reworked the patch a bit, giving some more functionality to `remoteCache`. This is something I wanted to do later, however you were right pointing out that `remoteCache` was not needed as a different module cache. However now it stores only the source codes, and these get compiled later, and put into simple cache if `require` is called. I'm open to any suggestions :)

yichoi:
LGTM

PR #1449, opened 2018-01-25
Revert "Remove Validated struct from the https module. (#1429)"

rzr:
This reverts commit 24388e4391f2a872468efd989b3cebb4d1a5887d.

This commit should not have been merged because it does not build:

"_thttps_datahi" is a an existing symbol (replace misake?)

(Maybe the full.profile would have help to detect this).


glistening:
@rzr Thank you. We can find at least two bugs from your report.

We check the build integrity with Travis CI by building maximum possible modules using `test/profiles/*.profile`. I found HTTPS is missing. We will investigate it.

Could you fix the typo instead of reverting? Perhaps you already fixed this problem.

glistening:
@rzr I fixed build failure. Please see #1450.

rzr:
Thanks @glistening for #1450 
About full features check:
https://github.com/Samsung/iotjs/pull/1400

PR #1456, opened 2018-01-30
Add tizen gbs config file

haesik:
.gbs.conf file will be copied in the root folder when we contribute source code to the review.tizen.org


yichoi:
LGTM

PR #1460, opened 2018-02-01
Merge code duplication to reduce code size of system I/O modules

LaszloLango:
Binary size is reduced by ~1 Kbyte when all six modules are enabled (ADC, I2C, GPIO, PWM, SPI, UART)
The code maintainability is also improved by the elimination of code duplication.


yichoi:
LGTM

PR #1469, opened 2018-02-05
Improve defines in 'iotjs_binding.h' to generate better code.

LaszloLango:

### DIFF
--- a/src/iotjs_binding.h
+++ b/src/iotjs_binding.h
@@ -108,6 +108,10 @@ jerry_value_t iotjs_jhelper_eval(const char* name, size_t name_len,
                                  const uint8_t* data, size_t size,
                                  bool strict_mode);
 
+/* Note:
+ *      Defines started with underscores should not be used
+ *      outside of this header.
+ */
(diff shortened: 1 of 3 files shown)


LaszloLango:
**RPi2 `.text` sizes**

|  | .text (default) | .text (target) |
| - | ---- | ---- |
|master| 294961 | 372004 |
| branch | 294833 | 371820 |

src/iotjs_binding.h
```diff
@@ -108,6 +108,11 @@ jerry_value_t iotjs_jhelper_eval(const char* name, size_t name_len,
                                  const uint8_t* data, size_t size,
                                  bool strict_mode);
 
+/* Note:
+ *      Defines started with underscores should not be used
+ *      outside of this header.
```
hs0225: How about using `#undef`?

LaszloLango: Good idea. I'd look better with `#undef`. I'll update the patch.

hs0225: Sorry. This case cannot do because it is a recursive definition.

yichoi:
LGTM

PR #1471, opened 2018-02-06
Sample: GPIO control panel

haesik:
Allows to add/remove gpio pins, set/monitor their state and config
(This is the update version of the previous PR #1282.)


yichoi:
LGTM

PR #1473, opened 2018-02-06
Modify the status badge links to point to the firebase storage

robertsipka:


daeyeon:
I've ever seen the current badge is updated according to the test result, but can't remember how it works. Can you explain why this patch is needed? Does this patch mean that not only the current badge in IoT.js but also the one in jerryscript doesn't work?

yichoi:
LGTM

Issue #1380, opened 2017-12-20
Assertion 'jerry_value_is_object(jbuiltin)' failed in iotjs_bufferwrap_from_jbuiltin


renatahodovan:
###### IoT.js version:

```
Checked revision: afe242a
Build command: tools/build.py --buildtype debug
```
###### OS:

```
Ubuntu 17.10
```
###### Test case:

```javascript
var buff5 = new Buffer('a1b2c3');
buff5._builtin.copy(unescape, 4, 2, 6);
```

###### Backtrace:
```
iotjs/src/modules/iotjs_module_buffer.c:63: Assertion 'jerry_value_is_object(jbuiltin)' failed.

[Backtrace]:
#0  iotjs_bufferwrap_from_jbuiltin (jbuiltin=72) at iotjs/src/modules/iotjs_module_buffer.c:63
#1  0x000055555557548f in iotjs_bufferwrap_from_jbuffer (jbuffer=15915) at iotjs/src/modules/iotjs_module_buffer.c:75
#2  0x00005555555764b2 in Copy (jfunc=4875, jthis=15467, jargv=0x7fffffffd030, jargc=4) at iotjs/src/modules/iotjs_module_buffer.c:266
#3  0x00005555555a100e in ecma_op_function_call (func_obj_p=0x55555583ffa8 <jerry_global_heap+4872>, this_arg_value=15467, arguments_list_p=0x7fffffffd030, arguments_list_len=4)
    at iotjs/deps/jerry/jerry-core/ecma/operations/ecma-function-object.c:465
#4  0x00005555555b1b0f in opfunc_call (frame_ctx_p=0x7fffffffd080) at iotjs/deps/jerry/jerry-core/vm/vm.c:425
#5  0x00005555555b6ab0 in vm_execute (frame_ctx_p=0x7fffffffd080, arg_p=0x7fffffffd478, arg_list_len=3) at iotjs/deps/jerry/jerry-core/vm/vm.c:2861
#6  0x00005555555b6d38 in vm_run (bytecode_header_p=0x555555842c10 <jerry_global_heap+16240>, this_binding_value=14163, lex_env_p=0x55555583ecc8 <jerry_global_heap+40>, is_eval_code=false, 
    arg_list_p=0x7fffffffd478, arg_list_len=3) at iotjs/deps/jerry/jerry-core/vm/vm.c:2941
[...]

Issue #1361, opened 2017-12-09
Assertion 'buffer != NULL' failed in iotjs_bufferwrap_from_jbuiltin


renatahodovan:
###### IoT.js version:

```
Checked revision: dbd52a7
Build command: tools/build.py --buildtype debug
```
###### OS:

```
Ubuntu 17.10
```
###### Test case:

```javascript
var buf = new Buffer("a")
buf._builtin.copy({ _builtin: {} } , 1, 1, 1);
```

```
iotjs/src/modules/iotjs_module_buffer.c:66: Assertion 'buffer != NULL' failed.

(gdb) bt
#0  iotjs_bufferwrap_from_jbuiltin (jbuiltin=15403) at iotjs/src/modules/iotjs_module_buffer.c:66
#1  0x0000555555575498 in iotjs_bufferwrap_from_jbuffer (jbuffer=15115) at iotjs/src/modules/iotjs_module_buffer.c:75
#2  0x00005555555764bb in Copy (jfunc=4867, jthis=15507, jargv=0x7fffffffcfc0, jargc=4) at iotjs/src/modules/iotjs_module_buffer.c:266
#3  0x00005555555a1783 in ecma_op_function_call (func_obj_p=0x555555843fa0 <jerry_global_heap+4864>, this_arg_value=15507, arguments_list_p=0x7fffffffcfc0, arguments_list_len=4)
    at iotjs/deps/jerry/jerry-core/ecma/operations/ecma-function-object.c:467
#4  0x00005555555b23f8 in opfunc_call (frame_ctx_p=0x7fffffffd010) at iotjs/deps/jerry/jerry-core/vm/vm.c:425
#5  0x00005555555b750c in vm_execute (frame_ctx_p=0x7fffffffd010, arg_p=0x7fffffffd428, arg_list_len=3) at iotjs/deps/jerry/jerry-core/vm/vm.c:2793
#6  0x00005555555b7794 in vm_run (bytecode_header_p=0x555555846c00 <jerry_global_heap+16224>, this_binding_value=14155, lex_env_p=0x555555842cc8 <jerry_global_heap+40>, is_eval_code=false, 
    arg_list_p=0x7fffffffd428, arg_list_len=3) at iotjs/deps/jerry/jerry-core/vm/vm.c:2873
[...]

PR #1487, opened 2018-02-12
Remove the _builtin property from the Buffer object

galpeter:
By removing the _builtin property from the Buffer object
the internal operations are hidden.

Additional benefit is a bit of code/binary size reduction.


galpeter:
This also fixes #1380 & #1361

src/modules/iotjs_module_buffer.c
```diff
@@ -211,47 +198,42 @@ jerry_value_t iotjs_bufferwrap_create_buffer(size_t len) {
 
 
 JS_FUNCTION(Buffer) {
-  DJS_CHECK_THIS();
   DJS_CHECK_ARGS(2, object, number);
 
-  const jerry_value_t jbuiltin = JS_GET_THIS();
-  const jerry_value_t jbuffer = JS_GET_ARG(0, object);
+  const jerry_value_t jbuiltin = JS_GET_ARG(0, object);
```
daeyeon: I suggest naming `jobject` or `jcontext` instead of `jbuiltin`. `iotjs_bufferwrap_create` also needs the same.

test/run_pass/test_buffer_builtin.js
```diff
@@ -13,99 +13,10 @@
...
-assert.equal(buff3.toString(), 'testabcdefgh');
-
 
-assert.equal(Buffer.byteLength('\u007F'), 1);
-assert.equal(Buffer.byteLength('\u008F'), 2);
-assert.equal(Buffer.byteLength('\u08FF'), 3);
-assert.equal(Buffer.byteLength('abc'), 'abc'.length);
-assert.notEqual(Buffer.byteLength('\u2040'), '\u2040'.length);
+/* A internal workings of a Buffer should not be accessible */
+assert.equal(typeof(buff1._builtin), "undefined");
```
daeyeon: Removing this file is ok to me.

galpeter:
@daeyeon I've updated the PR, thanks!

PR #1478, opened 2018-02-07
Enable the skipped Linux tests

bzsolt:
IoT.js-DCO-1.0-Signed-off-by: Zsolt Borbély zsborbely.u-szeged@partner.samsung.com

PR #1488, opened 2018-02-12
Fix failure detection in the Python testrunner

galpeter:
During the test execution the exit code was not always
checked. This made the testrunner to report PASS status for
a test even if it did not returned zero as the exit code.


LaszloLango:
Good catch. Thanks for the fix. It looks like `test_fs_mkdir_rmdir.js` is still flaky on Travis. Please put it back to the skip list as before #1478. LGTM after that.

galpeter:
@LaszloLango ok I'll add it to the skiplist in this PR, but it should be on it's own. But we need to get this landing before trying to fix the test. Without this PR any other PRs could report falsely that they are ok.

yichoi:
LGTM

PR #1490, opened 2018-02-13
Allow passing output level for testrunners

galpeter:
The option 'run-test' was made to a tri-state option:
* without the '--run-test' option the build script does not run tests
* with the '--run-test=full', testing will be done and all test output will be shown
* with the '--run-test=quiet', testing will be done and only PASS/FAIL will be shown
* and the '--run-test' option will behave as the '--run-test=quiet' option

Also fixes Travis testrun output to show every information.
During Travis testing the output for each test was not visible.

### DIFF
--- a/tools/build.py
+++ b/tools/build.py
@@ -176,8 +176,9 @@ def init_options():
         action='store_true', default=False,
         help='Disable test execution with valgrind after build')
     parser.add_argument('--run-test',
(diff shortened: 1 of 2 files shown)


bzsolt:
LGTM (informally)

yichoi:
LGTM

PR #1491, opened 2018-02-14
Fix to load index[.js] as a default entry of module

daeyeon:
This patch gets `module` to work properly for following cases:
a) try loading index[.js] as default even if `package.json` doesn't exist.
b) filter `undefined` as the given param in `tryPath`.

### DIFF
--- a/src/js/module.js
+++ b/src/js/module.js
@@ -102,22 +102,22 @@ iotjs_module_t.resolveFilepath = function(id, directories) {
 
     // 3. package path id/
     var jsonpath = modulePath + '/package.json';
-    filepath = iotjs_module_t.tryPath(jsonpath);
-    if (filepath) {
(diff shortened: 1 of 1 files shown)


src/js/module.js
```diff
@@ -102,22 +102,22 @@ iotjs_module_t.resolveFilepath = function(id, directories) {
 
     // 3. package path id/
     var jsonpath = modulePath + '/package.json';
-    filepath = iotjs_module_t.tryPath(jsonpath);
-    if (filepath) {
+
+    if (jsonpath && iotjs_module_t.tryPath(jsonpath)) {
```
hs0225: `jsonpath` is always true.

daeyeon: Thanks, It's fixed.

Issue #1360, opened 2017-12-08
Assertion '!throws' failed in iotjs_jhelper_call_ok


renatahodovan:
###### IoT.js version:

```
Checked revision: dbd52a7
Build command: tools/build.py --buildtype=debug
```
###### OS:

```
Ubuntu 17.10
```
###### Test case:

```javascript
process.on('exit', function() {});       
setTimeout(function() { Array.prototype.slice += 'B' }, 500);       
```

###### Backtrace:

```
iotjs/src/iotjs_binding.c:330: Assertion '!throws' failed.

[Backtrace]:

(gdb) bt
#0  iotjs_jhelper_call_ok (jfunc=2155, jthis=72, jargs=0x555555841020 <jargs_empty>) at iotjs/src/iotjs_binding.c:330
#1  0x0000555555571d05 in iotjs_process_next_tick () at iotjs/src/iotjs_binding_helper.c:89
#2  0x0000555555571e11 in iotjs_make_callback_with_result (jfunction=15859, jthis=14835, jargs=0x555555841020 <jargs_empty>)
    at iotjs/src/iotjs_binding_helper.c:125
#3  0x0000555555571dac in iotjs_make_callback (jfunction=15859, jthis=14835, jargs=0x555555841020 <jargs_empty>)
    at iotjs/src/iotjs_binding_helper.c:108
#4  0x0000555555581b1c in iotjs_timerwrap_on_timeout (timerwrap=0x555555883790)
    at iotjs/src/modules/iotjs_module_timer.c:92
#5  0x0000555555581796 in TimeoutHandler (handle=0x5555558837e0) at iotjs/src/modules/iotjs_module_timer.c:59
#6  0x00005555555fc78e in uv__run_timers (loop=0x555555841640 <default_loop_struct>)
    at iotjs/deps/libtuv/src/unix/timer.c:203
#7  0x00005555555f22f3 in uv_run (loop=0x555555841640 <default_loop_struct>, mode=UV_RUN_ONCE)
    at iotjs/deps/libtuv/src/unix/core.c:318
#8  0x000055555556f57b in iotjs_start (env=0x555555841440 <current_env>) at iotjs/src/iotjs.c:147
#9  0x000055555556f8f5 in iotjs_entry (argc=2, argv=0x7fffffffdec8) at iotjs/src/iotjs.c:218
#10 0x000055555556f0da in main (argc=2, argv=0x7fffffffdec8) at iotjs/iotjs_linux.c:19
```

[...]

PR #1494, opened 2018-02-15
Check if next tick value has an error

DanielBallaSZTE:
Fixes issue #1360

### DIFF
--- a/src/iotjs_binding_helper.c
+++ b/src/iotjs_binding_helper.c
@@ -86,12 +86,15 @@ bool iotjs_process_next_tick() {
   IOTJS_ASSERT(jerry_value_is_function(jon_next_tick));
 
   jerry_value_t jres =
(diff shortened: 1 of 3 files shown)


src/iotjs_binding_helper.c
```diff
@@ -86,12 +86,15 @@ bool iotjs_process_next_tick() {
...
-                            iotjs_jargs_get_empty());
+      iotjs_jhelper_call(jon_next_tick, jerry_create_undefined(),
+                         iotjs_jargs_get_empty());
 
-  IOTJS_ASSERT(jerry_value_is_boolean(jres));
+  bool ret = false;
+
+  if (!jerry_value_has_error_flag(jres)) {
+    ret = iotjs_jval_as_boolean(jres);
+  }
```
hs0225: You can change the following:
```js
bool ret = jerry_value_has_error_flag(jers);
...
return !ret;
```

PR #1504, opened 2018-02-21
Fix a modified link in the README

robertsipka:
The link to the stm32f4-discovery results has changed after the improvements of the remote testrunner.
Technically, it was just a name extension (stm32->stm32f4dis) in the Firebase to define a more accurate device name.


bzsolt:
LGTM (informally)

yichoi:
LGTM

PR #1506, opened 2018-02-22
Enable build mode on tizen 

haesik:
Default build_mode is release, if you want to build as debug mode
you can use --define='build_mode debug' option in gbs command


yichoi:
LGTM

PR #1508, opened 2018-02-22
Refactor buffer slice method

Achie72:
In iots_module_buffer.c the Slice method contained redundant code, which was moved out into it's seperate function now called `bufferslice_idx_normalizer`. The lenght was asked multiple times by function, so that was moved to the beginning, and saved in it's own variable.

### DIFF
--- a/src/modules/iotjs_module_buffer.c
+++ b/src/modules/iotjs_module_buffer.c
@@ -174,6 +174,20 @@ size_t iotjs_bufferwrap_copy(iotjs_bufferwrap_t* bufferwrap, const char* src,
   return iotjs_bufferwrap_copy_internal(bufferwrap, src, 0, len, 0);
 }
 
+static size_t index_normalizer(int64_t index, size_t max_length) {
+  size_t idx;
+  if (index < 0) {
+    if ((size_t)(-index) > max_length) {
+      idx = SIZE_MAX;
+    } else {
+      idx = (size_t)index + max_length;
+    }
+  } else {
+    idx = (size_t)index;
+  }
+
+  return bound_range(idx, 0, max_length);
+}
 
 jerry_value_t iotjs_bufferwrap_create_buffer(size_t len) {
   jerry_value_t jglobal = jerry_get_global_object();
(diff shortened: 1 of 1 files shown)


src/modules/iotjs_module_buffer.c
```diff
@@ -174,6 +174,22 @@ size_t iotjs_bufferwrap_copy(iotjs_bufferwrap_t* bufferwrap, const char* src,
   return iotjs_bufferwrap_copy_internal(bufferwrap, src, 0, len, 0);
 }
 
+static size_t bufferslice_idx_setter(int64_t point, size_t len) {
```
galpeter: Perhaps a better name would be `buffer_index_normalizer` and the parameters could be named `index` and `max_length`

src/modules/iotjs_module_buffer.c
```diff
@@ -174,6 +174,22 @@ size_t iotjs_bufferwrap_copy(iotjs_bufferwrap_t* bufferwrap, const char* src,
...
+    if ((size_t)(-point) > len) {
+      idx = SIZE_MAX;
+    } else {
+      idx = (size_t)point + len;
+    }
+  } else {
+    idx = (size_t)point;
+  }
+
+
```
galpeter: There is no need for the double empty line here, one is enough.

src/modules/iotjs_module_buffer.c
```diff
@@ -174,6 +174,22 @@ size_t iotjs_bufferwrap_copy(iotjs_bufferwrap_t* bufferwrap, const char* src,
...
+  idx = bound_range(idx, 0, len);
+  return idx;
```
galpeter: Just directly return with the bound_range value: `return bound_range(idx, 0, len);`

src/modules/iotjs_module_buffer.c
```diff
@@ -348,31 +364,9 @@ JS_FUNCTION(Slice) {
...
-      end_idx = (size_t)end + len;
-    }
-  } else {
-    end_idx = (size_t)end;
-  }
-  end_idx = bound_range(end_idx, 0, iotjs_bufferwrap_length(buffer_wrap));
-
+  size_t len = iotjs_bufferwrap_length(buffer_wrap);
+  start_idx = bufferslice_idx_setter(start, len);
+  end_idx = bufferslice_idx_setter(end, len);
```
galpeter: You can add the variable definitions to these lines, no need to have them in a separate line, for example:
```c
size_t start_idx =...;
size_t end_idx = ...;
```

src/modules/iotjs_module_buffer.c
```diff
@@ -174,6 +174,20 @@ size_t iotjs_bufferwrap_copy(iotjs_bufferwrap_t* bufferwrap, const char* src,
   return iotjs_bufferwrap_copy_internal(bufferwrap, src, 0, len, 0);
 }
 
+static size_t index_normalizer(int64_t index, size_t max_lenght) {
```
galpeter: There is one typo: `max_lenght` should be `max_length`

yichoi:
LGTM

PR #1510, opened 2018-02-23
Fix wrong behaviour of iotjs_process_exitcode

### LINKED ISSUE #1507: Assertion 'jerry_value_is_number(jexitcode)' failed in iotjs_process_exitcode
###### IoT.js version:

```
Checked revision: 5da8a00
Build command: tools/build.py --buildtype=debug --jerry-profile=es2015-subset
```
###### OS:

```
Ubuntu 17.10, x86_64
```
###### Test case:

```javascript
var x = require("console");
process.exitCode = x.id_0; 
```

###### Backtrace:

```
iotjs/src/iotjs_binding_helper.c:145: Assertion 'jerry_value_is_number(jexitcode)' failed.

(gdb) bt
#0  iotjs_process_exitcode () at iotjs/src/iotjs_binding_helper.c:145
#1  0x000055555557133d in iotjs_start (env=0x5555558a54c0 <current_env>) at iotjs/src/iotjs.c:148
#2  0x000055555557161c in iotjs_entry (argc=2, argv=0x7fffffffdd48) at iotjs/src/iotjs.c:204
#3  0x0000555555570e1a in main (argc=2, argv=0x7fffffffdd48) at iotjs/src/platform/linux/iotjs_linux.c:19

```

<sup>Found by [Fuzzinator](http://fuzzinator.readthedocs.io/) with [grammarinator](https://github.com/renatahodovan/grammarinator)</sup>

irishair7:
process.exitCode can be overwritten by anything.
when it's not a number type, just let iotjs_process_exitcode() returns default number.

fixes #1507

### DIFF
--- a/src/iotjs_binding_helper.c
+++ b/src/iotjs_binding_helper.c
@@ -142,12 +142,17 @@ int iotjs_process_exitcode() {
 
   jerry_value_t jexitcode =
       iotjs_jval_get_property(process, IOTJS_MAGIC_STRING_EXITCODE);
-  IOTJS_ASSERT(jerry_value_is_number(jexitcode));
-
-  const int exitcode = (int)iotjs_jval_as_number(jexitcode);
+  uint8_t exitcode = 0;
+  jerry_value_t num_val = jerry_value_to_number(jexitcode);
+  if (jerry_value_has_error_flag(num_val)) {
+    exitcode = 1;
+    jerry_value_clear_error_flag(&num_val);
+  } else {
+    exitcode = (uint8_t)iotjs_jval_as_number(num_val);
+  }
+  jerry_release_value(num_val);
   jerry_release_value(jexitcode);
-
-  return exitcode;
+  return (int)exitcode;
 }
 
 
--- a/test/run_pass/issue/issue-1507.js
+++ b/test/run_pass/issue/issue-1507.js
@@ -1,0 +1,17 @@
+/* Copyright 2018-present Samsung Electronics Co., Ltd. and other contributors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+var x = require("console");
+process.exitCode = x.id_0;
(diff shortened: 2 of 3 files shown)


src/iotjs_binding_helper.c
```diff
@@ -142,7 +142,11 @@ int iotjs_process_exitcode() {
 
   jerry_value_t jexitcode =
       iotjs_jval_get_property(process, IOTJS_MAGIC_STRING_EXITCODE);
-  IOTJS_ASSERT(jerry_value_is_number(jexitcode));
+
+  if (!jerry_value_is_number(jexitcode)) {
+    jerry_release_value(jexitcode);
+    jexitcode = jerry_create_number(1);
```
galpeter: I'm not sure this is the correct solution. I've checked the node.js to see what the returned exit code will be in case of the test. In node.js it returned `0`. So I think we should do the same here.

Another Node.js example:
```sh
$ node -e "process.exitCode = {valueOf: function(){return 1; }};"
$ echo $?
1
```

It seems that in node the toInteger operation is applied to the `exitCode` value.
IMHO reasonable thing to adapt in IoT.js, just make the engine more node-compatible.

test/run_fail/test-issue-1507.js
```diff
@@ -0,0 +1,17 @@
+/* Copyright 2017-present Samsung Electronics Co., Ltd. and other contributors
```
galpeter: The date should be `2018`

src/iotjs_binding_helper.c
irishair7: Good opinion, I'll update the commit by your comment.
thank you.

galpeter:
The main question here is that: do we want to follow node.js in this case?
If yes, then your solution will not be compatible with node.js. As mentioned in a previous comment the example: `process.exitCode = {valueOf: function(){return 1; }};` should return an exit code of `1` in node.js, whereas your solution will return `0` in iot.js.

irishair7:
Updated to be node-compatible using toNumber operation.
plz check new commit.

src/iotjs_binding_helper.c
```diff
@@ -142,12 +142,34 @@ int iotjs_process_exitcode() {
 
   jerry_value_t jexitcode =
       iotjs_jval_get_property(process, IOTJS_MAGIC_STRING_EXITCODE);
-  IOTJS_ASSERT(jerry_value_is_number(jexitcode));
+  jerry_type_t val_type = jerry_value_get_type(jexitcode);
```
galpeter: There is no need for this type of checking. Only call the `jerry_value_to_number` and extract the value. It has a quick path for numbers which does the same as you are doing here.
Pseudo code:
```
num_val = jerry_value_to_number(jexitcode)
if (have_error(num_val) {
  exitcode = 1;
  remove_error_flag(num_val);
} else {
  exitcode = iotjs_jval_as_number(num_val);
}
jerry_release_value(num_val);
```

irishair7: You're right. commit updated again.
Thanks

galpeter:
lgtm

PR #1523, opened 2018-03-06
change IOTJS_RELEASE to safely release the given memory

daeyeon:

### DIFF
--- a/src/iotjs_util.h
+++ b/src/iotjs_util.h
@@ -33,7 +33,10 @@ void iotjs_buffer_release(char* buff);
   (type*)iotjs_buffer_allocate(sizeof(type))
 
 #define IOTJS_RELEASE(ptr) /* Release memory allocated by IOTJS_ALLOC() */ \
-  iotjs_buffer_release((char*)ptr)
+  ({                                                                       \
+    iotjs_buffer_release((char*)ptr);                                      \
+    ptr = NULL;                                                            \
+  })
 
 
 #endif /* IOTJS_UTIL_H */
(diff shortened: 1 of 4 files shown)


daeyeon:
Once accepted, I will apply this to other files too.

src/iotjs_util.h
```diff
@@ -35,5 +35,10 @@ void iotjs_buffer_release(char* buff);
 #define IOTJS_RELEASE(ptr) /* Release memory allocated by IOTJS_ALLOC() */ \
   iotjs_buffer_release((char*)ptr)
 
+#define IOTJS_SAFE_RELEASE(ptr)       \
+  ({                                  \
+    iotjs_buffer_release((char*)ptr); \
+    ptr = NULL;                       \
```
galpeter: Can't we just simply move the `ptr=NULL` part into the buffer release method? Is there any benefit having it as a separate macro?

daeyeon: Yes, we can move it into iotjs_buffer_release as you mentioned. There are two below options, and I prefer to use (b). Are you fine with (b)?

a) use double pointers
```
char *ptr;
iotjs_buffer_release(&ptr); 

void iotjs_buffer_release(char** buffer)
```

b) ptr passing by reference
```
iotjs_buffer_release(ptr); 

void iotjs_buffer_release(char* & buffer)
```

galpeter: IMHO, we can not pass by reference. We are in C and not C++. However now I understand why the new macro was created.

daeyeon: @galpeter You're right. b) isn't possible in C. I forgot that.

I receive another opinion to modify IOTJS_RELEASE instead of adding a new macro. I will change this commit like that.

galpeter: I'm ok with updating the IOTJS_RELEAS macro.

yichoi:
LGTM

PR #1524, opened 2018-03-06
Remove code that has no effect.

irishair7:
Remove code that has no effect.
in iotjs_handlewrap_validate


yichoi:
LGTM

PR #1527, opened 2018-03-07
Support base64 decoding for buffer construction and buffer write.

zherczeg:


src/modules/iotjs_module_buffer.c
```diff
@@ -88,31 +93,122 @@ static size_t bound_range(size_t index, size_t low, size_t upper) {
...
+  }
+
+  return (int32_t)(-1);
+}
+
+
+static size_t base64_decode(char* buf, size_t len, const char* src,
+                            const size_t srcLen) {
+  if (srcLen == 0) {
+    return 0 + 1;
```
daeyeon: It seems weird that the size decoded string is 1 when srcLen is 0. This causes that we need to reduce the returned value like `jerry_create_number(nbytes - 1);`.

zherczeg: The function returns with 0 on error, and len + 1 otherwise. I could use `ssize_t` and use -1 on error, but loosing half of the range of `size_t` for returning an error seems a waste. What do you prefer?

daeyeon: The current one is now fine with me. Thanks.

yichoi:
LGTM

PR #1528, opened 2018-03-07
Add the url module and the tests for it

pmarkee:

### DIFF
--- a/src/js/url.js
+++ b/src/js/url.js
@@ -1,0 +1,256 @@
+/* Copyright 2018-present Samsung Electronics Co., Ltd. and other contributors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+function Url(url) {
+  this.protocol = null;
+  this.auth = null;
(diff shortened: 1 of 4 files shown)


daeyeon:
This is seemingly a preliminary work to implement https module and related modules. IMHO, parsing protocol, host, (auth,) and port from url is necessary for that. However, whether we should consider the others much is a question mark.

It would be better to get this module to have the minimum required functionalities, reduce the size as many as possible, and use it only internally. Can you clarify if there are parts we could reduce?

pmarkee:
Do you mean we only need protocol, auth, host and port? We can definitely do that

daeyeon:
I think that we need to support passing an url string into `http.request` or `https.request`. If we can achieve that through any other way such as using a simple regular expression, then it would be better than this url module.

FYI: https://mathiasbynens.be/demo/url-regex

pmarkee:
I just updated the PR, I made the module much shorter and more basic. I think a regex could get quite complicated and it could be harder to change it later if it needs a change. This module could easily be used to support passing url strings to http.request and https.request IMO.

pmarkee:
The rpi2 binary size with the module is about 2,5 kbyte larger than before

yichoi:
Let's have more discussion about this. I am wondering whether we need to add "internal" module or not.

src/js/url.js
```diff
@@ -0,0 +1,256 @@
...
+      case 60:
+      case 62:
+      case 92:
+      case 94:
+      case 96:
+      case 123:
+      case 124:
+      case 125:
+        // Characters that are never ever allowed in a hostname from RFC 2396
+        throw new TypeError('Invalid URL: prohibited character in the host');
```
daeyeon: `dns` module will validate a passing hostname. Do we need to check if the given the hostname string meets the specification here?

yichoi:
please move this PR to https://github.com/Samsung/iotjs-modules

yichoi:
closing since https://github.com/Samsung/iotjs-modules/pull/1 has landed

PR #1531, opened 2018-03-07
Intoduce dynamic C modules

galpeter:
Initial support of dynamically loadable iotjs C modules

In this change the loading of <name>.iotjs files are added
via the require JS call.

The .iotjs files are special shared objects where
there is a fixed entry point which will be used
by the dynamic module loader in IoT.js

Also added a dynamically loadable IoT.js module template.
With this a dynamically loadable IoT.js module can be 
quickly developed.

The new template contains a simple readme on how
to build and test.

Usage:

./tools/iotjs-create-module.py --template shared <module_name>


daeyeon:
Please squash the commits into a single one, or else make a separate pr.

galpeter:
I've updated the PR (rebased and only kept the first commit). Also did a bit of refactoring: the dynamic loader is now a module which automatically gets enabled on Linux and Tizen systems.

galpeter:
Oh, and I'll update the module writing docs in the next PR.

src/modules.json
```diff
@@ -239,6 +244,14 @@
       "require": ["assert", "events", "stream", "tcp", "util"]
     },
     "process": {
+      "platforms": {
```
LaszloLango: I think this should be moved to the "module" at line 233.

docs/api/IoT.js-API-Module.md
```diff
@@ -42,6 +42,8 @@ For each directory in search paths above:
 - If a directory `id` exists, module system consider the directory as a package:
   - If `id/package.json` contains **main** property, load the file named **main** property.
   - If `id/package.json` exists, but neither the **main** property nor the file named **main** property exist, load `index.js`.
+- Extra step for Linx/Tizen targets:
```
zherczeg: Linux/Tizen

src/iotjs_binding.h
```diff
@@ -216,4 +219,37 @@ jerry_value_t iotjs_jhelper_eval(const char* name, size_t name_len,
 
 jerry_value_t vm_exec_stop_callback(void* user_p);
 
+/**
+ * Dynamic module defintions (.iotjs)
+ */
+
+#define IOTJS_MAKE_VERSION(major, minor, patch)               \
+  (((0x10 & 0xFF) << (3 * 8)) | ((major & 0xFF) << (2 * 8)) | \
+   ((minor & 0xFF) << (1 * 8)) | (patch & 0xFF))
```
zherczeg: Patch? A major/minor should be enough. Well I think even a major is enough with uint32 range.

src/iotjs_binding.h
```diff
@@ -216,4 +219,37 @@ jerry_value_t iotjs_jhelper_eval(const char* name, size_t name_len,
...
+
+typedef struct {
+  uint32_t iotjs_version;
+  uint32_t module_version;
+  uint32_t flags; /* reserverd for future use */
+  char name[255];
+  ModuleInitializer initializer;
+} iotjs_module;
+
+typedef iotjs_module* (*iotjs_module_info_getter)(void);
```
zherczeg: The negative side is that the structure must be static. Btw if the structure is static, the name could be `char*` above.

galpeter: If you use (and you should really use) the macro `IOTJS_MODULE` below, the static struct will be created correctly. I'm not against having the name as a `char*`

galpeter:
@zherczeg I've updated the PR with docs and changed the version to MAJOR,MINOR pair. The `require` API is not in this PR. It would be better to have it as a separate PR.

src/iotjs_binding.h
```diff
@@ -216,4 +219,35 @@ jerry_value_t iotjs_jhelper_eval(const char* name, size_t name_len,
 
 jerry_value_t vm_exec_stop_callback(void* user_p);
 
+/**
+ * Dynamic module defintions (.iotjs)
+ */
+
+#define IOTJS_MAKE_VERSION(major, minor) \
+  (((major & 0xFFFF) << (2 * 8)) | (minor & 0xFFFF))
```
zherczeg: My concern here is how we track when we need to increase version number.

src/iotjs_binding.h
```diff
@@ -216,4 +219,28 @@ jerry_value_t iotjs_jhelper_eval(const char* name, size_t name_len,
 
 jerry_value_t vm_exec_stop_callback(void* user_p);
 
+/**
+ * Dynamic module defintions (.iotjs)
+ */
+#define IOTJS_CURRENT_MODULE_VERSION() ((uint32_t)0x00000001)
```
zherczeg: The () after the macro name is a bit strange to me. And we could use simple decimal numbers, increasing a hex number is brain killing.

galpeter:
@zherczeg Updated based on your comment.

PR #1532, opened 2018-03-08
Change the lint style and remove lint monitoring

daeyeon:
This patch changes the output style to [codeframe](https://eslint.org/docs/user-guide/formatters/#codeframe) which looks similar to the clang output format.


yichoi:
LGTM

PR #1533, opened 2018-03-08
Remove iotjs_jhelper_call_ok

irishair7:
Remove iotjs_jhelper_call_ok
Because it has nothing but the call of iotjs_jhelper_call.
Changed to use iotjs_jhelper_call directly

### DIFF
--- a/src/modules/iotjs_module_buffer.c
+++ b/src/modules/iotjs_module_buffer.c
@@ -199,7 +199,8 @@ jerry_value_t iotjs_bufferwrap_create_buffer(size_t len) {
   iotjs_jargs_append_number(&jargs, len);
 
   jerry_value_t jres =
-      iotjs_jhelper_call_ok(jbuffer, jerry_create_undefined(), &jargs);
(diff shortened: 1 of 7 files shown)


zherczeg:
The patch is basically inlining a simple function. I would prefer to inline the function itself. Easier to maintain. Although I am neutral for the change.

src/modules/iotjs_module_buffer.c
```diff
@@ -199,7 +199,8 @@ jerry_value_t iotjs_bufferwrap_create_buffer(size_t len) {
   iotjs_jargs_append_number(&jargs, len);
 
   jerry_value_t jres =
-      iotjs_jhelper_call_ok(jbuffer, jerry_create_undefined(), &jargs);
+      iotjs_jhelper_call(jbuffer, jerry_create_undefined(), &jargs);
+  IOTJS_ASSERT(!jerry_value_has_error_flag(jres));
```
daeyeon: Why don't we move this assert into `iotjs_jhelper_call`?

PR #1534, opened 2018-03-09
Modify style checker to use regular expression

irishair7:
Modify style checker to use regular expression
tools/check_tidy.py


yichoi:
LGTM

PR #1537, opened 2018-03-12
Support base64 encoding for buffer toString() method.

zherczeg:
Further change: the first argument of toString() must be encoding similar to Node.js.


yichoi:
LGTM

PR #1542, opened 2018-03-14
Modify iotjs_bufferwrap_create_buffer to use jerry_construct_object

pmarkee:


src/modules/iotjs_module_buffer.c
```diff
@@ -293,9 +293,9 @@ jerry_value_t iotjs_bufferwrap_create_buffer(size_t len) {
 
   iotjs_jargs_t jargs = iotjs_jargs_create(1);
   iotjs_jargs_append_number(&jargs, len);
+  jerry_length_t jargc = iotjs_jargs_length(&jargs);
```
galpeter: Afaik, we don't even need the jargs. We can directly create a jerry number and pass it as an argument.

PR #1550, opened 2018-03-22
i2c: Add readBytes utility function

rzr:
Note this function is not critical and can be "injected" outside module.

But if inside, it will help developer to port from node to iotjs,
if API are aligned to npm's i2c module.

An alternative to consider is to create an other module (ie: compat/npm/i2c)
that wraps and fill the gap between iotjs's minimal implementation
and node/NPM one.

IoT.js-DCO-1.0-Signed-off-by: Philippe Coval <philippe.coval@osg.samsung.com>

### DIFF
--- a/src/js/i2c.js
+++ b/src/js/i2c.js
@@ -13,15 +13,25 @@
  * limitations under the License.
  */
 
+function extendI2CBus(i2cBus) {
+  i2cBus.readBytes = function(offset, len, callback) {
+    this.writeSync([offset]);
+    this.read(len, function(err, res) {
+      callback(err, res);
+    });
+  };
+  return i2cBus;
+}
+
 var i2c = {
   open: function(config, callback) {
     var i2cBus = new native(config, function(err) {
-      callback(err, i2cBus);
+      callback(err, extendI2CBus(i2cBus));
     });
     return i2cBus;
   },
   openSync: function(config) {
-    return new native(config);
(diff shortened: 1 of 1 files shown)


rzr:
Just shared for feedback, but I believe it's a similar situation as with http module from @pmarkee 

https://github.com/Samsung/iotjs-modules/pull/1

LaszloLango:
> Signed-off-by message is incorrect. The following line should be at the end of the f73b36f commit's message: 'IoT.js-DCO-1.0-Signed-off-by: Philippe Coval philippe.coval@osg.samsung.com'.

rzr:
Yes I know I noticed this afterwards but I don't think this change worth to be merged, I just shared for feedback, it can be closed without merge as we probably want this in iotjs-modules right ?

cc: @daeyeon ?

daeyeon:
I think that this patch needs to be PRed here. Any experimental or improvement patch related to codes managed in this repository should be requested here.

daeyeon:
I understand your preference. However, we'd like to keep the current api for the following reasons.

- As mentioned, users can have the same through the injection outside module.
- The module mentioned is not convinced as a node de-facto standard.
- The current APIs is also thoughtfully designed for user's usability. It shouldn't be changed simply to support a certain module.
- This change requires the verification process on each board. Document change is also needed. Nothing for that is included in this pr.

PR #1552, opened 2018-03-26
Change iotjs_buffer_release with IOTJS_RELEASE

daeyeon:
This patch is related to #1523.

- `iotjs_string_make_empty` is removed.
- Since `buf` in `iotjs_module_tcp/udp.c` is given as read-only,
  so IOTJS_RELEASE isn't used there.


yichoi:
LGTM

PR #1555, opened 2018-03-27
Improve Buffer.prototype.equals

TamasZakor:
IoT.js-DCO-1.0-Signed-off-by: Tamas Zakor ztamas@inf.u-szeged.hu

### DIFF
--- a/src/js/buffer.js
+++ b/src/js/buffer.js
@@ -152,7 +152,7 @@ Buffer.prototype.equals = function(otherBuffer) {
     throw new TypeError('Bad arguments: buffer.equals(Buffer)');
   }
 
-  return native.compare(this, otherBuffer) == 0;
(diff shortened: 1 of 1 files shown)


src/js/buffer.js
```diff
@@ -152,7 +152,7 @@ Buffer.prototype.equals = function(otherBuffer) {
     throw new TypeError('Bad arguments: buffer.equals(Buffer)');
   }
 
-  return native.compare(this, otherBuffer) == 0;
+  return this.compare(otherBuffer) == 0;
```
zherczeg: Calling an extra function is more costly than using the native function. I don't see much benefit of changing this.

yichoi:
@TamasZakor please update along with @zherczeg's comment

PR #1558, opened 2018-03-28
Clean inefficient codes

daeyeon:
Clean inefficient codes from buffer and tls test.

- remove 'bytes' not used afterward.
- remove duplicate declaration


irishair7:
LGTM

PR #1560, opened 2018-03-28
Use TLS module in HTTPS

LaszloLango:
* Removed the curl dependency
 * Updated the mbedtls configuration
 * Removed the code duplications of HTTP and HTTPS modules
 * Updated the related test files


zherczeg:
Quite good patch

src/js/http.js
```diff
@@ -13,17 +13,43 @@
  * limitations under the License.
  */
 
-var Server = require('http_server').Server;
+var net = require('net');
 var ClientRequest = require('http_client').ClientRequest;
 var HTTPParser = require('httpparser');
+var HTTPServer = require('http_server');
```
zherczeg: "http underscore server" and "httpparser" sounds inconsistent. Although this is legacy so you can keep it.

src/js/http_client.js
```diff
@@ -14,17 +14,15 @@
...
 var HTTPParser = require('httpparser').HTTPParser;
 
-function ClientRequest(options, cb) {
+function ClientRequest(options, cb, socket) {
   OutgoingMessage.call(this);
 
   // get port, host and method.
-  var port = options.port = options.port || 80;
-  var host = options.host = options.hostname || options.host || '127.0.0.1';
+  options.host = options.hostname || options.host || '127.0.0.1';
```
zherczeg: I would put this after the var declarations.

test/profiles/host-darwin.profile
```diff
@@ -1,2 +1,4 @@
 ENABLE_MODULE_IOTJS_BASIC_MODULES
 ENABLE_MODULE_IOTJS_CORE_MODULES
+ENABLE_MODULE_HTTPS
+
```
zherczeg: Extra newline?

test/run_pass/test_net_http_server.js
```diff
@@ -181,7 +183,7 @@ finalReq.end();
 var server2 = http.createServer();
 
 // Create server instance without new keyword.
-var server3 = Server(function(request, response) {});
+// var server3 = Server(function(request, response) {});
```
zherczeg: Why this is commented out?

test/run_pass/test_net_https_server.js
```diff
@@ -0,0 +1,54 @@
...
+
+var assert = require('assert');
+var https = require('https');
+var fs = require('fs');
+
+var server_options = {
+  key: fs.readFileSync('resources/my_key.pem').toString(),
+  cert: fs.readFileSync('resources/my_crt.pem').toString()
+};
+try { throw 1; } catch(e) {}
```
zherczeg: Debug?

src/js/http.js
LaszloLango: Yes, It bothers me too. :) I didn't want to change it in this PR, because it is quite unrelated.

test/run_pass/test_net_https_server.js
LaszloLango: Indeed, it is a leftover. I'll will fix it.

test/run_pass/test_net_http_server.js
LaszloLango: I forgot to uncomment it, but it works. I'll fix it.

LaszloLango:
Just a few notes about this PR:

1) It enables HTTPS and TLS modules by default
  a) so Travis can check the related tests
  b) but it will increase the binary size (also on https://samsung.github.io/iotjs-test-results/)
2) This reveals an AddressSanitizer bug in the mbedtls itself.
  a) We can disable HTTPS by default to make Travis happy, but it can make hard to catch bugs in this module. #1341 can be a good solution to be able to add new job for testing HTTPS and TLS.
  b) We might try to turn off asan checks in mbedtls
  c) Try to create a minimized testcase and report to the ARMmbed team
  d) other idea?

zherczeg:
What about creating another key?

LaszloLango:
@zherczeg unfortunately it does not work. I tried with both `openssl` and `mbedtls`, but every generated key fails.

daeyeon:
@LaszloLango @zherczeg It seems better to include HTTPS tests. We are fine with 2-(b), disabling asan in mbedtls, for now. However, 2-(c) is also needed. Once it's fixed, let's re-enable 2-(b). Could you open an issue for 2-(c) so that we can track it?

docs/api/IoT.js-API-HTTPS.md
```diff
@@ -2,10 +2,13 @@
 
  The following shows Https module APIs available for each platform.
 
-|  | Linux<br/>(Ubuntu) | Raspbian<br/>(Raspberry Pi) | Nuttx<br/>(STM32F4-Discovery) | TizenRT<br/>(Artik053) | Tizen<br/>(Artik 10) |
-| :---: | :---: | :---: | :---: | :---: | :---: |
-| https.request  | X | X | X | X | O |
-| https.get  | X | X | X | X | O |
+ |  | Linux<br/>(Ubuntu) | Raspbian<br/>(Raspberry Pi) | NuttX<br/>(STM32F4-Discovery) | TizenRT<br/>(Artik053) |
+ | :---: | :---: | :---: | :---: | :---: |
+ | http.createServer | O | O | △ ¹ | △ ¹ |
```
daeyeon: https

LaszloLango:
@daeyeon I've reported the issue: https://github.com/ARMmbed/mbedtls/issues/1550
I've updated the PR to ignore the address sanitizer flag in mbedtls.

Since #1341 landed we can test the HTTPS module without enabling it on default build. We can add it to `test/profiles/*.profile` and `testsets-*.json` (e.g: `testsets-host-linux.json`, etc.). So Travis could run the tests, but the default build (`tools/build.py --clean`) won't change. Which way do you prefer?

daeyeon:
@LaszloLango It's right not to include HTTPS on the default build. Please add it to `*.profile` and `testsets.json`. Having each json per each purpose is a question mark for now.

LaszloLango:
@daeyeon I've updated the PR.

src/modules.json
```diff
@@ -195,20 +195,7 @@
...
-      "require": ["buffer", "https_incoming", "stream", "util"]
-    },
-    "https_incoming": {
-      "js_file": "js/https_incoming.js",
-      "require": ["buffer", "stream", "util"]
-    },
-    "https_native": {
-      "native_files": ["modules/iotjs_module_https.c"],
-      "init": "InitHttps"
+      "require": ["https_client", "httpparser", "http_server", "net", "tls"]
```
zherczeg: The `https_client` is deleted. Why do we need it?

LaszloLango: Good catch, it is a typo. It should be `"http_client"` I'll fix it.

PR #1563, opened 2018-03-28
Get string value from buffers

TamasZakor:
Added the ability to pass buffers as parameters and get the string value from them.

IoT.js-DCO-1.0-Signed-off-by: Tamas Zakor ztamas@inf.u-szeged.hu


src/iotjs_binding.c
```diff
@@ -84,6 +84,31 @@ double iotjs_jval_as_number(jerry_value_t jval) {
...
+    IOTJS_ASSERT(check == size);
+    buffer[size] = '\0';
+
+    *getbuffer = iotjs_string_create_with_buffer(buffer, size);
+    jerry_release_value(to_string);
+
+    return true;
+  }
+  return false;
+}
```
zherczeg: Two newlines between functions.

src/iotjs_binding.c
```diff
@@ -84,6 +84,31 @@ double iotjs_jval_as_number(jerry_value_t jval) {
...
+
+    IOTJS_ASSERT(check == size);
+    buffer[size] = '\0';
+
+    *getbuffer = iotjs_string_create_with_buffer(buffer, size);
+    jerry_release_value(to_string);
+
+    return true;
+  }
+  return false;
```
zherczeg: What about strings?

src/iotjs_binding.c
```diff
@@ -84,6 +84,31 @@ double iotjs_jval_as_number(jerry_value_t jval) {
   return jerry_get_number_value(jval);
 }
 
+bool iotjs_jval_get_data(jerry_value_t jval, iotjs_string_t* getbuffer){
+  if(iotjs_jbuffer_is_bufferwrap(jval)){
+    jerry_value_t to_string = jerry_value_to_string(jval);
```
zherczeg: You can get the bytes directly, no need to convert it to string. This is an expensive operations.

src/modules/iotjs_module_buffer.c
```diff
@@ -81,6 +81,20 @@ size_t iotjs_bufferwrap_length(iotjs_bufferwrap_t* bufferwrap) {
...
 }
 
+bool iotjs_jbuffer_is_bufferwrap(const jerry_value_t jbuffer) {
+  if (jerry_value_is_object(jbuffer)) {
+    void *native_p;
+    const jerry_object_native_info_t *type_p;
+    bool has_p = jerry_get_object_native_pointer (jbuffer, &native_p, &type_p);
+
+    if (has_p) {
+      if (type_p == &this_module_native_info) {
```
LaszloLango: Please merge the if statements

src/modules/iotjs_module_buffer.c
```diff
@@ -82,6 +82,20 @@ size_t iotjs_bufferwrap_length(iotjs_bufferwrap_t* bufferwrap) {
 }
 
 
+bool iotjs_jbuffer_is_bufferwrap(const jerry_value_t jbuffer) {
+  if (jerry_value_is_object(jbuffer)) {
+    void *native_p;
+    const jerry_object_native_info_t *type_p;
+    bool has_p = jerry_get_object_native_pointer (jbuffer, &native_p, &type_p);
```
LaszloLango: The name is a bit confusing, because `_p` suffix is used for pointers, but `has_p` is a boolean. What about `has_native_pointer`?

src/iotjs_binding.c
```diff
@@ -85,6 +86,39 @@ double iotjs_jval_as_number(jerry_value_t jval) {
...
+    return true;
+  } else if(iotjs_jbuffer_is_bufferwrap(jval)){
+    iotjs_bufferwrap_t* buffer_wrap = iotjs_bufferwrap_from_jbuffer(jval);
+    char* bufferwrap_buffer = buffer_wrap->buffer;
+    jerry_size_t size = buffer_wrap->length;
+
+    char* buffer = iotjs_buffer_allocate(size + 1);
+    memcpy(buffer,bufferwrap_buffer,size);
+    *getbuffer = iotjs_string_create_with_buffer(buffer, size);
+    return true;
```
zherczeg: Newline usage is a bit caotic. Please follow the same layout for the two parts.

src/modules/iotjs_module_buffer.c
```diff
@@ -82,6 +82,21 @@ size_t iotjs_bufferwrap_length(iotjs_bufferwrap_t* bufferwrap) {
 }
 
 
+bool iotjs_jbuffer_is_bufferwrap(const jerry_value_t jbuffer) {
```
zherczeg: I would prefer something similar as `iotjs_bufferwrap_from_jbuffer` except it return NULL if not a buffer. That way we could avoid getting a property twice (property read is costly).

src/modules/iotjs_module_tls.c
```diff
@@ -210,32 +210,29 @@ JS_FUNCTION(TlsContext) {
   jerry_value_t jkey =
       iotjs_jval_get_property(joptions, IOTJS_MAGIC_STRING_KEY);
 
-  if (jerry_value_is_string(jcert) && jerry_value_is_string(jkey)) {
-    iotjs_string_t cert = iotjs_jval_as_string(jcert);
-    const char *cert_chars = iotjs_string_data(&cert);
+  iotjs_string_t bcert = iotjs_string_create();
+  iotjs_string_t bkey = iotjs_string_create();
```
galpeter: There is no need to call the `iotjs_string_create` functions here. The `iotjs_jval_get_data` will do the require calls to create the strings.

src/modules/iotjs_module_tls.c
```diff
@@ -246,23 +243,23 @@ JS_FUNCTION(TlsContext) {
   // User provided trusted certificates
   jerry_value_t jcert_auth =
       iotjs_jval_get_property(joptions, IOTJS_MAGIC_STRING_CA);
+  iotjs_string_t bjcert_auth = iotjs_string_create();
```
galpeter: Just like above, the `ioths_jval_get_data` will create the required strings, no need to call the `iotjs_string_create` method.

src/iotjs_binding.c
```diff
@@ -85,6 +86,42 @@ double iotjs_jval_as_number(jerry_value_t jval) {
 }
 
 
+bool iotjs_jval_get_data(jerry_value_t jval, iotjs_string_t* getbuffer) {
+  iotjs_bufferwrap_t* buffer_wrap = NULL;
+  if(jerry_value_is_string(jval)) {
```
galpeter: The spaces in the code are not consistent:
* Add space after the `if` keyword. (just like line 94).
* Add spaces after the comma.

src/iotjs_binding.c
```diff
@@ -85,6 +86,42 @@ double iotjs_jval_as_number(jerry_value_t jval) {
...
+
+    *getbuffer = iotjs_string_create_with_buffer(buffer, size);
+
+    return true;
+  } else if (iotjs_jbuffer_is_bufferwrap(jval, &buffer_wrap) && buffer_wrap != NULL) {
+    char* bufferwrap_buffer = buffer_wrap->buffer;
+    jerry_size_t size = buffer_wrap->length;
+    char* buffer = iotjs_buffer_allocate(size + 1);
+
+    memcpy(buffer,bufferwrap_buffer,size);
```
galpeter: Missing spaces after the comma(s).

src/iotjs_binding.c
```diff
@@ -85,6 +86,42 @@ double iotjs_jval_as_number(jerry_value_t jval) {
...
+  if(jerry_value_is_string(jval)) {
+    jerry_size_t size = jerry_get_string_size(jval);
+
+    if (size == 0) {
+      *getbuffer = iotjs_string_create();
+      return false;
+    }
+
+    char* buffer = iotjs_buffer_allocate(size + 1);
+    jerry_char_t* jerry_buffer = (jerry_char_t*)(buffer);
```
galpeter: There is no need for the extra variable. The cast can be done here the `buffer` is passed.

test/run_pass/test_tls.js
```diff
@@ -30,7 +30,7 @@
 
  var options = {
    key: fs.readFileSync('resources/my_key.pem').toString(),
-   cert: fs.readFileSync('resources/my_crt.pem').toString(),
+   cert: fs.readFileSync('resources/my_crt.pem'),
```
galpeter: The other `toString` should also be removed, or not?

src/modules/iotjs_module_tls.c
galpeter: For the variable maybe the `cert_auth_string` would be a better name. The `b` prefix is a bit awkward IMHO.

src/modules/iotjs_module_tls.c
galpeter: A better name for the variables would be: `cert_string` and `key_string`.

src/modules/iotjs_module_buffer.c
```diff
@@ -82,6 +82,24 @@ size_t iotjs_bufferwrap_length(iotjs_bufferwrap_t* bufferwrap) {
 }
 
 
+bool iotjs_jbuffer_is_bufferwrap(const jerry_value_t jbuffer,
+                                 iotjs_bufferwrap_t** bufferwrap) {
+  if (jerry_value_is_object(jbuffer)) {
```
galpeter: No need to do extra check if the `jbuffer` is an object or not, the native pointer call will handle it.

src/modules/iotjs_module_tls.c
TamasZakor: We need it because if we don't get buffer or string as an argument, these two values remain uninitialized.

src/modules/iotjs_module_tls.c
TamasZakor: Same as my comment above.

test/run_pass/test_tls.js
TamasZakor: This way both functions calls are tested. So it is like that.

galpeter: Ok, cool.

src/modules/iotjs_module_tls.c
galpeter: That is not a problem, you should not use those variables if the conversion failed. The string destroy calls can be moved inside the `if` below.

src/iotjs_binding.c
```diff
@@ -85,6 +86,44 @@ double iotjs_jval_as_number(jerry_value_t jval) {
...
+      return false;
+    }
+
+    buffer = iotjs_buffer_allocate(size + 1);
+    size_t check =
+        jerry_string_to_char_buffer(jval, (jerry_char_t*)buffer, size);
+
+    IOTJS_ASSERT(check == size);
+
+  } else if (buffer_wrap) {
```
zherczeg: No need newline above.

src/iotjs_binding.c
```diff
@@ -85,6 +86,44 @@ double iotjs_jval_as_number(jerry_value_t jval) {
 }
 
 
+bool iotjs_jval_get_data(jerry_value_t jval, iotjs_string_t* getbuffer) {
+  iotjs_bufferwrap_t* buffer_wrap = iotjs_jbuffer_is_bufferwrap(jval);
```
zherczeg: We should only do this check if value is not string.

zherczeg: I find the function name a bit strange because it is not a boolean function. `iotjs_jbuffer_get_bufferwrap_or_null`

src/iotjs_binding.c
```diff
@@ -85,6 +86,44 @@ double iotjs_jval_as_number(jerry_value_t jval) {
 }
 
 
+bool iotjs_jval_get_data(jerry_value_t jval, iotjs_string_t* getbuffer) {
```
zherczeg: rename `getbuffer` to `out_string`

src/modules/iotjs_module_tls.c
```diff
@@ -210,30 +210,30 @@ JS_FUNCTION(TlsContext) {
...
-      iotjs_string_destroy(&key);
+                                 (size_t)iotjs_string_size(&key_string) + 1,
+                                 NULL, 0);
 
       if (ret == 0) {
         tls_context->context_flags |= SSL_CONTEXT_HAS_KEY;
       }
     }
+    iotjs_string_destroy(&cert_string);
+    iotjs_string_destroy(&key_string);
```
zherczeg: What happens if one of them is sucessful, but the other is not?

src/modules/iotjs_module_buffer.c
```diff
@@ -82,6 +82,20 @@ size_t iotjs_bufferwrap_length(iotjs_bufferwrap_t* bufferwrap) {
 }
 
 
+iotjs_bufferwrap_t* iotjs_jbuffer_is_bufferwrap(const jerry_value_t jbuffer) {
```
LaszloLango: Please rename to `iotjs_jbuffer_get_bufferwrap`

src/iotjs_binding.c
```diff
@@ -85,6 +86,43 @@ double iotjs_jval_as_number(jerry_value_t jval) {
 }
 
 
+bool iotjs_jbuffer_get_bufferwrap_or_null(jerry_value_t jval,
```
LaszloLango: The function name does not reflect to its behaviour. Please give a proper name to it.

TamasZakor: What do you think about the `iotjs_jbuffer_is_bufferwrap_or_string`?

LaszloLango: IMHO, `iotjs_jbuffer_get_bufferwrap` is enough

yichoi:
@zherczeg please review whether your comments had been reflected.

src/iotjs_binding.c
```diff
@@ -85,6 +86,43 @@ double iotjs_jval_as_number(jerry_value_t jval) {
 }
 
 
+bool iotjs_jbuffer_get_bufferwrap(jerry_value_t jval,
```
zherczeg: I still don't get this name: `iotjs_jbuffer_get_bufferwrap`. It is a function which converts a jerry string or a buffer to iotjs string. It does not return with a bufferwrap as the name suggests.

Other than that the patch is ok.

PR #1573, opened 2018-04-08
Fix emitData on readable stream

daeyeon:
- emitting 'data' should be handled when calling readable.read.
- assert.equal shouldn't be used inside the module.


yichoi:
LGTM

PR #1577, opened 2018-04-13
Rework the tls write function

imiklos:
Added the ability to buffer the data while the secure connect event not triggered.
After the event we can write to the socket.

### DIFF
--- a/src/js/tls.js
+++ b/src/js/tls.js
@@ -32,6 +32,8 @@ function TLSSocket(socket, options) {
   EventEmitter.call(this);
 
   this.authorized = false;
+  this.secureConnectionDone = false;
+  this.buff = [];
 
   this._socket.on('connect', this.onconnect);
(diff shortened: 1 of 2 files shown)


src/js/tls.js
```diff
@@ -68,11 +70,23 @@ TLSSocket.prototype.write = function(data, callback) {
...
+  if (!this.secureConnectionDone) {
+    if (this.buff.length > 0) {
+      this.buff = Buffer.concat([this.buff, data]);
+      return;
+    }
+    this.buff = data;
+    return;
+  } else {
+    return writeOut(this, data, callback);
+  }
```
zherczeg: Looks like a manual implementation. Does writable provide any interface for this? If yes, perhaps we could inherit it.

yichoi:
@imiklos please rebase it to the master

zherczeg:
#1595 should solve this issue

PR #1583, opened 2018-04-17
Adding MQTT module(client)

DanielBallaSZTE:
This patch adds the MQTT client support to IoT.js.

Co-authored-by: Daniel Vince <vinced@inf.u-szeged.hu>


src/modules/iotjs_module_mqtt.c
```diff
@@ -0,0 +1,584 @@
...
+         strlen(enc_remaining_length));
+  offset += strlen(enc_remaining_length);
+
+  // Variable header
+  memcpy(buffer_wrap->buffer + offset, variable_header_protocol,
+         sizeof(variable_header_protocol));
+  offset += sizeof(variable_header_protocol);
+
+  // Flags
+  memcpy(buffer_wrap->buffer + offset, &connect_flag, sizeof(connect_flag));
```
zherczeg: No need `memcpy` here, simply `buffer_wrap->buffer[offset] = connect_flags`

src/modules/iotjs_module_mqtt.c
```diff
@@ -0,0 +1,584 @@
...
+  jerry_value_t jclient_id =
+      iotjs_jval_get_property(joptions, IOTJS_MAGIC_STRING_CLIENTID);
+  jerry_value_t jusername =
+      iotjs_jval_get_property(joptions, IOTJS_MAGIC_STRING_USERNAME);
+  jerry_value_t jpassword =
+      iotjs_jval_get_property(joptions, IOTJS_MAGIC_STRING_PASSWORD);
+  jerry_value_t jkeepalive =
+      iotjs_jval_get_property(joptions, IOTJS_MAGIC_STRING_KEEPALIVE);
+
+  uint8_t connect_flag = 0;
```
zherczeg: plural: flags

src/modules/iotjs_module_mqtt.c
```diff
@@ -0,0 +1,584 @@
...
+  char *buffer = buffer_wrap->buffer;
+
+  char first_byte = buffer[0];
+  char remaining_length = buffer[1];
+  char packet_type = (first_byte >> 4) & 0b00001111;
+
+  switch (packet_type) {
+    case CONNACK:
+      if (remaining_length != 2) {
+        return JS_CREATE_ERROR(COMMON, "MQTT: CONNACK packet is corrupted");
```
zherczeg: Maybe we should think about closing the connection.

src/modules/iotjs_module_mqtt.c
```diff
@@ -0,0 +1,584 @@
...
+
+  jerry_value_t jsref = JS_GET_ARG(0, object);
+  jerry_value_t jparam = JS_GET_ARG(1, object);
+
+  iotjs_bufferwrap_t *buffer_wrap = iotjs_bufferwrap_from_jbuffer(jparam);
+  char *buffer = buffer_wrap->buffer;
+
+  char first_byte = buffer[0];
+  char remaining_length = buffer[1];
+  char packet_type = (first_byte >> 4) & 0b00001111;
```
zherczeg: 0x0f

src/modules/iotjs_module_mqtt.h
```diff
@@ -0,0 +1,241 @@
...
+    uint8_t retain : 1;
+    uint8_t qos : 2;
+    uint8_t dup : 1;
+    uint8_t type : 4;
+  } bits;
+} MQTTHeader;
+
+enum {
+  // Reserved bit, must be 0
+  FLAG_RESERVED = 1 << 0,
```
zherczeg: MQTT_FLAG_....

src/modules/iotjs_module_mqtt.h
```diff
@@ -0,0 +1,241 @@
...
+  FLAG_USERNAME = 1 << 7
+} iotjs_mqtt_connect_flag_t;
+
+
+typedef struct {
+  struct {
+    uint8_t msb;
+    uint8_t lsb;
+    char *buffer;
+  } data;
```
zherczeg: Do we need this struct in struct?

DanielBallaSZTE: No, we don't thanks for pointing out.

DanielBallaSZTE:
@zherczeg I've changed those things, thank you for your time!

src/modules/iotjs_module_mqtt.c
```diff
@@ -0,0 +1,677 @@
...
+    (*offset)++;
+   } while ((c & 128) != 0);
+
+  return remaining_length;
+}
+
+
+static void iotjs_mqtt_calculate_msb_lsb(uint16_t len, uint8_t *msb,
+                                         uint8_t *lsb) {
+  *msb = (uint8_t)((len & 0xFF00) >> 8);
```
zherczeg: no need `& 0xFF00`

src/modules/iotjs_module_mqtt.c
```diff
@@ -0,0 +1,677 @@
...
+  size_t full_len = sizeof(header_byte) + payload_len + variable_header_len +
+                    strlen(enc_remaining_length);
+
+  jerry_value_t jbuff = iotjs_bufferwrap_create_buffer(full_len);
+  iotjs_bufferwrap_t *buffer_wrap = iotjs_bufferwrap_from_jbuffer(jbuff);
+
+  uint8_t *buff_ptr = (uint8_t *)buffer_wrap->buffer;
+
+  // Fixed header
+  *buff_ptr++ = header_byte;
```
zherczeg: I think this is the typical case when you don't need comment. It is obvious what you do here.

src/modules/iotjs_module_mqtt.c
```diff
@@ -0,0 +1,677 @@
...
+      return JS_CREATE_ERROR(COMMON, "MQTT: Unknown error");
+  }
+}
+
+
+/*
+ * Waits for a uint32_t remaining_length.
+ * Returns the encoded value with the correct size.
+ */
+static char *iotjs_encode_remaining_length(uint32_t len) {
```
zherczeg: I feel this is overcomplicated. You could write the data directly into the stream.

src/modules/iotjs_module_mqtt.c
```diff
@@ -0,0 +1,677 @@
...
+  size_t rc = 0;
+  // It is an uint32_t at maximum
+  char temp_buff[4];
+  do {
+    char d = len % 128;
```
zherczeg: `len & 0x7F;`

src/modules/iotjs_module_mqtt.c
```diff
@@ -0,0 +1,677 @@
...
+    len /= 128;
```
zherczeg: `len >>= 7`

src/modules/iotjs_module_mqtt.h
```diff
@@ -0,0 +1,243 @@
...
+} iotjs_mqtt_connection_error;
+
+/*
+ * First byte of the message's fixed header.
+ * Contains:
+ * - MQTT Control Packet type,
+ * - Specific flags to each MQTT Control Packet.
+ */
+typedef struct {
+#ifdef MQTT_REVERSED
```
zherczeg: How can this be reversed???? There is no bit order, only byte order.

DanielBallaSZTE: This is left over from some old code, we never use it, it's going to get deleted when we clean up the header file.

src/modules/iotjs_module_mqtt.c
DanielBallaSZTE: We could, but we need the size of that data to allocate the buffer.

DanielBallaSZTE: I figured out a solution for this.

DanielBallaSZTE:
@zherczeg I have updated the PR, thank you!

src/modules/iotjs_module_mqtt.c
```diff
@@ -0,0 +1,777 @@
...
+
+static uint32_t iotjs_decode_remaining_length(char *buffer, size_t *offset) {
+  unsigned char c;
+  uint32_t remaining_length = 0;
+  uint32_t length = 0;
+  uint32_t multiplier = 1;
+
+  do {
+    if (++length > IOTJS_MODULE_MQTT_MAX_REMAINING_LENGTH_BYTES) {
+      return JS_CREATE_ERROR(COMMON, "MQTT: The arrived message is too big");
```
zherczeg: You should return with a special value to show error. E.g. `UINT32_MAX`, since the message size is limited to 256 Mbyte.

src/modules/iotjs_module_mqtt.c
```diff
@@ -0,0 +1,777 @@
...
+    remaining_length += (c & 127) * multiplier;
+    multiplier *= 128;
+    (*offset)++;
+  } while ((c & 128) != 0);
+
+  return remaining_length;
+}
+
+
+static void iotjs_mqtt_calculate_msb_lsb(uint16_t len, uint8_t *msb,
```
zherczeg: In general I would do the following for such functions:
```
static uint8_t* iotjs_mqtt_encode_length(uint8_t *buffer, uint16_t len)
{
  buffer[0] = (uint8_t)(len >> 8);
  buffer[1] = (uint8_t)(len & 0xff);
  return buffer + 2;
}
```
So they copy their arguments into the stream and advances the string pointer.

src/modules/iotjs_module_mqtt.c
```diff
@@ -0,0 +1,777 @@
...
+
+  jerry_value_t jbuff = iotjs_bufferwrap_create_buffer(full_len);
+  iotjs_bufferwrap_t *buffer_wrap = iotjs_bufferwrap_from_jbuffer(jbuff);
+
+  uint8_t *buff_ptr = (uint8_t *)buffer_wrap->buffer;
+
+  *buff_ptr++ = header_byte;
+  *buff_ptr++ = remaining_length;
+  *buff_ptr++ = packet_identifier_msb;
+  *buff_ptr++ = packet_identifier_lsb;
```
zherczeg: For the last bytes you can use `buff_ptr[0]` - `buff_ptr[3]`. The last ++ has no effect anyway.

src/modules/iotjs_module_mqtt.c
```diff
@@ -0,0 +1,777 @@
...
+JS_FUNCTION(MqttUnsubscribe) {
+  DJS_CHECK_THIS();
+  DJS_CHECK_ARGS(1, object);
+
+  jerry_value_t jtopic = JS_GET_ARG(0, object);
+
+  iotjs_mqtt_payload_t topic_payload =
+      iotjs_create_payload(iotjs_bufferwrap_from_jbuffer(jtopic));
+
+  uint8_t header_byte = 0;
```
zherczeg: `uint8_t header_byte = (UNSUBSCRIBE << 4);`
No need every step in separeted lines.

src/modules/iotjs_module_mqtt.c
```diff
@@ -0,0 +1,777 @@
...
+
+  uint8_t packet_identifier_msb = 0;
+  uint8_t packet_identifier_lsb = 0;
+
+  size_t payload_len = topic_payload.size;
+  size_t variable_header_len =
+      sizeof(packet_identifier_msb) + sizeof(packet_identifier_lsb);
+  uint32_t remaining_length = payload_len + variable_header_len;
+  size_t full_len = sizeof(header_byte) + payload_len + variable_header_len +
+                    get_remaining_length_size(remaining_length);
```
zherczeg: I would compute these in message order:
`sizeof(header_byte) + get_remaining_length_size(remaining_length) + variable_header_len + payload_len`

src/modules/iotjs_module_mqtt.c
```diff
@@ -0,0 +1,777 @@
...
+
+  jerry_value_t jbuff = iotjs_bufferwrap_create_buffer(full_len);
+  iotjs_bufferwrap_t *buffer_wrap = iotjs_bufferwrap_from_jbuffer(jbuff);
+
+  uint8_t *buff_ptr = (uint8_t *)buffer_wrap->buffer;
+
+  *buff_ptr++ = header_byte;
+
+  buff_ptr += iotjs_encode_remaining_length(remaining_length, buff_ptr);
```
zherczeg: if `iotjs_encode_remaining_length` returns with buffer, no need `+=`. The `=` is always faster since the compiler does not need to preserve the value of `buffer`.

src/modules/iotjs_module_mqtt.c
```diff
@@ -0,0 +1,777 @@
...
+
+  jerry_value_t joptions = JS_GET_ARG(0, object);
+
+  jerry_value_t jtopic =
+      iotjs_jval_get_property(joptions, IOTJS_MAGIC_STRING_TOPIC);
+  jerry_value_t jqos =
+      iotjs_jval_get_property(joptions, IOTJS_MAGIC_STRING_QOS);
+
+  uint8_t qos = 0;
+  if (!jerry_value_is_undefined(jqos) && !jerry_value_has_error_flag(jqos)) {
```
zherczeg: I think there is a utility now which converts strings and buffers to char*, which is exactly what you need here. It also returns with error if the conversion is not possible.

src/modules/iotjs_module_mqtt.c
```diff
@@ -0,0 +1,777 @@
...
+  DJS_CHECK_ARGS(2, object, object);
+
+  jerry_value_t jsref = JS_GET_ARG(0, object);
+  jerry_value_t jparam = JS_GET_ARG(1, object);
+
+  iotjs_bufferwrap_t *buffer_wrap = iotjs_bufferwrap_from_jbuffer(jparam);
+  char *buffer = buffer_wrap->buffer;
+
+  char first_byte = buffer[0];
+  char packet_type = (first_byte >> 4) & 0b00001111;
```
zherczeg: `0x0f`. Please don't use 0b

DanielBallaSZTE:
@zherczeg Thank you again for the review, I have modified lots of things, and removed some unused things from the header file. What we still need are some checks if the user gives the correct input.
Besides from that, I'm open for any suggestions :)

src/js/mqtt.js
```diff
@@ -0,0 +1,308 @@
...
+  this._protocol = options.protocol || 'net';
+
+  this._clientOptions = Object.create(options, {
+    host: { value: options.host || '127.0.0.1'},
+    port: { value: options.port || 8883 },
+    qos: { value: options.qos || 0 },
+    keepalive: { value: options.keepalive || 60 },
+  });
+
+  this._socket = new net.Socket();
```
zherczeg: Is it a simple `net.Socket` even if tls is specified? I would remove the _protocol member and would simply accept a duplex stream as options.socket. If it is not specified, a new `net.Socket` is created.

src/js/mqtt.js
```diff
@@ -0,0 +1,308 @@
...
+  this._socket.on('data', function(data) {
+    jsref.ondata(jsref, data);
+  });
+  this._socket.on('error', function(e) {
+    jsref.emit('error', e);
+  });
+  this._socket.on('end', this.ondisconnect);
+};
+
+MQTTClient.prototype.disconnect = function(error) {
```
zherczeg: I think `end` is more common than `disconnect`.

src/js/mqtt.js
```diff
@@ -0,0 +1,292 @@
...
+MQTTClient.prototype.unsubscribe = function(topic) {
+  if (!Buffer.isBuffer(topic)) {
+    topic = new Buffer(topic);
+  }
+
+  var buf = native.unsubscribe(topic);
+  this._socket.write(buf);
+}
+
+MQTTClient.prototype.onpingresp = function() {
```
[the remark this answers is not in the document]
zherczeg: Let me rephrase the question. These seem internal functions. However they are exposed through `MQTTClient.prototype`. What about simply create a `function onpingresp() {...}`. This function would be not visible externally, but could use anything internally.

src/js/mqtt.js
```diff
@@ -0,0 +1,308 @@
...
+}
+
+function getClient(connectionOptions) {
+  if (util.isUndefined(connectionOptions.clientId)) {
+    connectionOptions.clientId = defaultClientId();
+  }
+  if (!Buffer.isBuffer(connectionOptions.clientId)) {
+    connectionOptions.clientId = new Buffer(connectionOptions.clientId);
+  }
+  return MQTTClient(connectionOptions);
```
zherczeg: `return new MQTTClient(connectionOptions);`

src/modules/iotjs_module_mqtt.h
```diff
@@ -0,0 +1,130 @@
...
+  DISCONNECT = 0xE
+} iotjs_mqtt_control_packet_type;
+
+/*
+ * The values of the Quality of Service.
+ */
+enum {
+  QoS0 = 0b00, // At most once delivery.
+  QoS1 = 0b01, // At least once delivery.
+  QoS2 = 0b10  // Exactly once delivery.
```
zherczeg: Please use 0x, or here you can use decimal numbers.

src/modules/iotjs_module_mqtt.c
```diff
@@ -0,0 +1,719 @@
...
+    case UNAUTHORISED:
+      return JS_CREATE_ERROR(COMMON, "MQTT: Connection refused: unauthorised");
+    default:
+      return JS_CREATE_ERROR(COMMON, "MQTT: Unknown error");
+  }
+}
+
+
+static uint8_t *iotjs_encode_remaining_length(uint32_t len,
+                                              unsigned char *buffer) {
```
zherczeg: `uint8_t *buffer`. Usually buffer (= destination byte-stream pointer) is always the first argument for such functions.

src/modules/iotjs_module_mqtt.c
```diff
@@ -0,0 +1,719 @@
...
+  while (len != 0) {
+    len >>= 7;
+    n++;
+  }
+
+  return n;
+}
+
+
+static uint32_t iotjs_decode_remaining_length(char *buffer, size_t *offset) {
```
zherczeg: Since `get_remaining_length_size` is called earlier, I would move the error checking there, and just use some asserts here. Perhaps get_remaining_length_size could return 0 on error, since at least 1` byte is required if the encoding is possible.

src/modules/iotjs_module_mqtt.c
```diff
@@ -0,0 +1,719 @@
...
+  uint32_t length = 0;
+  uint32_t multiplier = 1;
+
+  do {
+    if (++length > IOTJS_MODULE_MQTT_MAX_REMAINING_LENGTH_BYTES) {
+      return UINT32_MAX;
+    }
+    c = (unsigned char)buffer[*offset];
+    remaining_length += (c & 127) * multiplier;
+    multiplier *= 128;
```
zherczeg: `multiplier <<= 7`

zherczeg: Btw usually shift is faster than multiplication, so you could simply use a shift:
```
remaining_length += (c & 0x7F) << shift;
shift += 7
```
Also `|=` can be used instead of `+=` here, but they are equally fast so they don't provide any advantage.

src/modules/iotjs_module_mqtt.c
```diff
@@ -0,0 +1,719 @@
...
+    (*offset)++;
+  } while ((c & 128) != 0);
+
+  return remaining_length;
+}
+
+
+static void iotjs_mqtt_calculate_length(uint16_t *len, uint8_t msb,
+                                        uint8_t lsb) {
+  *len = (msb << 8) | lsb;
```
zherczeg: What about `static uint16_t iotjs_mqtt_calculate_length(uint8_t msb, uint8_t lsb) { ... }`?

src/modules/iotjs_module_mqtt.c
```diff
@@ -0,0 +1,719 @@
...
+}
+
+
+static uint8_t *iotjs_mqtt_string_serialize(uint8_t *dst_buffer,
```
zherczeg: For me encode/serialize is equally ok, but please only use on of them. E.g `iotjs_encode_remaining_length` -> `iotjs_mqtt_serialize_remaining_length` or `iotjs_mqtt_string_encode`. Perhaps encode sounds a bit better, but either of them is ok.

src/modules/iotjs_module_mqtt.c
```diff
@@ -0,0 +1,719 @@
...
+    connect_flags |= MQTT_FLAG_PASSWORD;
+  }
+  if (!jerry_value_is_undefined(jkeepalive)) {
+    uint16_t len = jerry_get_number_value(jkeepalive);
+    keep_alive_msb = (uint8_t)(len >> 8);
+    keep_alive_lsb = (uint8_t)(len & 0x00FF);
+  }
+
+
+  iotjs_bufferwrap_t *client_id = iotjs_bufferwrap_from_jbuffer(jclient_id);
```
zherczeg: No type check here. I mentioned that there is a function which creates iotjs string from strings/buffers, or return NULL on error.

src/modules/iotjs_module_mqtt.c
```diff
@@ -0,0 +1,719 @@
...
+  header_byte |= (jerry_get_boolean_value(jretain));
+
+  iotjs_bufferwrap_t *message_payload = iotjs_bufferwrap_from_jbuffer(jmessage);
+  iotjs_bufferwrap_t *topic_payload = iotjs_bufferwrap_from_jbuffer(jtopic);
+
+  uint8_t packet_identifier_lsb = 0;
+  uint8_t packet_identifier_msb = 0;
+
+  // IF NOT QOS
+  if (!jerry_value_is_undefined(jqos) &&
```
zherczeg: There isa  to_boolean function, which returns with `bool` (never throws error). I would use it here.

src/modules/iotjs_module_mqtt.c
```diff
@@ -0,0 +1,719 @@
...
+        return JS_CREATE_ERROR(COMMON, "MQTT: PingRESP packet is corrupted");
+      }
+      jerry_value_t fn =
+          iotjs_jval_get_property(jsref, IOTJS_MAGIC_STRING_ONPINGRESP);
+      iotjs_make_callback(fn, jsref, iotjs_jargs_get_empty());
+      jerry_release_value(fn);
+      break;
+    }
+    case DISCONNECT:
+      return JS_CREATE_ERROR(COMMON, "MQTT: Broker disconnected the client");
```
zherczeg: Why this is an error? Isn't it a notification?

src/modules/iotjs_module_mqtt.c
```diff
@@ -0,0 +1,719 @@
...
+
+    case CONNECT:
+    case SUBSCRIBE:
+    case UNSUBSCRIBE:
+    case PINGREQ:
```
zherczeg: Server cannot send ping?

src/modules/iotjs_module_mqtt.c
```diff
@@ -0,0 +1,719 @@
...
+      iotjs_make_callback(fn, jsref, iotjs_jargs_get_empty());
+      jerry_release_value(fn);
+      break;
+    }
+    case PUBLISH: {
+      MQTTHeader header = { 0 };
+      header.bits.type = PUBLISH;
+      header.bits.dup = first_byte & 0x08;
+      header.bits.qos = (first_byte & 0x06) >> 1;
+      header.bits.retain = first_byte & 0x01;
```
zherczeg: I think more checks would be useful here. E.g. jerry offers utf8 validating functions.

src/modules/iotjs_module_mqtt.c
```diff
@@ -0,0 +1,719 @@
...
+
+      iotjs_jargs_destroy(&args);
+      jerry_release_value(jmessage);
+      jerry_release_value(jtopic);
+
+      break;
+    }
+    case PUBACK: {
+      if (remaining_length != 2) {
+        return JS_CREATE_ERROR(COMMON, "MQTT: PUBACK packet is corrupted");
```
zherczeg: Is the connection is aborted on error? I think the standard expects it.

src/modules/iotjs_module_mqtt.c
```diff
@@ -0,0 +1,719 @@
...
+
+  jerry_release_value(jmessage);
+  jerry_release_value(jtopic);
+  jerry_release_value(jretain);
+
+  return jbuff;
+}
+
+
+JS_FUNCTION(MqttHandle) {
```
zherczeg: What happens if the package is not fully arrived?

zherczeg:
Travis: `Signed-off-by message is incorrect. The following line should be at the end of the e5135d6 commit's message: 'IoT.js-DCO-1.0-Signed-off-by: Daniel Vince vinced@inf.u-szeged.hu'. `

src/js/mqtt.js
DanielBallaSZTE: There actually is a packet type which has the name `DISCONNECT`, causing the server to properly disconnect the client.

src/js/mqtt.js
DanielBallaSZTE: Yeah about the prototype thing, we haven't refactored the JS code yet, were focusing on the C side. 
We were planning on changing them as you mentioned it.

src/modules/iotjs_module_mqtt.c
DanielBallaSZTE: I don't think it is needed since JS side a buffer is always created for the client id, and that's passed to the C side.

src/modules/iotjs_module_mqtt.c
vincedani: The maximum TCP package size is 64K, but due to memory limitations the IoT.js can't handle such a big packages.

DanielBallaSZTE:
@daeyeon I've updated the commit! Thank you

yichoi:
LGTM

PR #1585, opened 2018-04-18
Allow the tls module to work with Duplex streams as well

pmarkee:

### DIFF
--- a/src/js/tls.js
+++ b/src/js/tls.js
@@ -15,7 +15,7 @@
 
 var net = require('net');
 var util = require('util');
-var EventEmitter = require('events').EventEmitter;
+var Duplex = require('stream').Duplex;
 
 function TLSSocket(socket, options) {
   if (!(this instanceof TLSSocket)) {
@@ -29,7 +29,7 @@ function TLSSocket(socket, options) {
   this._socket = socket;
   socket._tlsSocket = this;
 
-  EventEmitter.call(this);
+  Duplex.call(this);
 
   this.authorized = false;
 
@@ -48,15 +48,23 @@ function TLSSocket(socket, options) {
 
   native.TlsInit(this, options, secureContext);
   this._socketState = socket._socketState;
-}
-util.inherits(TLSSocket, EventEmitter);
 
-TLSSocket.prototype._read = native.read;
-TLSSocket.prototype._write = native.write;
-TLSSocket.prototype._connect = native.connect;
+  var self = this;
+  if (socket._writableState.ready && !options.isServer) {
+    process.nextTick(function() {
+      self._native_connect(options.servername || options.host || 'localhost');
+      self._native_read(null);
+    });
+  }
+}
+util.inherits(TLSSocket, Duplex);
+
+TLSSocket.prototype._native_read = native.read;
+TLSSocket.prototype._native_write = native.write;
+TLSSocket.prototype._native_connect = native.connect;
 
 TLSSocket.prototype.connect = function(options, callback) {
-  this._connect(options.servername || options.host || 'localhost');
+  this._native_connect(options.servername || options.host || 'localhost');
 
   if (util.isFunction(callback)) {
     this.on('secureConnect', callback);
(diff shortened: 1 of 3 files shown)


src/js/tls.js
```diff
@@ -48,15 +48,20 @@ function TLSSocket(socket, options) {
...
   native.TlsInit(this, options, secureContext);
   this._socketState = socket._socketState;
+
+  if (socket._writableState.ready && !options.isServer) {
+    this.native_connect(options.servername || options.host || 'localhost');
+    this.native_read(null);
+  }
 }
-util.inherits(TLSSocket, EventEmitter);
+util.inherits(TLSSocket, stream.Duplex);
```
zherczeg: I would investigate a bit more what is the best option here (mem/complexity).

src/js/tls.js
```diff
@@ -48,15 +48,20 @@ function TLSSocket(socket, options) {
...
 
-TLSSocket.prototype._read = native.read;
-TLSSocket.prototype._write = native.write;
-TLSSocket.prototype._connect = native.connect;
+TLSSocket.prototype.native_read = native.read;
```
zherczeg: `native_read` -> `_native_read` Rename the other three as well.

zherczeg:
I have been thinking and there is no good way to avoid Duplex inheritance. So this is the best solution we have now.

src/js/tls.js
```diff
@@ -15,7 +15,7 @@
 
 var net = require('net');
 var util = require('util');
-var EventEmitter = require('events').EventEmitter;
+var stream = require('stream');
```
zherczeg: var Duplex = require('stream').Duplex;

yichoi:
@pmarkee please update along with @zherczeg's comment

src/js/tls.js
```diff
@@ -48,51 +50,40 @@ function TLSSocket(socket, options) {
...
-    if (!Buffer.isBuffer(data)) {
-      data = new Buffer(data);
-    }
-    data = this._write(data, true);
-  } else {
-    data = this._write(null, true);
-  }
-
-  this._socket.end(data, callback);
+  Writable.prototype.end.call(this, data, callback);
```
zherczeg: I think Duplex.prototype.end would work here as well, and we could remove the Writable requirement.

src/js/tls.js
```diff
@@ -48,51 +48,40 @@ function TLSSocket(socket, options) {
...
     this.on('secureConnect', callback);
   }
-
-  this._socket.connect(options);
-};
-
-TLSSocket.prototype.write = function(data, callback) {
-  if (!Buffer.isBuffer(data)) {
-    data = new Buffer(data);
+  if (this._socket instanceof net.Socket) {
```
zherczeg: Do we need to check this? Connect should not be called unless `_socket` has a connect (tls.connect calls this).

src/js/tls.js
```diff
@@ -48,51 +48,40 @@ function TLSSocket(socket, options) {
 
   native.TlsInit(this, options, secureContext);
   this._socketState = socket._socketState;
+
+  var self = this;
+  process.nextTick(function() {
+    if (socket._writableState.ready && !options.isServer) {
```
zherczeg: I would check this first, and call `process.nextTick` if it is true.

src/js/tls.js
```diff
@@ -212,10 +197,9 @@ function createServer(options, secureConnectionListener) {
 
 function connect(arg0, arg1, arg2, callback) {
   var options;
-  var tlsSocket;
   if (typeof arg0 == 'object') {
     options = Object.create(arg0, {
-      isServer: { value: false, enumerable: true },
+      isServer: { value: arg0.isServer || false, enumerable: true },
```
zherczeg: `|| false` is pointless. Just let it be `arg0.isServer`.

src/js/tls.js
```diff
@@ -257,8 +241,14 @@ function connect(arg0, arg1, arg2, callback) {
...
     }
   }
-  tlsSocket = new TLSSocket(new net.Socket(), options);
-  tlsSocket.connect(options, callback);
+
+  var tlsSocket = new TLSSocket(options.socket || new net.Socket(), options);
+  if (tlsSocket._socket instanceof net.Socket) {
+    tlsSocket.connect(options, callback);
+  }
+  if (util.isFunction(callback)) {
```
zherczeg: Isn't this should be an `else if`?

src/js/tls.js
zherczeg: I just realiezd I don't understand this. How can a server connects???

pmarkee: I left that in by accident, it doesn't have to be there

src/js/tls.js
pmarkee: I just realized you're right, yes, it has to be else if. The way I did it caused https tests to fail or timeout.

yichoi:
LGTM

PR #1589, opened 2018-04-19
Add backtrace printing for uncaught exceptions

LaszloLango:
Print backtrace of an uncaught exception by default in debug build.
This feature is not enabled by default in release builds.


tools/build.py
```diff
@@ -343,7 +347,16 @@ def build_iotjs(options):
 
     # --jerry-debugger
     if options.jerry_debugger:
-        cmake_opt.append('-DFEATURE_DEBUGGER=ON')
+        cmake_opt.append("-DFEATURE_DEBUGGER=ON")
```
zherczeg: What is the change in this line?

LaszloLango: Single quote to double quote to be more consistent. It is unrelated, so I can revert is you want.

zherczeg: No need. I haven't noticed it.

yichoi:
LGTM

PR #1590, opened 2018-04-19
Update 'test_fs_mkdir_rmdir.js' to fix an erroneous behaviour

LaszloLango:
The path of 'root' directory was reused after the 'fs.rmdir(root, ...)'
call which is an async call. 'root' directory might still exists when
'fs.mkdir(root, ...)' is called again. This patch moves the related code
into the callback of 'fs.rmdir' to avoid this erroneous behaviour.


yichoi:
LGTM

PR #1591, opened 2018-04-20
Add checking uvloop

haesik:
uv_default_loop() can return NULL, but we didn't check the return value


LaszloLango:
typo in the commit message: `uv_devault_loop` -> `uv_default_loop`

PR #1592, opened 2018-04-20
Fix main sequence to avoid errors

daeyeon:
This commit includes the below fixes:
- Not clean up jerry when iotjs_initialize fails
- Segment faults due to debuglog initialization
- Naming mismatching to iotjs_start
- Missing the state change


yichoi:
LGTM

PR #1596, opened 2018-04-23
Fix iotjs-string allocation

haesik:
The size inside of iotjs_string not include null character,
So, size + 1 buffer is needed to store a string

### DIFF
--- a/src/iotjs_string.c
+++ b/src/iotjs_string.c
@@ -21,6 +21,7 @@
 #include <string.h>
 
 
+
 iotjs_string_t iotjs_string_create() {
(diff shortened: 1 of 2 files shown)


src/iotjs_string.c
```diff
@@ -21,6 +21,7 @@
 #include <string.h>
 
 
+
```
LaszloLango: Unnecessary new line

LaszloLango:
See also: #479 and #636

haesik:
ok, I understand, I will close this pr..

PR #1600, opened 2018-04-24
Update sample.gbs.conf

haesik:
I have moditified sample.gbs.conf to test latest version of tizen 4.0 unified and tizen 5.0 unified


yichoi:
LGTM

PR #1602, opened 2018-04-25
Add calling js functions based on string parameters

daeyeon:
As written in iotjs_tizen_service_app.c, iot.js app for tizen is a
hybrid one, which combines features of both native and js app.
This commit provides a way to call js function bound with tizen
module in the app model.

This also includes a patch for the bug in getResPath.

### DIFF
--- a/include/iotjs_module_tizen.h
+++ b/include/iotjs_module_tizen.h
@@ -1,0 +1,47 @@
+/* Copyright 2018-present Samsung Electronics Co., Ltd. and other contributors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef IOTJS_MODULE_TIZEN_H
+#define IOTJS_MODULE_TIZEN_H
+
+#ifndef IOTJS_EXTERN_C
+#ifdef __cplusplus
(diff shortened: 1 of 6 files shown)


src/modules/tizen/iotjs_module_tizen-tizen.c
```diff
@@ -17,20 +17,21 @@
...
 
-// application control
+// # tizen app-control
+#include <app_control.h>
 #include <app_control_internal.h>
+#include <bundle.h>
 #include <bundle_internal.h>
 
-iotjs_error_t send_launch_request(const char* json, void* hbridge) {
+iotjs_error_t tizen_send_launch_request(const char* json, void* hbridge) {
```
zherczeg: Name should be `iotjs_tizen_send_launch_request`

src/modules/tizen/iotjs_module_tizen-tizen.c
```diff
@@ -149,3 +151,106 @@ void iotjs_tizen_func(const char* command, const char* message, void* handle) {
...
+
+static char* create_string_buffer(const char* src, size_t size) {
+  char* dest = IOTJS_CALLOC(size + 1, char);
+  strncpy(dest, src, size);
+  dest[size] = '\0';
+  return dest;
+}
+
+
+bool bridge_native_call(const char* module_name, const char* func_name,
```
zherczeg: Can it be a `static` function?

include/iotjs_module_tizen.h
```diff
@@ -0,0 +1,46 @@
...
+
+IOTJS_EXTERN_C
+int iotjs_tizen_bridge_native(const char* fn_name, unsigned fn_name_size,
+                              const char* message, unsigned message_size,
+                              user_callback_t cb);
+
+#define IOTJS_TIZEN_CALL_JFUNC(name, msg, cb)                              \
+  ({                                                                       \
+    if (name != NULL && (msg) != NULL)                                     \
+      iotjs_tizen_bridge_native(name, sizeof(name), msg, sizeof(msg), cb); \
```
galpeter: Is the `sizeof` correct here? Shouldn't this be `strlen` ?

yichoi:
@galpeter please review whether your comments had been reflected.

PR #1605, opened 2018-04-26
Add tizen.org patch

haesik:
Add some patches for uploading iot.js to tizen.org


haesik:
When we uploading iot.js to tizen.org, we should satisfy the Tizen's rules.
Patch file and script can be use at that time..

yichoi:
LGTM

PR #1607, opened 2018-04-26
Disable PIE when iotjs build type set to debug

### LINKED ISSUE #1566: Missing crash backtrace above gcc 4.8
###### IoT.js version:

```
Checked revision: e3cd187
Build command: tools/build.py --buildtype debug
```
###### OS:

```
Ubuntu 17.10, x86_64
```

Debug IoT.js binaries built with gcc-4.8 and below provide immediate backtrace with the error message in case of a crash. However, if the binary was built with a newer gcc then the backtrace contains empty frames (`??:0`)

Related [commit](https://github.com/Samsung/iotjs/commit/5c00a91e1f8ebc2fee81445d24245ce8ce4c1563)

imiklos:
In certain operating systems the gcc is configured to build `-pie` binaries by default.
In this case we do not have backtrace information because the addr2line cannot resolve the memory addresses.
Fixes #1566

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 cmake_minimum_required(VERSION 2.8)
+include(CheckCCompilerFlag)
 
 project(IOTJS C)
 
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -67,9 +68,14 @@ macro(iotjs_add_link_flags)
   iotjs_add_flags(IOTJS_LINKER_FLAGS ${ARGV})
 endmacro()
 
+CHECK_C_COMPILER_FLAG("-no-pie" HAS_NO_PIE)
```
daeyeon: I suggest removing the double-quotes here like `CHECK_C_COMPILER_FLAG(-no-pie HAS_NO_PIE)`

crippin: Thank you, I updated the PR as you suggested.

imiklos:
@daeyeon Thank you, I updated the PR as you suggested.

yichoi:
LGTM

PR #1608, opened 2018-04-27
Hide internal functions in process

daeyeon:


daeyeon:
This may not the best approach. Better suggestions are welcome.

yichoi:
LGTM

PR #1612, opened 2018-05-02
Always enable debug log on Tizen

hs0225:
Because of the Tizen policy, the log is also displayed in release mode.


yichoi:
LGTM

PR #1613, opened 2018-05-02
Fix iotjs_file_read

haesik:
ftell can return -1, but we didn't check it on release mode


yichoi:
LGTM

PR #1615, opened 2018-05-02
bug fix async callback of bridge

haesik:
Allocation async structure should saved in the bridgecall structure


yichoi:
LGTM

PR #1616, opened 2018-05-02
Fix the token for SonarQube

daeyeon:
The token already landed turns out wrong. It seems to occur because
the previous one wasn't properly bound to the iot.js repo. This patch
includes a new token secured for that.


yichoi:
LGTM

PR #1620, opened 2018-05-03
Expose the API implemented for the Tizen service app to the template

hs0225:


samples/tizen-bridge-native/tizen_bridge_native.c
```diff
@@ -14,7 +14,7 @@
  */
 
 #include "iotjs.h"
-#include "iotjs_module_tizen.h"
```
daeyeon: This sample assumes that an `include path` containing `iotjs_tizen_service_app.h` is set in compile time. So we can use `iotjs_tizen_service_app.h`.

hs0225: I fixed it.

PR #1622, opened 2018-05-03
Update the documentations

LaszloLango:


docs/build/Build-Script.md
```diff
@@ -48,76 +48,91 @@ If given path is not exist, build.py will create it.
...
 ```
-./tools/build.py --target-arch=arm
+./tools/build.py --clean
 ```
---
-#### `--target-os`
-* `linux` | `darwin` | `osx` | `nuttx`
 
-Specify target OS.
+--
```
daeyeon: Seemingly `--` in the previous doc was intended for `---`. Why don't we fix them this time?

docs/build/Build-Script.md
```diff
@@ -130,123 +145,170 @@ If you have multiple link flags, supply it with multiple use of this option;
...
 
 --
-#### `--nuttx-home`
-To build for nuttx os, nuttx home directory must be given.
+#### `--jerry-heap-section`
+Specify the name of the JerryScript heap section.
 
 ```
-./tools/build.py --target-os=nuttx --target-arch=arm --target-board=stm32f4dis --nuttx-home="..."
+./tools/build.py --jerry-heaplimit=80
```
daeyeon: this needs to be for `--jerry-heap-section`

docs/Getting-Started.md
```diff
@@ -1,24 +1,51 @@
...
+  tools/build.py --clean
+```
+
+#### Frequently used build options
+
+`--clean` Clean build directory before build (default: False).
+
+`--no-snapshot` Disable snapshot generation for IoT.js. It is useful for debugging sessions.
+
+`--profile PROFILE` Specify the module profile file for IoT.js. It is used for enable and disable modules. See also ["How to write a new module"](devs/Writing-New-Module.md#Profile)
```
hs0225: #Profile -> #profile

docs/build/Build-Script.md
```diff
@@ -48,76 +48,91 @@ If given path is not exist, build.py will create it.
...
+./tools/build.py --config=build.arm.nuttx.stm32f4dis.config
 ```
 
---
-#### `--target-board`
-* `stm32f4dis` | empty
+`build.default.config` file is in the source tree for default setting.
 
-Specify target board.
+If this option is not specified, `build.config` file will be applied. If the file is not exist, it will be copied from `build.default.config`.
```
hs0225: is not exist -> does not exist

docs/build/Build-Script.md
```diff
@@ -130,123 +145,170 @@ If you have multiple link flags, supply it with multiple use of this option;
...
+./tools/build.py --jerry-memstat
 ```
 
-`build.default.config` file is in the source tree for default setting.
+--
+#### `--jerry-profile`
+* `es5.1` | `es2015-subset`
 
-If this option is not specified, `build.config` file will be applied. If the file is not exist, it will be copied from `build.default.config`.
+Specify the profile for JerryScript (default: es5.1).
```
daeyeon: Can you add an example of how to specify details of the profiles? e.g) how to selectively enable `arrow function` and `template literals` among the es2015-subsets supported. I think it would be helpful.

LaszloLango: Hmm, I am a bit surprised about this. I did not know how badly this feature is documented in both IoT.js and JerryScript. Furthermore the `build.py` of IoT.js only supports the built-in profiles. It should work with custom profiles too, but if you give it an absolute path (e.g.: `tools/build.py --clean --jerry-profile=/home/iotjs/custom.profile`), then it does not work. Output: `build.py: error: argument --jerry-profile: invalid choice: '/home/iotjs/custom.profile' (choose from 'es5.1', 'es2015-subset')`

So the build script must be updated as well. I would like to do it in a followup patch if it is good for you.

PR #1624, opened 2018-05-03
Followup patch for the MQTT Client

DanielBallaSZTE:
The followup patch contains the following changes:
- Fix for received messages with QoS level 0
- Support TLS as a socket to ensure secure communication possibility aswell.
- Add packet buffering, therefore messages sliced into multiple packets by the lowlevel network layer will be processed correctly.


src/modules/iotjs_module_mqtt.h
```diff
@@ -127,4 +127,11 @@ enum {
   MQTT_FLAG_USERNAME = 1 << 7
 } iotjs_mqtt_connect_flag_t;
 
+typedef struct {
+  jerry_value_t jobject;
```
zherczeg: Do we need this member? It seems never used.

src/modules/iotjs_module_mqtt.c
```diff
@@ -349,6 +380,138 @@ JS_FUNCTION(MqttPublish) {
...
+  iotjs_bufferwrap_t *buff_recv = iotjs_bufferwrap_from_jbuffer(jbuffer);
+  IOTJS_ASSERT(buff_recv->length != 0);
+
+  uint8_t first_byte;
+  char *tmp_ptr;
+  if (!mqttclient->buffer_length) {
+    // The local buffer is empty therefore either full packages arrived
+    // or this is the first packet.
+    first_byte = (buff_recv->buffer[0] >> 4) & 0x0F;
+    tmp_ptr = buff_recv->buffer;
```
zherczeg: What happens if mqttclient->buffer only contains a single byte?

I would see the following options:
1) Always do concatenation first.
2) Create a byte reader function, which virtually concatenates the two buffers, and returns a byte provided by an index.

The 2nd uses less memory, but the 1st is easier to implement. Which one you prefer?

DanielBallaSZTE:
@zherczeg Thank you for the suggestions! My approach to solve this problem was to check if the packet size is one. If so, we just concatenate the two buffers. However, there was still one issue, if only a part of `remaining length` has arrived. If so, we need to wait for the next packet and must concatenate it with the `local_buffer` in the next function call.

src/js/mqtt.js
```diff
@@ -75,6 +78,9 @@ MQTTClient.prototype.connect = function(callback) {
     this._socket.on('connect', function() {
       MqttConnect(this, jsref._clientOptions);
     });
+  } else if (Object.keys(tls).length != 0 &&
```
zherczeg: `"TLSSocket" in tls`
Btw what happens if tls is not enabled?

src/modules/iotjs_module_mqtt.c
```diff
@@ -24,26 +24,23 @@
 #include "iotjs_handlewrap.h"
 #include "iotjs_reqwrap.h"
 
+static void iotjs_mqttclient_destroy(iotjs_mqttclient_t *mqttclient) {
+  IOTJS_RELEASE(mqttclient->local_buffer);
```
zherczeg: mqttclient->local_buffer can be NULL, can't be?

src/modules/iotjs_module_mqtt.c
```diff
@@ -79,18 +76,34 @@ static uint32_t iotjs_decode_remaining_length(char *buffer, size_t *offset) {
   uint32_t remaining_length = 0;
   uint32_t length = 0;
   uint32_t shift = 0;
+  size_t local_offset = 0;
```
zherczeg: `size_t local_offset = (offset != NULL) ? *offset : 0;`

src/modules/iotjs_module_mqtt.c
```diff
@@ -79,18 +76,34 @@ static uint32_t iotjs_decode_remaining_length(char *buffer, size_t *offset) {
...
+  if (offset) {
+    local_offset = *offset;
+  }
+  // The size of the remaining length to determine if we received only
+  // a fragment of it in a packet.
+  size_t rem_len_size = strlen(buffer) - local_offset;
 
   do {
+    if (!rem_len_size) {
```
zherczeg: I would prefer `if (rem_len_size == 0) {`

src/modules/iotjs_module_mqtt.c
```diff
@@ -565,8 +587,175 @@ JS_FUNCTION(MqttHandle) {
...
+static void iotjs_mqtt_concat_buffers(iotjs_mqttclient_t *mqttclient,
+                                      iotjs_bufferwrap_t *buff_recv) {
+  char *tmp_buf = mqttclient->local_buffer;
+  mqttclient->local_buffer = (char *)malloc(
+      sizeof(char) * (buff_recv->length + mqttclient->buffer_length));
+  memcpy(mqttclient->local_buffer, tmp_buf, mqttclient->buffer_length);
+  memcpy(mqttclient->local_buffer + mqttclient->buffer_length,
+         buff_recv->buffer, buff_recv->length);
+  mqttclient->buffer_length += buff_recv->length;
+  free(tmp_buf);
```
zherczeg: we use IOTJS_RELEASE elsewhere. Perhaps the malloc should be replaced as well.

src/modules/iotjs_module_mqtt.c
```diff
@@ -565,8 +587,175 @@ JS_FUNCTION(MqttHandle) {
...
+      return JS_CREATE_ERROR(COMMON, "MQTT: Unknown error");
+  }
+}
+
+
+JS_FUNCTION(MqttReceive) {
+  DJS_CHECK_THIS();
+  DJS_CHECK_ARGS(2, object, object);
+
+  jerry_value_t jnat = JS_GET_ARG(0, object);
```
zherczeg: Cannot we receive this through `this`?

src/modules/iotjs_module_mqtt.c
```diff
@@ -565,8 +587,175 @@ JS_FUNCTION(MqttHandle) {
...
+  bool has_p =
+      jerry_get_object_native_pointer(jnat, &native_p, &out_native_info);
+  if (!has_p || out_native_info != &mqttclient_native_info) {
+    return JS_CREATE_ERROR(COMMON, "MQTT native pointer not available");
+  }
+  iotjs_mqttclient_t *mqttclient = (iotjs_mqttclient_t *)native_p;
+
+  jerry_value_t jbuffer = JS_GET_ARG(1, object);
+  iotjs_bufferwrap_t *buff_recv = iotjs_bufferwrap_from_jbuffer(jbuffer);
+  IOTJS_ASSERT(buff_recv->length != 0);
```
zherczeg: I would simply return here.

src/modules/iotjs_module_mqtt.c
```diff
@@ -565,8 +587,175 @@ JS_FUNCTION(MqttHandle) {
...
+
+  // Concat the buffers if we previously needed to
+  if (mqttclient->must_concat) {
+    iotjs_mqtt_concat_buffers(mqttclient, buff_recv);
+    mqttclient->must_concat = false;
```
zherczeg: I am not sure I understand the purpose of `must_concat`/ If the local buffer is not empty, we always need to do a concat first.

src/modules/iotjs_module_mqtt.c
```diff
@@ -565,8 +587,175 @@ JS_FUNCTION(MqttHandle) {
...
+
+  return 0;
+}
+
+
+static void iotjs_mqtt_concat_buffers(iotjs_mqttclient_t *mqttclient,
+                                      iotjs_bufferwrap_t *buff_recv) {
+  char *tmp_buf = mqttclient->local_buffer;
+  mqttclient->local_buffer = (char *)malloc(
+      sizeof(char) * (buff_recv->length + mqttclient->buffer_length));
```
zherczeg: the order is `mqttclient->buffer_length + buff_recv->length`. I would assign it to a local variable and just write 
`mqttclient->buffer_length = local_variable` below. Assign is faster tham +=

src/modules/iotjs_module_mqtt.c
```diff
@@ -565,8 +587,175 @@ JS_FUNCTION(MqttHandle) {
...
+  // Only 1 byte arrived, the local buffer is empty, we can't get remaining
+  // length, just concat the receive buffer on the next packet
+  if (!mqttclient->buffer_length && buff_recv->length == 1) {
+    mqttclient->must_concat = true;
+    return jerry_create_undefined();
+  }
+
+  uint32_t packet_size = 0;
+  packet_size = iotjs_decode_remaining_length(buff_recv->buffer + 1, NULL) + 2;
+  if (!packet_size) {
```
zherczeg: What happens on error?

src/modules/iotjs_module_mqtt.c
```diff
@@ -565,8 +587,175 @@ JS_FUNCTION(MqttHandle) {
...
+
+  // Keep on going, if data remains just continue looping
+  while (mqttclient->buffer_length + buff_recv->length >= packet_size) {
+    int ret_val = 0;
+    jerry_value_t jdata = iotjs_bufferwrap_create_buffer(packet_size);
+    iotjs_bufferwrap_t *data = iotjs_bufferwrap_from_jbuffer(jdata);
+
+    if (!mqttclient->buffer_length) {
+      memcpy(data->buffer, buff_recv->buffer, packet_size);
+      if ((ret_val = iotjs_mqtt_handle(jnat, data))) {
```
zherczeg: Cannot we directly pass an uint8_t* pointer and a size rather than creating a buffer?

DanielBallaSZTE:
@zherczeg Thank you ! I have updated the PR!

src/modules/iotjs_module_mqtt.c
```diff
@@ -74,24 +69,30 @@ static size_t get_remaining_length_size(uint32_t len) {
...
 
-static uint32_t iotjs_decode_remaining_length(char *buffer, size_t *offset) {
-  unsigned char c;
-  uint32_t remaining_length = 0;
+static uint32_t iotjs_decode_remaining_length(char *buffer,
+                                              uint32_t *out_length) {
+  uint32_t decoded_length = 0;
   uint32_t length = 0;
   uint32_t shift = 0;
+  uint32_t c;
```
zherczeg: Another idea: since we know that at least two bytes will be parsed, this could be reworked to:
```
uint32_t c = (uint32_t)(*buffer);
uint32_t decoded_length = (c & 0x7F);
uint32_t shift = 7;
uint32_t length = 1;

buffer++;
```

src/modules/iotjs_module_mqtt.c
```diff
@@ -74,24 +69,30 @@ static size_t get_remaining_length_size(uint32_t len) {
...
+    decoded_length += (c & 0x7F) << shift;
+
     shift += 7;
+    buffer++;
+  } while ((c & 0x80) != 0);
 
-    (*offset)++;
-  } while ((c & 128) != 0);
+  if (c == 0) {
+    return 0;
```
zherczeg: Add comment:
// The length must be encoded with the least amount of bytes.
// This rule is not fulfilled if the last byte is zero.

src/modules/iotjs_module_mqtt.c
zherczeg: Keep it now. Btw it wouldn't be bad to implement a CHECK_THIS like macro for first argument as well, some tls code could be reworked to be more private as well.

src/modules/iotjs_module_mqtt.c
DanielBallaSZTE: Great idea!

yichoi:
LGTM

PR #1163, opened 2017-08-31
[WIP] Use correct headers when building jerry and libtuv

pmarcinkiew:
This patch fixes previous issue with build system not using
correct TizenRT path, thus creating completely broken build.

IoT.js-DCO-1.0-Signed-off-by: Jaroslaw Pelczar j.pelczar@samsung.com

PR #1628, opened 2018-05-07
Use correct headers when building jerry and libtuv

rtakacs:
Pass the external include directories to the subprojects as well. This patch is
based on #1163 that was closed before merge.


CMakeLists.txt
```diff
@@ -171,6 +171,18 @@ set(ARCHIVE_DIR ${CMAKE_BINARY_DIR}/lib)
 
 include(ExternalProject)
 
+set(TARGET_BUILD_C_FLAGS "${CMAKE_C_FLAGS}")
+
+if(NOT ("${EXTERNAL_LIBC_INTERFACE}" STREQUAL ""))
+	if("${TARGET_OS}" MATCHES "TIZENRT|NUTTX")
```
LaszloLango: Please change tabs to spaces.

rtakacs:
@LaszloLango Thanks, I've updated the patch.

rtakacs:
Note: `EXTERNAL_LIBC_INTERFACE` is not used in Jerry, so that can be removed. In this patch, the `isystem` option is added to the `CMAKE_C_FLAGS`, that is passed to jerry and libtuv as well:
  * https://github.com/Samsung/iotjs/blob/master/cmake/jerry.cmake#L119
  * https://github.com/Samsung/iotjs/blob/master/cmake/libtuv.cmake#L36

yichoi:
LGTM

PR #1630, opened 2018-05-07
Change iotjs_jval_as_string to use utf8 buffers

### LINKED ISSUE #1562: console.log() does not print out using UTF-8 encoding
As can be seen here in this video: https://www.youtube.com/watch?v=Xe9r9maWmdI

The sample code `hello.js` contains:
```
console.log('Hello, world!');
console.log('This is a demo of the WIP JerryScript <=> Chrome Devtools debugger proxy!');
console.log('It works with stock JerryScript and iot.js');
console.log('🎉🦄✨');
console.log('kthnxbye!');
```

The emoji's aren't printed using UTF-8 encoding but using CESU-8 encoding instead.
AFAIK, UTF-8 has been the default string encoding for Linux terminals. (CESU-8 is probably not even an option...?)

DanielBallaSZTE:
Fixes #1562
With this patch strings coming from JS side are handled as UTF8 strings.

### DIFF
--- a/src/iotjs_binding.c
+++ b/src/iotjs_binding.c
@@ -126,7 +126,7 @@ bool iotjs_jbuffer_as_string(jerry_value_t jval, iotjs_string_t* out_string) {
 iotjs_string_t iotjs_jval_as_string(jerry_value_t jval) {
   IOTJS_ASSERT(jerry_value_is_string(jval));
 
-  jerry_size_t size = jerry_get_string_size(jval);
+  jerry_size_t size = jerry_get_utf8_string_size(jval);
 
   if (size == 0)
     return iotjs_string_create();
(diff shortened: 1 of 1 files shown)


LaszloLango:
I'm not sure about this is the correct fix for the mentioned issue. I think it would be better to change `jerry_string_to_char_buffer` to `jerry_string_to_utf8_char_buffer` in `iotjs_jval_as_string`.

DanielBallaSZTE:
@LaszloLango Yeah but that would mean any string in IoT.js would turn into UTF8 strings.

zherczeg:
Question is what we want. Do we need utf8 in iotjs or cesu8? I think most strings are ascii, and there are fast paths for them.

zherczeg:
(Btw not all js strings can be converted to valid utf8)

yorkie:
+1 on @LaszloLango's suggestion, that would be a common string problems in IoT.js :p

DanielBallaSZTE:
Thank you all for the suggestions, I have updated the Pull Request.

yichoi:
LGTM

PR #1631, opened 2018-05-08
Add error handling of TLS

haesik:
'data' will be passed to the iotjs_bio_write even if it is NULL


zherczeg:
I think passing NULL is used during the handshake, which goes through the Read function. Since length is 0, passing a NULL to iotjs_bio_write should not cause any issues.

haesik:
Oh, I am sorry, I mistaken.

PR #1634, opened 2018-05-09
Add jerryscript profile for Tizen

daeyeon:


test/profiles/tizen.profile
```diff
@@ -9,3 +9,8 @@ ENABLE_MODULE_PWM
 ENABLE_MODULE_SPI
 ENABLE_MODULE_TIZEN
 ENABLE_MODULE_UART
+
+# Jerryscript Profile
```
LaszloLango: IMHO this should be placed into a new file even if this work now.

daeyeon: I understand your concern. A new file is added.

PR #1637, opened 2018-05-11
Fix the 'end' event of TLSSocket when it has a stream.Duplex

pmarkee:


pmarkee:
It is tested, it was added in the previous patch #1585

pmarkee:
@LaszloLango thanks for the review, I updated the PR according to your suggestions and also renamed test_tls_2.js to test_tls_stream_duplex.js

yichoi:
LGTM

PR #1638, opened 2018-05-11
Move iotjs_build_info.js to the test folder

rtakacs:
Moved the `iotjs_build_info.js` to the `test/tools` folder because it logically belongs to that folder and the script is created just for the testrunner.

Since js-remote-test copies the `test` folder to the devices, it would be nice if the script could be in one place with all the tests.


yichoi:
LGTM

Issue #1595, opened 2018-04-20
Proposal of the test set descriptor update


LaszloLango:
I'd like to propose an update for `testsets.json`. You sad before that not everyone happy with current situation with extra testsets files which can update the default settings. That modification was necessary to test different configurations. This is a proposal for the expected `testsets.jons`, if we want to use one single json descriptor file again (before #1341). With this modification we can keep the ability to test different configurations with a single json.

```
{
  "directory": [
    { "name": "filename",
      "skip": ["all"],
      "reason": "reason of skipping",
      "timeout": seconds,
      "expected-failure": true,
      "required-modules": ["my_module"],
      "required-features": ["es-262-feature"]
    },
    ...
  ]
}
```

 - _directory_: group of tests
 - _name_: filename = testname
 - _skip_: platform where the test must be skipped. ["all", "darwin", "linux", "nuttx", "tizen", "tizenrt"] **(optional)**
 - _reason_: it belongs to skip property, reason of skipping. **(optional)**
 - _timeout_: timeout in seconds **(optional)**
 - _expected-failure_: identifies the "must fail" testcases. Still catches segfaults, IOTJS_ASSERT and JERRY_ASSERT. Default: false [true, false]  **(optional)**
 - _required-modules_: List of the required modules. Skips test if the module is not present int the current build. **(optional)**
 - _required-features_: List of the required language features. eg.: Promise **(optional)**

**Note**:
  - _"required-modules"_ and _"required-features"_ are compared with the output of `tools/iotjs_build_info.js` which is currently used by the existing testrunners. It must be extended with the language related features, like Promis, ArrayBuffer, etc.

What do you think?

CC: @daeyeon, @hs0225, @galpeter

hs0225:
@LaszloLango I agree your proposal. I look better to combine the test set file into one.

PR #1643, opened 2018-05-15
Update testsets.json and the testrunner

pmarkee:
Added the optional required-modules and required-features properties to the test cases in testsets.json. This way we can test different configurations with a single json file. See #1595 


pmarkee:
@galpeter @LaszloLango thanks for your suggestions, I updated the PR

test/tools/iotjs_build_info.js
```diff
@@ -21,8 +21,45 @@ if (process.env.IOTJS_ENV.indexOf("experimental") > -1)
 else
     stability = "stable"
 
+/* Check if certain es2015 features are available */
+var features = {}
+
+var hasPromise = (typeof Promise) != 'undefined';
```
LaszloLango: Wouldn't it be much simpler if `features` will be an array?

e.g.:
```js
var features = [];
if ((typeof Promise) != 'undefined') {
  features.push('Promise');
}
```

pmarkee: @LaszloLango it could be an array, I just wanted to keep the output format consistent. The ``builtins`` property is an object, so I thought ``features`` should be an object as well, but I can change it if necessary

LaszloLango: It really depends on the python part (`testrunner.py`), so I am curious about the opinion of @galpeter. I saw in a previous review comment that we could use function of `set` in python like `contains`. I think if it is an array, than it it will be parsed as a set, but I am not sure.

tools/testrunner.py
```diff
@@ -273,32 +255,42 @@ def run_test(self, testfile, timeout):
...
+        unsupported_features = required_features - self.features
+        skipped_modules = required_modules.intersection(skip_list)
 
-        tested_module = name_parts[1]
+        # Skip the test if the tested module requires a module
+        # which is not compiled into the binary
+        if unsupported_modules:
+            test["reason"] = "Required module(s) \""
+            test["reason"] += '\", \"'.join(sorted(unsupported_modules))
+            test["reason"] += "\" unsupported by iotjs build"
```
galpeter: What if we use this output format?: `Required module(s) unsupported by build: tls, kkk,`

pmarkee: Yes, maybe that will be a better format

yichoi:
LGTM

PR #1647, opened 2018-05-18
Build module_entry.c test-resource into different target folders.

rtakacs:
This patch adds some missing functionalities for the `dynamicmodule.iotjs` test-resource file:

  * Support `--clean` build option to remove and re-compile that shared library.
  * Add the created shared library to the `.gitignore` file.
  * Create the `dynamicmodule.iotjs` into different target folders (e.g. `test/dynamicmodule/build/linux` `test/dynamicmodule/build/tizen` `test/dynamicmodule/build/nuttx` ...) because there could be different toolchains for the targets.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,7 @@
 /src/iotjs_string_ext.inl.h
 /src/iotjs_module_inl.h
 /test/tmp/*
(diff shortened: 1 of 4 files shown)


yichoi:
LGTM

PR #1648, opened 2018-05-22
Fix a typo in the MQTT module

DanielBallaSZTE:
Fixing a typo in the mqtt module, and move a check to return earlier if fails.


yichoi:
LGTM

PR #1145, opened 2017-08-25
build: Install headers to system

rzr:
make install will install headers to iotjs headers namespace

IoT.js-DCO-1.0-Signed-off-by: Philippe Coval philippe.coval@osg.samsung.com

PR #1649, opened 2018-05-22
build: Install headers to system

rzr:
make install will install headers to iotjs headers namespace.
This is needed for build developer packages,
on debian based system like raspbian for pi0.

This change was applied to debian packaging for supporting RPI.

Bug: https://github.com/Samsung/iotjs/pull/1145

### DIFF
--- a/cmake/iotjs.cmake
+++ b/cmake/iotjs.cmake
@@ -488,6 +488,13 @@ endif()
 
 install(TARGETS ${TARGET_STATIC_IOTJS} DESTINATION ${LIB_INSTALL_DIR})
 
+# Install headers
+if("${INCLUDE_INSTALL_DIR}" STREQUAL "")
+  set(INCLUDE_INSTALL_DIR "include/iotjs")
+endif()
+file(GLOB IOTJS_HEADERS include/*.h)
+install(FILES ${IOTJS_HEADERS} DESTINATION ${INCLUDE_INSTALL_DIR})
+
 # Configure the libiotjs.so
 if (NOT BUILD_LIB_ONLY AND CREATE_SHARED_LIB)
   set(TARGET_SHARED_IOTJS shared_iotjs)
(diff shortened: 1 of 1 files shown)


cmake/iotjs.cmake
```diff
@@ -508,6 +508,14 @@ if (NOT BUILD_LIB_ONLY AND CREATE_SHARED_LIB)
     ${EXTERNAL_LIBS})
 endif()
 
+# Install headers
+if("${INCLUDE_INSTALL_DIR}" STREQUAL "")
```
daeyeon: It seems better to put the relevant codes in one place. I suggest moving this code block to the line, 488.

cmake/iotjs.cmake
```diff
@@ -508,6 +508,14 @@ if (NOT BUILD_LIB_ONLY AND CREATE_SHARED_LIB)
...
+  set(INCLUDE_INSTALL_DIR "include/iotjs")
+endif()
+install(TARGETS ${TARGET_LIB_IOTJS} DESTINATION ${LIB_INSTALL_DIR})
```
daeyeon: isn't TARGET_LIB_IOTJS undefined?

cmake/iotjs.cmake
```diff
@@ -508,6 +508,14 @@ if (NOT BUILD_LIB_ONLY AND CREATE_SHARED_LIB)
...
+file(GLOB IOTJS_HEADERS include/*.js src/*.h)
```
daeyeon: a) include/*.js -> include/*.h?
b) why is src/*.h needed?

cmake/iotjs.cmake
```diff
@@ -506,6 +513,7 @@ if (NOT BUILD_LIB_ONLY AND CREATE_SHARED_LIB)
     ${MBEDTLS_LIBS}
     -Wl,--no-whole-archive
     ${EXTERNAL_LIBS})
+install(TARGETS ${TARGET_SHARED_IOTJS} DESTINATION ${LIB_INSTALL_DIR})
```
daeyeon: wrong indentation

PR #1656, opened 2018-05-28
Rework MQTT to use less memory.

zherczeg:
Less bind() calls, less properties, better context hiding.

Furthermore add a test for fragmented packets.


src/js/mqtt.js
```diff
@@ -332,7 +296,8 @@ function getClient(connectOptions) {
       connectOptions.message = new Buffer(connectOptions.message.toString());
     }
   }
-  return new MQTTClient(connectOptions);
+
+  return new MQTTClient(connectOptions, callback);
 }
 
-exports.getClient = getClient;
+exports.connect = connect;
```
galpeter: This changes the API, thus the documentation should also be updated.

zherczeg: Done.

yichoi:
LGTM

PR #1658, opened 2018-05-30
Correctly display test outputs in the testrunner when using Python3

galpeter:
The data returned by the subprocess calls in the testrunner
is a byte-sequence on Python3. Byte-sequences have a
b'..' format when directly printed out. This change
will convert all output to utf-8 so there will be no
extra characters displayed.


yichoi:
LGTM

PR #1665, opened 2018-06-01
Improve MQTT interface to follow other implementations and improve security.

zherczeg:
A message storage is also added to keep outgoing packages.


DanielBallaSZTE:
Good patch!
I especially like the idea of temporary buffers !
One thing I'm unsure about is do we actually properly close the underlying network connection whenever an error occurs?
Since in most of the cases the standard asks us to do so.

zherczeg:
Thnak you.  The `MQTTHandle.prototype.onEnd` should be called when abort is required.

daeyeon:
Please turn ENABLE_MODULE_MQTT on for the Linux case on Travis.

zherczeg:
@daeyeon I have enabled the mqtt module, but it seems mqtt://test.mosquitto.org randomly refuses the connection. Unfortunately it happens frequently.

Error: connect failed - status: ECONNREFUSED

It is probably a DOS attack protection, but it makes the test_mqtt.js test flaky. A long term solution could be setting up a local mosquitto server on the Travis, but that is not the focus of this patch. What shall we do? Enable the mqtt module or still keep it disabled?

DanielBallaSZTE:
For me, I would have it disabled this patch, and the next patch where we enable the mosquitto server on travis we turn on MQTT as well.

daeyeon:
@zherczeg @DanielBallaSZTE Thank you for the clarification. It seems right to keep it disabled for this patch.

PR #1668, opened 2018-06-04
Use correct file stat macros on Windows

galpeter:
The file stat macros are defined in a different name on Windows.
Adding a few compatibility defines in case of Windows build.


zherczeg:
```
 #ifndef S_ISREG
-#define S_ISREG(mode) (((mode) & S_IFMT) == S_IFREG)
+#define S_ISREG(mode) (((mode)&S_IFMT) == S_IFREG)
 #endif
```
this style error looks invalid. what confused the style checker?

PR #1669, opened 2018-06-04
Add Windows support

galpeter:
Main points of the change:
* With Visual Studio the configuration type must be specified
  when searching for the built libraries.
* New cmake variable `CONFIG_TYPE` will be set to Debug/Release depending
  on VS configuration selection.
* New cmake variable `USING_MSVC`.
* New macro `build_lib_name` Windows and Linux builds the libraries in
  different names. The macro helps to build the libraries name.
* Updated library names.
* Added MSVC specific arguments when the magic strings are extracted
  from the `iotjs_magic_strings.h` file.
* Updated build scripts.
* Updated a few tests with windows specific parts.
* Improved module path resolving to be usable with Windows paths.
* DNS resolving selects the first IPv4 address if possible.
* Added Windows build documentation.

Notes:
* There are tests where the expected result requires the '\n' line ending.
  However on Windows the line ending is '\r\n'.
* After a JerryScript update the snapshot mode could be enabled.
* On Windows the build is now a two step process:
   1) Call the build.py to generate the VS solution file.
   2) Open the solution file with VS and build.


galpeter:
Related libtuv PR: https://github.com/Samsung/libtuv/pull/112

yichoi:
Please do set up appvoyer in advance.

galpeter:
I've updated the PR to contain the AppVeyor config also. I'm not sure if I have enough credentials to enable AppVeyor for this project.

yichoi:
@daeyeon please enable the AppVeyor

daeyeon:
@galpeter @yichoi AppVeyor's now been enabled.

galpeter:
ohh btw: the appveyor build will always create an artifact containing the iotjs.exe binary.

PR #1677, opened 2018-06-11
Remove 'iotjs_jhelper_call' function from bindings

LaszloLango:

### DIFF
--- a/src/iotjs_binding.c
+++ b/src/iotjs_binding.c
@@ -22,9 +22,6 @@
 #include <string.h>
 
 
-static iotjs_jargs_t jargs_empty;
-
-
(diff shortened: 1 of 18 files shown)


LaszloLango:
The patch reduces the binary size by 1 Kbyte.

zherczeg:
Nice!

daeyeon:
I'm worried the side effect for modules not verified on Travis. How about breaking these into smaller patches? e.g) verified group and not-verified one

LaszloLango:
If you like this change and this is a good direction then I can divide the patch to smaller ones.

daeyeon:
@LaszloLango We've decided to agree with your idea because of the size benefit. The previous one seems somewhat handier though.

PR #1688, opened 2018-06-14
build: Add fallback linux config for unsupported arch

rzr:
This rule is helpful for non supported arch,
to rely on system compiler with default flags.

For instance debian uses i386 arch (while upstream supports only i686),
this could also apply to other non supported archs.

For the record most debian arch were building on 1.0 (not ia64, sh4):

https://buildd.debian.org/status/package.php?p=iotjs&suite=unstable

### DIFF
--- a/cmake/config/noarch-linux.cmake
+++ b/cmake/config/noarch-linux.cmake
@@ -1,0 +1,15 @@
+# Copyright 2018-present Samsung Electronics Co., Ltd. and other contributors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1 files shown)


rzr:
well I rebased  my single commit on master, what you saw above is just think to this patches in other branches (github reviews are a bit missleading sometime).

If curious I used this for:
https://s-opensource.org/2018/06/21/webthing-iotjs/

yichoi:
@rzr, of course, why not? if possible, could you please add the endorsement of our team? :)

Issue #1685, opened 2018-06-13
Remove 'iotjs_jargs_t' from bindings to reduce binary size.


LaszloLango:
Proof of concept patch was presented in #1677. Removal of `iotjs_jargs_t` reduces the binary size by 1 Kbyte. The removal will be done in multiple PRs as it was discussed in the comments of #1677.

**Tasks**:

 - [x] Remove `iotjs_jhelper_call` function

Change the `iotjs_make_callback` calls in
 - [x] basic modules
 - [x] core modules
 - [x] peripheral modules
 - [x] other modules (bridge, tizen, etc.)

LaszloLango:
Closing because if was finished in #1704

PR #1692, opened 2018-06-20
Change 'iotjs_make_callback' calls

LaszloLango:
Change 'iotjs_make_callback' calls to 'iotjs_invoke_callback' which
does not use 'iotjs_jargs_t'. Basic and core modules are touched in
this patch. Related issue: #1685


src/modules/iotjs_module_http_parser.c
```diff
@@ -131,19 +131,20 @@ static void iotjs_http_parserwrap_flush(
       iotjs_jval_get_property(jobj, IOTJS_MAGIC_STRING_ONHEADERS);
   IOTJS_ASSERT(jerry_value_is_function(func));
 
-  iotjs_jargs_t argv = iotjs_jargs_create(2);
   jerry_value_t jheader = iotjs_http_parserwrap_make_header(http_parserwrap);
-  iotjs_jargs_append_jval(&argv, jheader);
-  jerry_release_value(jheader);
+  size_t argc = 1;
+  jerry_value_t argv[2] = { jheader, jerry_create_null() };
```
daeyeon: There seem two approaches to this.

a) explicitly adding `jerry_null` into the `argv`before giving it to `iotjs_invoke_callback`.
b) verifying the given `argv` inside `iotjs_invoke_callback` and auto-filling `jerry_null` if any item is `null`.

What would be better?

zherczeg: There could be an init function, but for such a small amount of values, its cost is higher than its worth.

The auto fill will probably not work, because argv has jerry_value_t, and its values are random, not pre-defined, since it is a local value allocated on the stack.

daeyeon: In this case, the values can be pre-defined as 0 like `jerry_value_t argv[2] = { jheader };`. For all zero items, `jerry_value_t argv[2] = {};`. It may be simpler than the current way.

LaszloLango: `jerry_value_t argv[2] = {};` does not guarantee the values will be `0`. It is allocated on the stack so it may contain non-zero values too.

daeyeon: At least, it's allowed in C++, as I remember, but not sure for C standard. If not, we could do `jerry_value_t argv[2] = {0};` for initializing all elements, IMO.

LaszloLango:
I've updated the PR.

PR #1697, opened 2018-06-22
Fix for setHeader

daeyeon:
This patch fixes the error from the following case:

```js
var str = '';
res.setHeader('Content-Length', str.length);
```


LaszloLango:
Could you add the mentioned testcase also?

daeyeon:
@LaszloLango I've added it.

PR #1698, opened 2018-06-26
samples: Add hello sample to test console

rzr:
This is the simplest example, that can be used for smoketest.


daeyeon:
- For smoke test, using the test suite should be our guide.
- We have https://github.com/Samsung/iotjs/blob/master/test/run_pass/test_console.js which looks similar.

PR #1700, opened 2018-06-27
Update TLS certificates to extend their lifetime

DanielBallaSZTE:


DanielBallaSZTE:
@zherczeg 365 days

yichoi:
How about writing a document a certificate with openCA and others ?

DanielBallaSZTE:
@yichoi You can find those in the `test_tls_ca.js` file, in the comment section.

yichoi:
@DanielBallaSZTE OK. 'test_tls_ca.js' is fine for me. Somewhat not fully satisfied with me since some tests related to TLS would be failed after next year which you have prolonged

DanielBallaSZTE:
@yichoi You are right. But in my opinion it is very easy to renew those files, doesn't even take about 5 minutes, following the steps written in the test file. :)

zherczeg:
What about extending it to 10 years? These are just test files.

DanielBallaSZTE:
@zherczeg Yeah that's another possibility, I could do it as well.

DanielBallaSZTE:
@zherczeg @yichoi I have updated the certs to be valid for 10 more years.

yichoi:
LGTM.
You guys think our project would be perished in 10 years :)

PR #1706, opened 2018-07-05
Fix exitcode returning 0 in some cases

DanielBallaSZTE:
There are some special cases, where the exitcode should return an error, however it returns with 0.
This patch fixes it.
Related test: test-issue-1570.js

I could add more tests, however all of them would be in seperate files, and those would mean plus 3-4 tests.

### DIFF
--- a/src/iotjs_binding_helper.c
+++ b/src/iotjs_binding_helper.c
@@ -145,6 +145,11 @@ int iotjs_process_exitcode() {
   } else {
     exitcode = (uint8_t)iotjs_jval_as_number(num_val);
   }
+
+  uint8_t native_exitcode = iotjs_environment_get()->exitcode;
+  if (native_exitcode != exitcode && native_exitcode) {
+    exitcode = native_exitcode;
+  }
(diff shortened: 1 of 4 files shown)


src/iotjs_binding_helper.c
```diff
@@ -145,6 +145,11 @@ int iotjs_process_exitcode() {
   } else {
     exitcode = (uint8_t)iotjs_jval_as_number(num_val);
   }
+
+  uint8_t native_exitcode = iotjs_environment_get()->exitcode;
+  if (native_exitcode != exitcode && native_exitcode) {
+    exitcode = iotjs_environment_get()->exitcode;
```
rtakacs: You can use the `native_exitcode` instead of the `iotjs_environment_get()->exitcode`.

DanielBallaSZTE:
@rtakacs Thank you for the review!

src/iotjs_binding_helper.c
DanielBallaSZTE: Right, thanks

PR #1707, opened 2018-07-06
[testrunner] Tests with expected failure results should not return with exitcode 0

robertsipka:
Follow up patch for #1706. That patch needs to be landed first.


robertsipka:
The #1706 was landed. I've rebased it to the current master.

PR #1712, opened 2018-07-17
Change ../../libapps paths in NuttX configuration Makefile

Achie72:
With the old paths, IoT.js was built with NuttX paths, which won't work if we use symlinks instead of copying IoT.js files, because the path is not the same. With these changes, the old method still works, and application files are buildable both with symlink and cp method.


rtakacs:
LGTM (informally)

PR #1719, opened 2018-07-25
ignore: tmp and debian generated files

rzr:
tmp/ can be used for developers playground

Also ignore Debian generated files and objects if build in .

Bug: https://github.com/Samsung/iotjs/issues/1389

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,7 @@
 /test/tmp/*
 /test/dynamicmodule/build/*
 eslint.log
+/obj-*-*/
 
 # IDE related files
 nbproject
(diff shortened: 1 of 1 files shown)


hs0225:
@rzr We don't write anything to `/tmp` and there is no file in debian. If you want to ignore the file in git personally, modify `.git/info/exclude` file.

rzr:
well this is a change I have applied downstream, so I thought it could help others if merged upstream as it will not cause any regressions... feel free to close request if it is not desired, it's not a big deal.

hs0225:
@rzr This commit is not needed right now. so I'll close it.

PR #1721, opened 2018-08-03
Fix websocket client to support multiple clients at the same time.

imiklos:
The generated key for the handshake validation has moved to the header.


src/modules/iotjs_module_websocket.c
```diff
@@ -244,8 +253,17 @@ JS_FUNCTION(PrepareHandshakeRequest) {
...
+  JNativeInfoType *out_native_info;
+  bool has_p =
+    jerry_get_object_native_pointer(jsref, &native_p, &out_native_info);
+  if (!has_p || out_native_info != &wsclient_native_info) {
+    return WS_ERR_NATIVE_POINTER_ERR;
+  }
+  iotjs_wsclient_t *wsclient = (iotjs_wsclient_t *)native_p;
+
+  wsclient->generated_key = ws_generate_key(jsref);
+  size_t generated_key_len = strlen((char *)wsclient->generated_key);
```
DanielBallaSZTE: This might cause some issues, since the string is not null terminated.

imiklos:
@DanielBallaSZTE Thank you! I've already added to it, just haven't updated the PR, yet.

src/modules/iotjs_module_websocket.c
```diff
@@ -244,8 +255,17 @@ JS_FUNCTION(PrepareHandshakeRequest) {
...
+  JNativeInfoType *out_native_info;
+  bool has_p =
+      jerry_get_object_native_pointer(jsref, &native_p, &out_native_info);
+  if (!has_p || out_native_info != &wsclient_native_info) {
+    return WS_ERR_NATIVE_POINTER_ERR;
+  }
+  iotjs_wsclient_t *wsclient = (iotjs_wsclient_t *)native_p;
+
+  wsclient->generated_key = ws_generate_key(jsref);
+  size_t generated_key_len = 24;
```
daeyeon: I'm curious on if we can assume the length of the key will constantly be 24.

daeyeon: How about using `strnlen` if you can assume the maximum number of the key?

DanielBallaSZTE: Some reasons I think it's fine as it is:
1.: The `generated_key` ptr is not '\0' terminated, therefore if there are any valid characters behind it, it would get read by the function using the `generated_key_len`.
2.: The `ws_generate_key()` function always generates 16 random bytes, and base64 encodes them, which gives 24 constantly. The only exception is when the base64 fails, but in that case an exception is thrown, and execution is terminated.

daeyeon: In case of the reason of 2, is the exception thrown for the base64 failure? Or else, just `error` event is emitted? If iot.js isn't immediately terminated at the exception, segment fault is able to occur because of `memcpy` from the source, `wsclient->generated_key`, NULL.

BTW, Is base64 encoding in `ws_generate_key` function able to be failed? All the inputs seem always valid.

DanielBallaSZTE: Hmm you are right, IoT.js should be terminated, since it is an uncaught exception, but the user could just catch the error, and continue running the program. 
The base64 encoding can fail as well.
By the way, thank you for clarifying your point, now I get it. You suggestion is right on using  `strlen`.

imiklos: @daeyeon @DanielBallaSZTE Thank you! I've updated the PR.

src/modules/iotjs_module_websocket.c
```diff
@@ -110,14 +110,25 @@ static char *iotjs_ws_write_data(char *buff, void *data, size_t size) {
...
+  if (!has_p || out_native_info != &wsclient_native_info) {
+    return false;
+  }
+  iotjs_wsclient_t *wsclient = (iotjs_wsclient_t *)native_p;
+
   unsigned char *out_buff = NULL;
 
-  size_t concatenated_size = strlen(WS_GUID) + strlen((char *)generated_key);
+  size_t concatenated_size =
+      strlen(WS_GUID) + strlen((char *)wsclient->generated_key);
```
galpeter: We should use `strnlen` in case of the `generate_key` to maximize the key size and avoid buffer overruns.

src/modules/iotjs_module_websocket.c
```diff
@@ -110,14 +110,25 @@ static char *iotjs_ws_write_data(char *buff, void *data, size_t size) {
...
   unsigned char concatenated[concatenated_size + 1];
 
-  memcpy(concatenated, generated_key, strlen((char *)generated_key));
-  memcpy(concatenated + strlen((char *)generated_key), WS_GUID,
+  memcpy(concatenated, wsclient->generated_key,
+         strlen((char *)wsclient->generated_key));
+  memcpy(concatenated + strlen((char *)wsclient->generated_key), WS_GUID,
```
galpeter: Here the `strlen` is used two times, we should only do it once. (and also use the strnlen method)

galpeter: Ohh I've just noticed, it is called three times in this method. please extract it to a local variable.

src/modules/iotjs_module_websocket.c
```diff
@@ -243,8 +254,17 @@ JS_FUNCTION(PrepareHandshakeRequest) {
...
+  void *native_p;
+  JNativeInfoType *out_native_info;
+  bool has_p =
+      jerry_get_object_native_pointer(jsref, &native_p, &out_native_info);
+  if (!has_p || out_native_info != &wsclient_native_info) {
+    return WS_ERR_NATIVE_POINTER_ERR;
+  }
+  iotjs_wsclient_t *wsclient = (iotjs_wsclient_t *)native_p;
+
+  wsclient->generated_key = ws_generate_key(jsref);
```
galpeter: In the ws_generate_key, the base64 encode returns the buffer length, would it make sense to use directly that value? Maybe extend the `ws_generate_key` to accept an additional output parameter?

src/modules/iotjs_module_websocket.c
```diff
@@ -243,8 +254,17 @@ JS_FUNCTION(PrepareHandshakeRequest) {
...
   };
 
-  generated_key = ws_generate_key(jsref);
-  size_t generated_key_len = strlen((char *)generated_key);
+  void *native_p;
+  JNativeInfoType *out_native_info;
+  bool has_p =
+      jerry_get_object_native_pointer(jsref, &native_p, &out_native_info);
+  if (!has_p || out_native_info != &wsclient_native_info) {
+    return WS_ERR_NATIVE_POINTER_ERR;
```
galpeter: Shouldn't this be a JS_CREATE_ERROR call?

src/modules/iotjs_module_websocket.c
imiklos: Sure, good idea!

src/modules/iotjs_module_websocket.c
imiklos: Good point! It should be.

imiklos:
@galpeter Thank you for the review. I've updated the PR with the changes you requested.

PR #1730, opened 2018-08-24
Remove and replace iotjs_reqwrap_t struct

galpeter:
The `iotjs_reqwrap_t` was used to perform jobs on the uv loop in most cases.
Additionally from this struct there were a derivative reqwrap structs which contained a `iotjs_reqwrap_t` member and an `uv_req_t` member. In the `iotjs_reqwrap_t` there was a pointer to an `uv_req_t` which was set to the value of the req member provided by derivative reqwrap struct. Furthermore the data pointer of the `uv_req_t` was set to the derivative member. This caused a circular reference between the reqwrap structs.

These changes remove the `iotjs_reqwrap_t` and replace them with the direct usage of the `uv_req_t` types.

New methods:
* `iotjs_uv_request_create`
* `iotjs_uv_request_destroy`

New macros:
* `IOTJS_UV_REQUEST_JSCALLBACK`
* `IOTJS_UV_REQUEST_EXTRA_DATA`

For more details look into the `iotjs_uv_request.h` file.

Benefits:
* Reduced number of allocations.
* Minor code size reduction.
* Memory usage reduction.


haesik:
Please squash the commits into a single one.

Issue #1726, opened 2018-08-13
wish: version release


rzr:
Hi, 

Dear iotjs team,

Is there a new release planned soon ?
If not, may I suggested this.

The reason is that current master branch has really improved support for Tizen:Rt (which currently ship a snapshot before 1.0) since some downstream changes have been upstreamed.. so I guess it is valuable to upgrade snapshot in TizenRt.

Relate-to:

https://github.com/Samsung/TizenRT/pull/2018#issuecomment-412433508

Again this is just a suggestion, I don't want to change any plans.

https://github.com/Samsung/TizenRT/pull/2018#issuecomment-412433508

daeyeon:
We do encourage suggestions for IoT.js but do not guarantee IoT.js which each individual applies to external projects. And, its acceptance is beyond our jurisdiction.

#1732 can answer your query about the plan. A selected version of the platform is applied to our CI before a release. That's for post-maintenance and would be the same for next release.

PR #1732, opened 2018-08-29
tizenrt: Relocate downstream Makefile to config dir (along Kconfig)

rzr:
[Philippe Coval]

After some (recent) refactoring in build script,
This file was over imported iotjs module in TizenRT.

Like done previously with Kconfig file,
it will be better to minimize downstream patches.

One benefit to have those build related files,
is that then IoT.js can be upgraded into TizenRT
by just cloning iotjs's again into subdir:
TizenRT/external/iotjs

Note: Extra minor changes has been done over TizenRT's patch
to make some options overridable from env var (profile file...)

[Sunghan Chang]

Makefile: move IoT.js-specific build step to IoT.js folder

Because Makefile.unix has TizenRT-common build step,
it is not good including IoT.js-specific step.
Let's add Makefile in IoT.js and execute all of step in it.

Thanks-to: sunghan-chang <sh924.chang@samsung.com>
Change-Id: Iddeb272dc6ad6c283ccad9f92bf02754f9ba3240
Bug-TizenRT: https://github.com/Samsung/TizenRT/pull/2111
Origin: https://github.com/Samsung/TizenRT/commit/bbd3cdbd498ff11eeca3f021496423cc0222497e
Relate-to: https://github.com/Samsung/iotjs/issues/1726
Forwarded: https://github.com/Samsung/iotjs

### DIFF
--- a/config/tizenrt/Makefile
+++ b/config/tizenrt/Makefile
@@ -1,0 +1,54 @@
+###########################################################################
+#
+# Copyright 2018 Samsung Electronics All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+# either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+#
+###########################################################################
+
+-include $(TOPDIR)/.config
+-include $(TOPDIR)/Make.defs
+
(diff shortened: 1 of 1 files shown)


daeyeon:
IoT.js supports the latest public release of TizenRT. And, our TravisCI verifies each PR with the version as follow. https://github.com/Samsung/iotjs/blob/44cda4095d838ab3c89fb7fc45ad7ad5a001ef09/tools/travis_script.py#L45

This commit is based on the branch assumed as a development branch. That means this commit will never be tested in our current TravisCI environment.

We need an official release tag when discussing upgrading IoT.js snapshot for each target.

rzr:
Ok I understand your valid point , but aren't we in a chicken and egg situation ?
Wasnt this the case with the Kconfig file that was merged ?

Meanwhile it can be applied downstream... 
I'll update refs and prepare 2018-09-01 import

Anyway this change should be harmless to your current test env.

daeyeon:
In the maintenance perspective, this isn't the case you told. 
This commit is harmless but meaningless for now.

Issue #1379, opened 2017-12-20
Assertion 'jerry_value_is_function(jbuffer)' failed in iotjs_bufferwrap_create_buffer


renatahodovan:
###### IoT.js version:

```
Checked revision: afe242a
Build command: tools/build.py --buildtype debug
```
###### OS:

```
Ubuntu 17.10
```
###### Test case:

```javascript
var http = require('http');
var net = require('net');
var server = net.createServer(function(socket) {
    socket.end('Bad http\r\n\r\n');
    server.close();
});
server.listen(3085, '127.0.0.1');
var request = http.request({
    host: '127.0.0.1',
    port: 3085,
    method: 'GET',
    path: '/'
});
request.on('error', function(err) {
    Buffer++;
});
request.end();
```
###### Backtrace:
```
iotjs/src/modules/iotjs_module_buffer.c:221: Assertion 'jerry_value_is_function(jbuffer)' failed.

[Backtrace]:

(gdb) bt
#0  iotjs_bufferwrap_create_buffer (len=40) at iotjs/src/modules/iotjs_module_buffer.c:221
#1  0x0000555555580a6d in OnRead (handle=0x555555880da0, nread=40, buf=0x7fffffffae50) at iotjs/src/modules/iotjs_module_tcp.c:502
#2  0x00005555555f6932 in uv__read (stream=0x555555880da0) at iotjs/deps/libtuv/src/unix/stream.c:1234
#3  0x00005555555f6bf5 in uv__stream_io (loop=0x55555583d640 <default_loop_struct>, w=0x555555880e10, events=1) at iotjs/deps/libtuv/src/unix/stream.c:1300
#4  0x00005555555fbcda in uv__io_poll (loop=0x55555583d640 <default_loop_struct>, timeout=0) at iotjs/deps/libtuv/src/unix/linux-core.c:389
#5  0x00005555555eed99 in uv_run (loop=0x55555583d640 <default_loop_struct>, mode=UV_RUN_ONCE) at iotjs/deps/libtuv/src/unix/core.c:305
#6  0x000055555556f4fb in iotjs_start (env=0x55555583d440 <current_env>) at iotjs/src/iotjs.c:147
#7  0x000055555556f875 in iotjs_entry (argc=2, argv=0x7fffffffe308) at iotjs/src/iotjs.c:218
#8  0x000055555556f05a in main (argc=2, argv=0x7fffffffe308) at iotjs/src/platform/linux/iotjs_linux.c:19
```

[...]

PR #1736, opened 2018-09-04
Improve Buffer module internals

galpeter:
Main points of the change:
* The iotjs_bufferwrap_create_buffer method required
  the global object's Buffer property correctness.
  However by doing a "Buffer++" the global property
  will be changed and the new buffer's prototype will
  be set incorrect. This change exposes the JS side Buffer
  for the C side via the 'native.Buffer' property and as
  the native objects are not as accessible as global properties
  modifying the global buffer will not affect the C side.

* The util.isBuffer used the global's Buffer value, however
  it can also be changed via the "Buffer++" code.
  The isBuffer implementation is moved into the Buffer module.
  However, if the util module should not call require for the
  buffer module because that would cause a cyclical reference as
  the buffer module already requires the util module.

* Removed buffer module's dependency on the util module and
  the util module includes the buffer module to expose the isBuffer
  method.

### DIFF
--- a/src/js/buffer.js
+++ b/src/js/buffer.js
@@ -13,8 +13,6 @@
  * limitations under the License.
  */
 
-var util = require('util');
-
 
 function checkInt(buffer, value, offset, ext, max, min) {
   if (value > max || value < min)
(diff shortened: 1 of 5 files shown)


galpeter:
This PR should fix the last part of the #1379 issue.

DanielBallaSZTE:
Nice, I like this idea a lot.

PR #1737, opened 2018-09-04
Update jerry submodule

Yuyupo:
IoT.js-DCO-1.0-Signed-off-by: Daniella Barsony bella@inf.u-szeged.hu


src/iotjs_string_ext.c
```diff
@@ -48,9 +48,9 @@ static const jerry_length_t magic_string_lengths[] = {
 //
 // declare strings table
 //
-static const jerry_char_ptr_t magic_string_items[] = {
+static const jerry_char_t *magic_string_items[] = {
 #define MAGICSTR_EX_DEF(NAME, STRING) \
-  (const jerry_char_ptr_t) jerry_magic_string_ex_##NAME,
+  (const jerry_char_t *)jerry_magic_string_ex_##NAME,
```
robertsipka: Space is missing after the closing bracket.

tools/travis_script.py
```diff
@@ -135,10 +135,15 @@ def build_iotjs(buildtype, args=[], env=[]):
         for buildtype in BUILDTYPES:
             if buildtype == 'release':
                 set_release_config_tizenrt()
+            # Fixme: EXTRA_LIBPATHS and EXTRA_LIB can be deleted
+            # when TizenRT uses jerry-ext
```
robertsipka: Use uppercase letters for fixme, and a dot is missing from the end of the sentence.

tools/travis_script.py
```diff
@@ -135,10 +135,15 @@ def build_iotjs(buildtype, args=[], env=[]):
...
             exec_docker(DOCKER_TIZENRT_OS_PATH, [
                         'make', 'IOTJS_ROOT_DIR=' + DOCKER_IOTJS_PATH,
                         'IOTJS_BUILD_OPTION='
-                        '--profile=test/profiles/tizenrt.profile'])
+                        '--profile=test/profiles/tizenrt.profile',
+                        'EXTRA_LIBPATHS=-L' + DOCKER_IOTJS_PATH +
+                        '/build/arm-tizenrt/' + buildtype + '/lib/',
+                        'EXTRA_LIBS=-ljerry-ext'])
```
robertsipka: Do you really need to provide the -L / -l option flags in the `EXTRA_LIBPATHS` and `EXTRA_LIBS` values?

robertsipka: No need to answer, I see you need to give it, but I was surprised at first.

src/iotjs_string_ext.c
LaszloLango: Actually IoT.js style must be applied here (not JerryScript style), so the space is not needed.

tools/travis_script.py
Yuyupo: Thanks, i missed that. Updated the PR accordingly.

LaszloLango:
@daeyeon usually the linker deals with the unused functions/features. @Yuyupo, @robertsipka,  is this patch cause significant binary gain in any platform? IMHO the main goal of the `jerry-ext` should be to contain all of the non-standard features, like `print builtin` or `debug server`.

zherczeg:
Yes, the design goal of jerry-ext is that the extensions are not depend on each other. So the linker can cut the unused parts. However this would be good to check in practice. I agree to check the binary size change.

robertsipka:
After the libjerry update, the binary sizes with target and test specific profile builds - that contain all the modules that the tests require - are the following:
 - on STM32F4-Discovery : 247235 bytes.
 - on ARTIK 053 : 249596 bytes.
 - on Raspberry Pi 2 : 290130 bytes.

With the latest commit on IoT.js (https://github.com/Samsung/iotjs/commit/40ee244395ce4bf239f62e27ff68a95d251882a4) are the following:
 - on STM32F4-Discovery : 247731 bytes.
 - on ARTIK 053 : 252125 bytes.
 - on Raspberry Pi 2 : 292631 bytes.

Conclusion: The binary sizes will be decreased after this patch on these devices.

zherczeg:
Nice! Thank you for the measurement.

daeyeon:
The size will be more decreased if we can take only debugger from the extensions. 

I'm fine that jerry-ext contains all the non-standard features. However, the more non-standard features merged in future will increase the size of jerry-ext, and that will affect the size of IoT.js.

I think that jerry should provide a way to selectively use the features we only want. (Currently only debugger)

LaszloLango:
@Yuyupo https://github.com/jerryscript-project/jerryscript/pull/2507 is merged. Could you update the jerry hash?

Yuyupo:
@LaszloLango Updated jerry to the latest.

LaszloLango:
@daeyeon We can open an issue in JerryScript to make the extensions configurable. It would be a good improvement in the engine. It might reduce the compile time too.

daeyeon:
@LaszloLango Thank you for the opinion.

@Yuyupo I can't guess how much this update is urgent. Could you let me know whether waiting for the configuration is fine to you?

PR #1739, opened 2018-09-04
Extract common code in some JS_FUNCTIONs into common functions

pmarkee:

### DIFF
--- a/src/modules/iotjs_module_fs.c
+++ b/src/modules/iotjs_module_fs.c
@@ -202,7 +202,13 @@ JS_FUNCTION(Open) {
 }
 
 
-JS_FUNCTION(Read) {
+typedef enum { IOTJS_FS_READ, IOTJS_FS_WRITE } iotjs_fs_op_t;
+
+jerry_value_t fs_do_read_or_write(const jerry_value_t jfunc,
+                                  const jerry_value_t jthis,
+                                  const jerry_value_t jargv[],
+                                  const jerry_length_t jargc,
+                                  const iotjs_fs_op_t fs_op) {
   DJS_CHECK_THIS();
   DJS_CHECK_ARGS(5, number, object, number, number, number);
   DJS_CHECK_ARG_IF_EXIST(5, function);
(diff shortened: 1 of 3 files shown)


daeyeon:
Please check the travis build failures.

src/modules/iotjs_module_uart.c
```diff
@@ -188,37 +188,45 @@ JS_FUNCTION(UartCons) {
...
   }
 
   return jerry_create_undefined();
 }
 
+JS_FUNCTION(Write) {
+  return uart_do_write_or_writesync(jfunc, jthis, jargv, jargc, true);
+}
+
+JS_FUNCTION(WriteSync) {
```
zherczeg: Two newlines between functions.

zherczeg: Hm, it seems this file has a different style. Ok, keep it this way.

src/modules/iotjs_module_fs.c
```diff
@@ -202,7 +202,10 @@ JS_FUNCTION(Open) {
 }
 
 
-JS_FUNCTION(Read) {
+jerry_value_t fs_do_read_or_write(const jerry_value_t jfunc,
+                                  const jerry_value_t jthis,
+                                  const jerry_value_t jargv[],
+                                  const jerry_length_t jargc, bool do_read) {
```
rtakacs: `do_read` should be in a new line, below the `jargc` parameter.

rtakacs: In general, it's better to use custom types for `boolean` typed parameters, because that values are more informative. For example:

```c
typedef enum {
  IOTJS_FS_READ,
  IOTJS_FS_WRITE
} iotjs_fs_t;

jerry_value_t fs_do_read_or_write(const jerry_value_t jfunc,
                                  const jerry_value_t jthis,
                                  const jerry_value_t jargv[],
                                  const jerry_length_t jargc,
                                  const iotjs_fs_t fs_op);

fs_do_read_or_write(jfunc, jthis, jargv, jargc, IOTJS_FS_WRITE);
```

pmarkee: clang-format wants `do_read` and `jargc` to be on the same line for some reason, style check was failing earlier because of it on Travis.

PR #1742, opened 2018-09-06
Improve JerryScript native object access

galpeter:
Changes:
* Modified iotjs_jval_get_object_native_handle to accept an additional
  parameter specifying the native pointer's expected info.
  If no info is specified the native pointer is returned in all cases.
* __JS_DECLARE_PTR renamed to JS_DECLARE_PTR.
* From the JS_DECLARE_PTR the direct native pointer access code parts
  are removed and it now uses the modified iotjs_jval_get_object_native_handle.
* Updated modules to use the modified iotjs_jval_get_object_native_handle
  method.

The change reduces the binary code size a bit.

### DIFF
--- a/src/iotjs_binding.c
+++ b/src/iotjs_binding.c
@@ -341,12 +341,17 @@ jerry_value_t iotjs_jval_get_property(jerry_value_t jobj, const char* name) {
 }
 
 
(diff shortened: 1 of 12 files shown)


LaszloLango:
@galpeter Please rebase the PR.

galpeter:
@LaszloLango PR is rebased, everything should be ok. But we'll see after the checks are done.

PR #1744, opened 2018-09-07
Merge some JS_FUNCTIONs in iotjs_module_udp.c

pmarkee:


pmarkee:
The `SetBroadcast`, `SetTTL`, `SetMulticastTTL` and `SetMulticastLoopback` functions in the udp c module had a lot of code in common. This patch creates a new `Configure` function, which does the same things as these 4 functions. The `Configure` function takes an additional numeric argument that tells what operation it has to do.

src/modules/iotjs_module_udp.c
```diff
@@ -283,61 +283,44 @@ JS_FUNCTION(GetSockeName) {
...
+      break;
+    }
+    case MULTICASTTTL: {
+      fn = &uv_udp_set_multicast_ttl;
+      break;
+    }
+    case MULTICASTLOOPBACK: {
+      fn = &uv_udp_set_multicast_loop;
+      break;
+    }
```
galpeter: A default case should be added, which contains an assert and returns nothing.

src/modules/iotjs_module_udp.c
```diff
@@ -283,61 +283,44 @@ JS_FUNCTION(GetSockeName) {
...
-#if !defined(__NUTTX__)
-  IOTJS_UV_SET_SOCKOPT(uv_udp_set_ttl);
-#else
-  IOTJS_ASSERT(!"Not implemented");
-
-  return jerry_create_null();
-#endif
-}
-
+enum config_type { BROADCAST, TTL, MULTICASTTTL, MULTICASTLOOPBACK };
```
galpeter: We should add a comment that the order should match the dgram.js order. A comment should be here and in the dgram.js so if someone updates these then it is clear where should the developer look further.

pmarkee:
@galpeter I have done the requested changes.

src/js/dgram.js
```diff
@@ -350,8 +350,16 @@ Socket.prototype.address = function() {
 };
 
 
+// These variables represent the different config types that
+// this._handle.configure can do.
+// The order of these must match the order in the udp C module.
+var BROADCAST = 0;
```
daeyeon: Can we declare these using `enum` just like the same in C module?

PR #1747, opened 2018-09-11
Add restart functionality

Yuyupo:
It was needed for the VSCode extension.

IoT.js-DCO-1.0-Signed-off-by: Daniella Barsony bella@inf.u-szeged.hu

### DIFF
--- a/src/iotjs.c
+++ b/src/iotjs.c
@@ -122,6 +122,25 @@ bool iotjs_initialize(iotjs_environment_t* env) {
   return true;
 }
 
+void iotjs_restart(iotjs_environment_t* env, jerry_value_t jmain) {
+  jerry_value_t abort_value = jerry_get_value_from_error(jmain, false);
+  if (jerry_value_is_string(abort_value)) {
+    /* TODO: When there is an api function to check for reset,
+    this needs an update. */
+    static const char restart_str[] = "r353t";
+
+    jerry_size_t str_size = jerry_get_string_size(abort_value);
+
+    if (str_size == sizeof(restart_str) - 1) {
+      jerry_char_t str_buf[5];
+      jerry_string_to_char_buffer(abort_value, str_buf, str_size);
+      if (memcmp(restart_str, (char*)(str_buf), str_size) == 0) {
+        iotjs_environment_config(env)->debugger->context_reset = true;
+      }
+    }
+  }
+  jerry_release_value(abort_value);
+}
 
 void iotjs_run(iotjs_environment_t* env) {
 // Evaluating 'iotjs.js' returns a function.
@@ -134,7 +153,10 @@ void iotjs_run(iotjs_environment_t* env) {
                           iotjs_js_modules_l, module_iotjs_idx, 0);
 #endif
 
-  if (jerry_value_is_error(jmain) && !iotjs_environment_is_exiting(env)) {
+  if (jerry_value_is_abort(jmain)) {
+    iotjs_restart(env, jmain);
+  } else if (jerry_value_is_error(jmain) &&
+              !iotjs_environment_is_exiting(env)) {
     jerry_value_t errval = jerry_get_value_from_error(jmain, false);
     iotjs_uncaught_exception(errval);
     jerry_release_value(errval);


src/iotjs.c
```diff
@@ -134,7 +134,27 @@ void iotjs_run(iotjs_environment_t* env) {
                           iotjs_js_modules_l, module_iotjs_idx, 0);
 #endif
 
-  if (jerry_value_is_error(jmain) && !iotjs_environment_is_exiting(env)) {
+  if (jerry_value_is_abort(jmain)) {
+    jerry_value_t abort_value = jerry_get_value_from_error(jmain, false);
+    if (jerry_value_is_string(abort_value)) {
+      static const char restart_str[] = "r353t";
+
+      jerry_value_t str_val = jerry_value_to_string(abort_value);
```
galpeter: AFAIK: As the value is already a string, there is no need to call the `toString` conversion.

src/iotjs.c
```diff
@@ -134,7 +134,27 @@ void iotjs_run(iotjs_environment_t* env) {
...
+      jerry_size_t str_size = jerry_get_string_size(str_val);
+
+      if (str_size == sizeof(restart_str) - 1) {
+        JERRY_VLA(jerry_char_t, str_buf, str_size);
```
galpeter: We already know the size of the reset string, no need to VLA:
```
jerry_char_t* str_buf[5];
```

src/iotjs.c
```diff
@@ -134,7 +134,27 @@ void iotjs_run(iotjs_environment_t* env) {
                           iotjs_js_modules_l, module_iotjs_idx, 0);
 #endif
 
-  if (jerry_value_is_error(jmain) && !iotjs_environment_is_exiting(env)) {
+  if (jerry_value_is_abort(jmain)) {
+    jerry_value_t abort_value = jerry_get_value_from_error(jmain, false);
```
galpeter: IMHO: This whole code block inside the `if` should be moved into its own method.

src/iotjs.c
```diff
@@ -134,7 +134,27 @@ void iotjs_run(iotjs_environment_t* env) {
...
+        jerry_string_to_char_buffer(str_val, str_buf, str_size);
+        if (memcmp(restart_str, (char*)(str_buf), str_size) == 0) {
+          iotjs_environment_config(env)->debugger->context_reset = true;
+        }
+      }
+      jerry_release_value(str_val);
+    }
+    jerry_release_value(abort_value);
+  }
+
```
galpeter: No need for the extra empty line

Yuyupo:
Thanks for the review @galpeter. Updated the PR according to what you requested.

yorkie:
@Yuyupo Just curious how do u integrate on Visual Studio Code :)

Yuyupo:
@yorkie We already have a VSCode extension
https://marketplace.visualstudio.com/items?itemName=Samsung.iotjs-vscode-extension
The restart functionality is already in there but the message it was sending to IoTjs wasn't supported yet.

src/iotjs.c
```diff
@@ -122,6 +122,23 @@ bool iotjs_initialize(iotjs_environment_t* env) {
   return true;
 }
 
+void iotjs_restart(iotjs_environment_t* env, jerry_value_t jmain) {
+  jerry_value_t abort_value = jerry_get_value_from_error(jmain, false);
+  if (jerry_value_is_string(abort_value)) {
+    static const char restart_str[] = "r353t";
```
daeyeon: Is this a jerry debugger feature, or else a feature only for vscode iotjs extension? It seems better to directly refer to the value defined in any of them if possible.

zherczeg: We could add an api function to check this. Sounds like a good idea.

src/iotjs.c
```diff
@@ -134,7 +151,11 @@ void iotjs_run(iotjs_environment_t* env) {
                           iotjs_js_modules_l, module_iotjs_idx, 0);
 #endif
 
-  if (jerry_value_is_error(jmain) && !iotjs_environment_is_exiting(env)) {
+  if (jerry_value_is_abort(jmain)) {
+    iotjs_restart(env, jmain);
+  }
+
```
galpeter: There should be no newline here, and the format should be:
```
...
} else if {
...
```

src/iotjs.c
galpeter: As adding an extra jerry api function would require also a submodule update here. I would put a TODO mark about it here and merge this PR as soon as possible. In the mean time the api function can be added.

PR #1749, opened 2018-09-17
Disable jerryscript debugger code if not enabled

rzr:
Observed build issue on TizenRT:

    iotjs.c:58: undefined reference to `jerryx_debugger_tcp_create'
    iotjs.c:59: undefined reference to `jerryx_debugger_ws_create'
    iotjs.c:58: undefined reference to `jerryx_debugger_after_connect'


daeyeon:
- Travis fails.

- It looks like an error occurred in compile time. I'm curious on how it's not been failed on Travis before this patch? Can anyone answer that?

hs0225:
@rzr Can you give the command that failed the build?

rtakacs:
TizenRT has an embedded (old version) IoT.js. TizenRT has Makefiles (`TizenRT/os/FlatLibs.mk`, `TizenRT/os/LibTargets.mk`, `TizenRT/os/ProtectedLibs.mk`) that define which libraries should be used when linking Tinyara. These Makefiles do not include the `jerry-ext` library, because the embedded IoT.js is an old one, and that version of IoT.js does not use `jerry-ext`.

In the https://github.com/Samsung/iotjs/pull/1737 (last jerry update) patch, there is a workaround to eliminate the linking errors by using the `EXTRA_LIBPATHS` and `EXTRA_LIBS` environment variables (used by TizenRT) in the [travis_script.py:L140](https://github.com/Samsung/iotjs/blob/master/tools/travis_script.py#L140) file.

rzr:
Yes the build error occurs on unsupported TizenRT master branch, I can try to fix it using the EXTRA_LIBS hint shared by @rtakacs.

ps: This patch may not be merged but as all downstream patches applying to iotjs, I thought it was good to show them upstream.

src/iotjs.c
```diff
@@ -55,11 +55,13 @@ static bool jerry_initialize(iotjs_environment_t* env) {
 
   if (iotjs_environment_config(env)->debugger != NULL) {
     uint16_t port = iotjs_environment_config(env)->debugger->port;
+#ifdef JERRY_DEBUGGER
```
galpeter: If the debugger is disabled why not disable the whole debugger part here? (from line 56 to 65).
Currently some of the builds on Travis are failing due to the unused port variable. (note: as it is used on line 64, but that is only a debug logging which is not present in release builds, thus the port variable will be unused)

PR #1756, opened 2018-09-27
IoT.js Module Generator

fbmrk:
The purpose of this tool to create a native iotjs module from a C / C ++ API by generating the binding layer. It's still in progress because the C ++ part has not been completed yet.

For more information and specification please read `docs/devs/IoT.js-Module-Generator.md`

IoT.js-DCO-1.0-Signed-off-by: Marko Fabo mfabo@inf.u-szeged.hu

### DIFF
--- a/tools/iotjs-generate-module.py
+++ b/tools/iotjs-generate-module.py
@@ -1,0 +1,221 @@
+#!/usr/bin/env python
+
+# Copyright 2019-present Samsung Electronics Co., Ltd. and other contributors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import os
+import sys
+
+from common_py import path
+from common_py.system.filesystem import FileSystem as fs
+
+from module_generator.source_generator import CSourceGenerator, \
+    CppSourceGenerator
+from module_generator.clang_translation_unit_visitor import ClangTUVisitor
+
+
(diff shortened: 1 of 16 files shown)


yichoi:
Looks promising! It would be better to provide a novel way to avoid memory leaks at this generator.

fbmrk:
What do you mean exactly? Are there problems with jerry values or the direct `malloc` calls in the binding layer?

yichoi:
It's general problem that Native layer and JS layer cannot trace each others memory life time exactly therefore memory could be leaked. I am just throwing my idea, please don't be distracted. :)

daeyeon:
Interesting idea. Comparing the other binding generating idea like a conventional IDL-based one, users might feel familiar because it's C/C++ language-based.

We'd like to know what the header for binding looks like for a _stateful_ module such as `Buffer`. Can you add the example also?

fbmrk:
I have updated the PR and made some refactoring. I have also add an example binding layer to `docs/devs/IoT.js-Module-Generator.md`.

The C++ part is still in progress.

fbmrk:
I have updated the PR. The tool has reached a usable state and it gives some help for developers to avoid writing a lots of code. It doesn't solve the whole problem of create a binding between C/C++ and JavaScript languages, but in simple cases it can fully do the task of the developer and works well. The tool can't handle perfect when there are complex user-defined types, pointers to structures, multiple indirections etc. in the native library.
For more information please check the documentation: `docs/devs/IoT.js-Module-Generator.md`

tools/iotjs-generate-module.py
```diff
@@ -0,0 +1,221 @@
+#!/usr/bin/env python
+
+# Copyright 2018-present Samsung Electronics Co., Ltd. and other contributors
```
rtakacs: The year should be updated.

fbmrk: Done, thank you

daeyeon:
1. Can you move the example (`/test_c`, `/text_cpp`, and `test.py`) into `test/{new folder}`?
2. In `docs/devs/IoT.js-Module-Generator.md`, it seems better to mention the current limitation you mentioned above.

fbmrk:
@daeyeon Thank you for your comments. I've updated the PR.

PR #1759, opened 2018-10-04
Pass the extra JerryScript build parameters for the snapshot tool build

galpeter:
When using `--jerry-cmake=..` argument for the build script
the arguments were not passed for the snapshot tool.
This could result in an generated incompatible snapshot for the
iotjs binary.

To fix this all `jerry-cmake` params should be passed for the
snapshot tool also.


rtakacs:
LGTM (informal)

LaszloLango:
@galpeter  Please fix the ASAN/UBSAN build failures.

PR #1766, opened 2018-10-09
Add options argument for http createServer

galpeter:
Improve compatibility with node.js by adding the `options`
argument for the createServer method in the http modules.

In addition allow "subclassing" the http.Incoming and httpServerResponse
objects. With this it is possible to augment the created request and
response objects passed for the `requestListener` method.

### DIFF
--- a/src/js/http_server.js
+++ b/src/js/http_server.js
@@ -14,6 +14,7 @@
  */
 
 var util = require('util');
+var IncomingMessage = require('http_incoming').IncomingMessage;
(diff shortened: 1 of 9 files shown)


src/js/http_server.js
```diff
@@ -145,10 +146,20 @@ ServerResponse.prototype.detachSocket = function() {
 
 
 function initServer(options, requestListener) {
+  if (util.isFunction(options)) {
+    requestListener = options;
+  }
+
+  if (options == null) {
```
zherczeg: `typeof options != "object"` ?

PR #1767, opened 2018-10-10
Implement stream.Readable.pipe() for iotjs

pmarkee:

### DIFF
--- a/src/js/stream_readable.js
+++ b/src/js/stream_readable.js
@@ -15,6 +15,7 @@
 
 
 var Stream = require('stream_internal');
+var Writable = require('stream_writable');
 var util = require('util');
 
 
@@ -138,6 +139,76 @@ Readable.prototype.push = function(chunk, encoding) {
 };
 
 
+Readable.prototype.pipe = function(destination, options) {
+  if (!(destination instanceof Writable || isDuplex(destination))) {
+    throw new TypeError('pipe excepts stream.Writable or' +
+                        ' stream.Duplex as argument');
+  }
+
+  options = options || {'end': true};
+
+  var listeners = {
+    readableListener: readableListener.bind(this),
+    dataListener: dataListener.bind(destination),
+    endListener: endListener.bind(destination),
+  };
+
+  this.on('readable', listeners.readableListener);
+  this.on('data', listeners.dataListener);
+
+  if (options.end) {
+    this.on('end', listeners.endListener);
+  }
+
+  this._piped = this._piped || [];
+  this._piped.push(destination);
+
+  this._piped_listeners = this._piped_listeners || [];
+  this._piped_listeners.push(listeners);
+
+  return destination;
+};
+
+
+Readable.prototype.unpipe = function(destination) {
+  if (destination === undefined) {
+    this.removeAllListeners();
+    this._piped = undefined;
+    this._piped_listeners = undefined;
+    return;
+  }
+
+  var idx = this._piped.indexOf(destination);
+  if (idx === -1) {
+    return;
+  }
+
+  this._piped.splice(idx, 1);
+  var listeners = this._piped_listeners.splice(idx, 1)[0];
+
+  this.removeListener('readable', listeners.readableListener);
+  this.removeListener('data', listeners.dataListener);
+  this.removeListener('end', listeners.endListener);
+
+  return destination;
+};
+
+
+function readableListener() {
+  this.resume();
+}
+
+
+function dataListener(data) {
+  this.write(data);
+}
+
+
+function endListener() {
+  this.end();
+}
+
+
 function readBuffer(stream, n) {
   var state = stream._readableState;
   var res;
@@ -195,6 +266,23 @@ function onEof(stream) {
   if (state.length == 0) {
     emitEnd(stream);
   }
+}
+
+
+function isDuplex(stream) {
+  if (!(stream instanceof Readable)) {
+    return false;
+  }
+
+  var wr_keys = Object.keys(Writable.prototype);
+  for (var i = 0; i < wr_keys.length; i++) {
+      var wr_key = wr_keys[i];
+      if (!stream[wr_key]) {
+          return false;
+      }
+  }
+
+  return true;
 }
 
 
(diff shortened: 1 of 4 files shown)


src/js/stream_readable.js
```diff
@@ -138,6 +139,33 @@ Readable.prototype.push = function(chunk, encoding) {
...
+  options = options || {'end': true};
+
+  var state = this._readableState;
+
+  this.on('readable', function() {
+    this.resume();
+  });
+
+  this.on('data', function(data) {
+    destination.write(data, function() {});
```
zherczeg: do we need this `function() {}` ?

pmarkee: We don't, I left it there by accident.

src/js/stream_readable.js
```diff
@@ -138,6 +139,76 @@ Readable.prototype.push = function(chunk, encoding) {
 };
 
 
+Readable.prototype.pipe = function(destination, options) {
+  if (!(destination instanceof Writable || isDuplex(destination))) {
```
rtakacs: I think it's enough to check if `destination` is readable:

```javascript
if (destination instanceof Readable) {
    throw new TypeError('pipe excepts writable stream argument');
}
```
In this case the `isDuplex` function can be removed.

test/run_pass/test_stream_pipe.js
```diff
@@ -0,0 +1,109 @@
...
+var Duplex = stream.Duplex;
+var Readable = stream.Readable;
+var Writable = stream.Writable;
+
+var readable = new Readable();
+
+var msg = [];
+var resp = [];
+
+// 1
```
rtakacs: It's better to use a more informative comment. E.g. `// Step 1` or `// Test 1` or `// Testcase 1`

src/js/stream_readable.js
```diff
@@ -138,6 +139,76 @@ Readable.prototype.push = function(chunk, encoding) {
...
+Readable.prototype.unpipe = function(destination) {
+  if (destination === undefined) {
+    this.removeAllListeners();
+    this._piped = [];
+    this._piped_listeners = [];
+    return;
+  }
+
+  var idx = this._piped.indexOf(destination);
+  if (idx == -1) {
```
rtakacs: Please use strict equality operator.

test/run_pass/test_stream_pipe.js
```diff
@@ -0,0 +1,109 @@
...
+}
+
+dest6._readyToWrite();
+
+readable2.push(msg[2]);
+readable2.pipe(dest6);
+readable2.unpipe(dest6);
+
+process.on('exit', function() {
+  assert.equal(msg[0], resp[0], true);
```
galpeter: Why is the third argument `true`? The third one should be an assert message if used.

test/run_pass/test_stream_pipe.js
```diff
@@ -0,0 +1,109 @@
...
+
+msg.push('Hello');
+
+dest1._write = function(chunk, callback, onwrite) {
+  resp.push(chunk.toString());
+}
+
+dest1._readyToWrite();
+
+readable.pipe(dest1);
```
galpeter: We should also add test case when the source and destination is a file and when it is a network socket.

test/run_pass/test_stream_pipe.js
galpeter: Based on these assert's I'm not sure what are we testing? Can we have some comments, assert messages?

src/js/stream_readable.js
```diff
@@ -138,6 +139,76 @@ Readable.prototype.push = function(chunk, encoding) {
...
+  options = options || {'end': true};
+
+  var listeners = {
+    'readableListener': readableListener.bind(this),
+    'dataListener': dataListener.bind(destination),
+    'endListener': endListener.bind(destination),
+  };
+
+  this.on('readable', listeners.readableListener);
+  this.on('data', listeners.dataListener);
```
galpeter: What will happen if there is a stream error? (source/destination)

src/js/stream_readable.js
rtakacs: Yes, You are right.

test/run_pass/test_stream_pipe.js
pmarkee: My bad, I knew wrong what the third parameter has to be. I will write some more informative assert messages.

src/js/stream_readable.js
```diff
@@ -138,6 +139,76 @@ Readable.prototype.push = function(chunk, encoding) {
 };
 
 
+Readable.prototype.pipe = function(destination, options) {
```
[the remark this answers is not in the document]
pmarkee: According to the node.js documentation, it has to be possible to attach multiple destination streams to the same Readable, and it works here too. I will add a test case for this to `test_stream_pipe.js`.

src/js/stream_readable.js
galpeter: If the node.js does not state anything else, then it's ok. I'm a bit against adding extra options.

src/js/stream_readable.js
```diff
@@ -138,6 +139,76 @@ Readable.prototype.push = function(chunk, encoding) {
...
+
+  return destination;
+};
+
+
+Readable.prototype.unpipe = function(destination) {
+  if (destination === undefined) {
+    this.removeAllListeners();
+    this._piped = [];
+    this._piped_listeners = [];
```
zherczeg: These two could be set to `undefined` since the array is created if needed.

src/js/stream_readable.js
```diff
@@ -138,6 +139,76 @@ Readable.prototype.push = function(chunk, encoding) {
...
+    throw new TypeError('pipe excepts stream.Writable or' +
+                        ' stream.Duplex as argument');
+  }
+
+  options = options || {'end': true};
+
+  var listeners = {
+    'readableListener': readableListener.bind(this),
+    'dataListener': dataListener.bind(destination),
+    'endListener': endListener.bind(destination),
```
zherczeg: No need quote around the property names.

pmarkee:
Thanks @zherczeg , I have updated the PR.

PR #1768, opened 2018-10-10
Fixed 'writeSync' method of GPIO after #1739.

LaszloLango:
IoT.js-DCO-1.0-Signed-off-by: László Langó llango.u-szeged@partner.samsung.com

PR #1775, opened 2018-10-16
Introduce mock linux platform for GPIO testing on Travis CI

LaszloLango:
Related issue: #1779


LaszloLango:
This is a proof of concept patch. IMHO it can help to catch bugs on the CI before patches merging. I know it is not equal to testing on real hardware. Although it revealed a few issues in the GPIO module. You can see the commented test cases in the newly added `test/run_pass/test_gpio_api.js`. If we accept this patch then we can do the same for `i2c`, `adc`, `pwm` and  `spi` (maybe for `uart` too). I think this patch can help to maintain the peripheral modules and makes the them more stable. Bugs like #1768 (caused by #1739) could be caught on the CI with this patch.

@daeyeon @hs0225 what do you think about this approach (peripheral mocking)?

hs0225:
@LaszloLango I agree with this PR. Although this does not test on the target, it is expected to catch some issues about API.

PR #1776, opened 2018-10-16
Move test_tls.js skip into iotjs

Achie72:
The test was removed from js-remote-test, because it fails on all artik053 boards, and before it was locally skipped in the testsystem.

### DIFF
--- a/test/testsets.json
+++ b/test/testsets.json
@@ -780,6 +780,10 @@
     },
     {
       "name": "test_tls.js",
+      "skip": [
+        "tizenrt"
+      ],
+      "reason": "Error: SSL setup failed (no enough memory for the test)",
       "required-modules": [
         "tls",
         "fs"


test/testsets.json
```diff
@@ -840,6 +840,13 @@
         "websocket",
         "tls"
       ]
+    },
+    {
+      "name": "test_tls.js",
+      "skip": [
+        "tizenrt"
+      ],
+      "reason": "Error: SSL setup failed (no enough memory for the test)",
```
rtakacs: The `test_tls.js` is already exist in the `testsets.json` file. It's enough to modify that entry with `skip` and `reason` information.

rtakacs: Note: the comma should be removed at the end of the `reason` line.

rtakacs:
LGTM (informal)

hs0225:
@rtakacs Why is not enough memory? Is the socket too open to run out of memory? If so, it would be better to modify the config of TizenRT or split the test file.

LaszloLango:
@rtakacs @Achie72  Could you try to split the test file into smaller ones?

rtakacs:
@hs0225 @LaszloLango The `test_tls.js` fails from the beginning (when TizenRT's mbedtls is used). As I debugged, either allocator function of TizenRT (malloc or calloc) could not allocate memory. (NULL pointer was returned for a valid memory request.)
Of course, we can investigate the issue in more detail, and if it's necessary, the test will be separated into smaller parts.

hs0225:
@rtakacs I want to solve this problem in a way this test pass, not skip.

LaszloLango:
@rtakacs Can we set a bigger jerry heap size in js-remote-test? That might be the easiest way to solve this.

rtakacs:
@LaszloLango The increased Jerry heap size doesn't solve the problem. The out-of-memory happens on system heap, and not on Jerry heap area.

The `malloc` function of TizenRT produces the following message before: `Allocation failed, size 16736`. I don't know too much about the allocator of TIzenRT, but a fragmented memory can cause similar cases.

@hs0225 I've checked the kernel config file of TizenRT but I didn't find any options that could solve this out-of-memory. I'm starting to split the `test_tls.js` file into smaller parts. I'm closing this pull request and I will create a new one for the new `test_tls_1.js` `test_tls_2.js` files.

PR #1784, opened 2018-10-30
Add node-like file streams to IoT.js

pmarkee:

### DIFF
--- a/docs/api/IoT.js-API-File-System.md
+++ b/docs/api/IoT.js-API-File-System.md
@@ -6,6 +6,8 @@ The following shows fs module APIs available for each platform.
 | :---: | :---: | :---: | :---: | :---: | :---: |
 | fs.close | O | O | O | O | O |
 | fs.closeSync | O | O | O | O | O |
+| fs.createReadStream | O | O | O | O | O |
+| fs.createWriteStream | O | O | O | O | O |
 | fs.exists | O | O | O | O | O |
 | fs.existsSync | O | O | O | O | O |
 | fs.fstat | O | O | O | X | X |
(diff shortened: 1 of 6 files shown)


src/js/fs.js
```diff
@@ -382,6 +385,16 @@ fs.readdirSync = function(path) {
 };
 
 
+fs.createReadStream = function(path) {
+  return new ReadStream(path);
```
zherczeg: Please add `options` to this function and pass it to `ReadStream`.

src/js/fs.js
```diff
@@ -382,6 +385,16 @@ fs.readdirSync = function(path) {
...
+}
+
+
+fs.createWriteStream = function(path) {
+  return new WriteStream(path);
```
zherczeg: Ditto.

src/js/fs.js
```diff
@@ -471,3 +484,101 @@ function checkArgString(value, name) {
...
+    return new ReadStream(path, options);
+  }
+
+  var defaultOptions = {
+    flags: 'r',
+    encoding: null,
+    fd: null,
+    mode: 0666,
+    autoClose: true,
+    bufferSize: 65536,
```
zherczeg: 64K buffer on embedded??? Max 4K but maybe 1K is enough.

src/js/fs.js
```diff
@@ -471,3 +484,101 @@ function checkArgString(value, name) {
...
+  fs.open(path, options.flags, options.mode, function(err, _fd) {
+    self._fd = _fd;
+
+    fs.read(self._fd, self._buffer, 0, self._buffer.length, -1, function(err, bytes_read, buffer) {
+      if (err) {
+        throw new Error("Error reading file");
+      }
+
+      if (bytes_read !== 0) {
+        self.push(self._buffer.slice(0, bytes_read));
```
zherczeg: Why slice is needed?

src/js/fs.js
```diff
@@ -471,3 +484,101 @@ function checkArgString(value, name) {
...
+      }
```
zherczeg: Isn't this call should recursively read the file?

src/js/fs.js
```diff
@@ -471,3 +484,101 @@ function checkArgString(value, name) {
...
+      closeFile(this._fd);
+    });
+  }
+}
+
+
+util.inherits(WriteStream, Writable);
+
+
+WriteStream.prototype._write = function(chunk, callback, onwrite) {
```
zherczeg: What happens with `callback` and `onwrite`? I think both needs to be called.

src/js/fs.js
pmarkee: It is needed because `self._buffer` may be bigger than the number of bytes we read, so there will be some zeros at the end of the buffer, which should not be `push`ed. This caused a problem when piping a readable filestream into a writable filestream, because the whole buffer will be written to the output file, so the file will contain all those zeros at the end. So for example:
```js
var fs = require('fs');

// let's say in.txt contains "hello world"
var rStream = fs.createReadStream('in.txt');
var wStream = fs.createWriteStream('out.txt');
wStream._readyToWrite();

rStream.pipe(wStream);
// out.txt will contain "hello world" and a lot of zeros at the end, depending on the buffer size of rStream
```

pmarkee:
Thanks for the review @zherczeg , I have updated the pull request. Some of the documentation is still missing, I will add that soon as well.

pmarkee:
@zherczeg I have updated the patch with the rest of the documentation and some fixes.

src/js/fs.js
```diff
@@ -382,6 +382,182 @@ fs.readdirSync = function(path) {
...
+    var defaultOptions = {
+      flags: 'r',
+      encoding: null,
+      fd: null,
+      mode: 438, // == octal 0666
+      autoClose: true,
+      bufferSize: 4096,
+    };
+
+    options = options || defaultOptions;
```
galpeter: What will happen if someone only specify one option? For example:
`fs.createReadStream("ble.txt", {flags: 'r'})`

src/js/fs.js
```diff
@@ -382,6 +382,182 @@ fs.readdirSync = function(path) {
...
+
+    this.bytesRead = 0;
+    this.path = path;
+    this._autoClose = options.autoClose;
+    this._bufferSize = options.bufferSize;
+    this._fd = options.fd;
+    this._buff = new Buffer(options.bufferSize);
+
+    var self = this;
+    if (!this._fd) {
```
galpeter: AFAIK this would mean that we can not open the stdin (that has an fd number of zero). It would be better to check the type of the fd.

pmarkee:
Thanks for the review @galpeter , I have updated the pull request.

src/js/fs.js
```diff
@@ -382,6 +382,196 @@ fs.readdirSync = function(path) {
...
+
+
+  fs.createWriteStream = function(path, options) {
+    return new WriteStream(path, options);
+  };
+} catch(e) {
+}
+
+
+function closeFile(stream) {
```
zherczeg: Why this is outside of try-catch?

docs/api/IoT.js-API-File-System.md
```diff
@@ -111,6 +111,162 @@ var fd = fs.openSync('test.txt', 'r');
...
+If it is `true` (as by default), the file will be closed automatically
+when end of file is reached or the stream ends.
+
+**Example**
+
+```js
+var fs = require('fs');
+
+var wStream = fs.createWriteStream('example.txt');
+wStream._readyToWrite();
```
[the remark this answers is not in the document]
pmarkee: We don't need to call this in node, but AFAIK in iotjs it is needed when using writable streams. How about putting this call in the constructor of `fs.WriteStream`?

src/js/fs.js
```diff
@@ -382,6 +382,196 @@ fs.readdirSync = function(path) {
...
+      });
+    }
+  }
+
+
+  util.inherits(WriteStream, Writable);
+
+
+  WriteStream.prototype._write = function(chunk, callback, onwrite) {
+    var buff = new Buffer(chunk);
```
galpeter: Isn't the chunk already a buffer?

src/js/fs.js
```diff
@@ -382,6 +382,196 @@ fs.readdirSync = function(path) {
...
+             function(err, bytes_written/* , buffer */) {
+      if (err) {
+        if (self._autoClose) {
+          closeFile(self);
+        }
+        throw err;
+      }
+      this.bytesWritten += bytes_written;
+    });
+    if (callback) {
```
galpeter: So we have two callbacks? `callback` and `onwrite`? Should these be called inside the `fs.write`?

pmarkee:
Thanks @galpeter @zherczeg , I have updated the PR.

src/js/fs.js
```diff
@@ -382,6 +382,196 @@ fs.readdirSync = function(path) {
...
+    }
+
+    var defaultOptions = {
+      flags: 'r',
+      encoding: null,
+      fd: null,
+      mode: 438, // == octal 0666
+      autoClose: true,
+      bufferSize: 4096,
+    };
```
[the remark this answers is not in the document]
zherczeg: What about using an empty object here?

`options = options || {}`

src/js/fs.js
```diff
@@ -382,6 +382,182 @@ fs.readdirSync = function(path) {
...
+        if (err) {
+          throw err;
+        }
+        self._fd = _fd;
+        self.emit('open', self._fd);
+        self.doRead();
+      });
+    }
+
+    this.on('open', function(/* _fd */) {
```
zherczeg: What about using once here?

src/js/fs.js
```diff
@@ -382,6 +382,182 @@ fs.readdirSync = function(path) {
...
+    return new ReadStream(path, options);
+  };
+
+
+  function WriteStream(path, options) {
+    if (!(this instanceof WriteStream)) {
+      return new WriteStream(path, options);
+    }
+
+    options = options || {
```
zherczeg: `options = options || {}`

pmarkee:
Thanks for the review @zherczeg, I have updated the PR.

test/run_pass/test_fs_write_stream.js
```diff
@@ -0,0 +1,55 @@
...
+
+writableFileStream.on('ready', function() {
+  writableFileStream.write(testData);
+});
+
+var buff = new Buffer(64);
+writableFileStream.on('close', function() {
+  // Check output correctness
+  fs.open(outputFile, 'r', 0666, function(err, fd) {
+    console.log(outputFile);
```
galpeter: Do we need the `console.log` call here?

docs/api/IoT.js-API-File-System.md
```diff
@@ -112,6 +112,161 @@ fs.closeSync(fd);
...
+
+var wStream = fs.createWriteStream('example.txt');
+
+wStream.on('ready', function() {
+  wStream.write('test data');
+  // 'test data' will be written into example.txt
+});
+```
+
+Note: `fs.WriteStream` inherits from `stream.Writable`, so `_readyToWrite()`
```
galpeter: Is this still relevant? the documentation does not use it all.

docs/api/IoT.js-API-File-System.md
```diff
@@ -112,6 +112,161 @@ fs.closeSync(fd);
...
+If it is `true` (as by default), the file will be closed automatically
+when end of file is reached or the stream ends.
+
+**Example**
+
+```js
+var fs = require('fs');
+
+var rStream = fs.createReadStream('example.txt');
+rStream.on('data', function(data) {
```
galpeter: The example uses the `data` event, but it is not documented at all here. However there are other events which are documented. It would be good to have a bit of info about the `data` also (if I remember correctly this is inherited from the readeble stream).

docs/api/IoT.js-API-File-System.md
```diff
@@ -112,6 +112,161 @@ fs.closeSync(fd);
 ```
 
 
+## Class: fs.ReadStream
```
galpeter: The new elements are not documented at the start of the file where the API table is.

src/js/fs.js
```diff
@@ -382,6 +382,164 @@ fs.readdirSync = function(path) {
...
+      return new ReadStream(path, options);
+    }
+
+    options = options || {};
+
+    Readable.call(this, {defaultEncoding: options.encoding || null});
+
+    this.bytesRead = 0;
+    this.path = path;
+    this._autoClose = options.autoClose || true;
```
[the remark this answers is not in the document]
pmarkee: `autoClose` has to be true if it is not specified so I think we should do it like this:
```
this.autoClose = util.isNullOrUndefined(options.autoClose) || options.autoClose;
```

pmarkee:
Thanks @galpeter @zherczeg , I have updated the PR.

PR #1791, opened 2018-11-07
Updated JerryScript submodule.

LaszloLango:


LaszloLango:
@robertsipka @tdusnoki the update contains debugger updates. The debugger version will upgrade from 5 to 7 with this patch. Can you check the IoT.js Code and VSCode extension? I think we should land this update when they are ready to handle it correctly.

robertsipka:
@LaszloLango : The current master on VSCode extension is handle that debugger version.

PR #1795, opened 2018-11-12
tizenrt: Disable debugger code if not enabled

rzr:
Observed build issue on TizenRT (2.0+):

    iotjs.c:58: undefined reference to `jerryx_debugger_tcp_create'
    iotjs.c:59: undefined reference to `jerryx_debugger_ws_create'
    iotjs.c:58: undefined reference to `jerryx_debugger_after_connect'

The whole part is disabled,
even if only jerry-ext functions are not linked.

This should not cause any issue on earlier versions of TizenRT

Change-Id: Id88c745c9712f6f620bbd200f493397572174ce6
Forwarded: https://github.com/Samsung/iotjs/pull/
Origin: https://github.com/Samsung/iotjs/pull/1749
IoT.js-DCO-1.0-Signed-off-by: Philippe Coval philippe.coval@osg.samsung.com

PR #1798, opened 2018-11-14
debugger: Disable code if not desired

rzr:
This part is not as critical for TizenRT,
but will help to make iotjs smaller.

Thanks-to: Daeyeon Jeong <daeyeon.jeong@samsung.com>
Relate-to: https://github.com/Samsung/iotjs/pull/1795

### DIFF
--- a/src/modules/iotjs_module_process.c
+++ b/src/modules/iotjs_module_process.c
@@ -41,11 +41,13 @@ JS_FUNCTION(Compile) {
   iotjs_string_t source = JS_GET_ARG(1, string);
 
   const char* filename = iotjs_string_data(&file);
-  const iotjs_environment_t* env = iotjs_environment_get();
 
+#ifdef JERRY_DEBUGGER
+  const iotjs_environment_t* env = iotjs_environment_get();
   if (iotjs_environment_config(env)->debugger != NULL) {
     jerry_debugger_stop();
   }
+#endif
 
   jerry_value_t jres =
       WrapEval(filename, strlen(filename), iotjs_string_data(&source),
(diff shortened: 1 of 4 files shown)


rzr:
Please 1st merge:
https://github.com/Samsung/iotjs/pull/1795
before:
https://github.com/Samsung/iotjs/pull/1798
Thanks

src/modules/iotjs_module_process.c
```diff
@@ -383,10 +390,12 @@ jerry_value_t InitProcess() {
   // Set iotjs
   SetProcessIotjs(process);
   bool wait_source = false;
+#ifdef JERRY_DEBUGGER
```
LaszloLango: IMHO, this way would be better:

```c
#ifdef JERRY_DEBUGGER
  bool wait_source = false;
  if (iotjs_environment_config(iotjs_environment_get())->debugger != NULL) {
    wait_source = iotjs_environment_config(iotjs_environment_get())
                      ->debugger->wait_source;
  }

  if (!wait_source) {
    SetProcessArgv(process);
  }
#else
  SetProcessArgv(process);
#endif
```

rzr: isnt compiler figuring it out itself and strip that unused var ?

daeyeon:
Please squash the commits into one.

src/modules/iotjs_module_process.c
rzr: and this same var is used just after so I prefer to declare it false that duplicating the 2 lines

I assume compiler is smart enough to generate the same binary.

PR #1800, opened 2018-11-15
Inherit OutgoingMessage from stream.Writable to support pipe() usage

pmarkee:


pmarkee:
For example: https://github.com/Samsung/iotjs-samples/pull/20/files#diff-e91606e01d14be39dfb13edfbf3e0691R42
In this example we have a http server and a client. The client requests a file, and the server can simply `pipe()` the contents of the requested file into `response`, which is an `OutgoingMessage`.

PR #1780, opened 2018-10-22
WIP: build: Update TizenRT to 2.0_Public_M2

rzr:
Because current version of TizenRT in IoT.js docker image is oudated,
let's fetch tag from upstream at build time,
to speed up build time, this part can be eventually removed
if image is updated with latest TizenRT's sources.

Use iotjs's defconfig as default config instead of file:
"tizenrt_release_config"
from container:
https://hub.docker.com/r/iotjs/ubuntu/

Is this file tracked anywhere?

Later debug, release configs can tracked in IoT.js's source tree,
and used as desired, for performance benchmarking.

jerryx extra lib has been dropped for now,
this should be reintroduced later (once TizenRT upgrade IoT.js ?)

Bug: https://github.com/Samsung/iotjs/issues/1777
IoT.js-DCO-1.0-Signed-off-by: Philippe Coval p.coval@samsung.com

Issue #1777, opened 2018-10-19
Support TizenRT 2.0


daeyeon:
A new TizenRT public release is recently released. It seems worth considering to support it although it's still in its second of the milestone. I think that the followings are required things in priority order.

- [ ] Change the TizenRT base to the tag: `2.0_Public_M2` and update IoT.js.
- [ ] Update TravisCI test
- [ ] Update RemoteTest using a real target
- [ ] Update documents
- [ ] Make a snapshot PR to TizenRT repo.

rzr:
Good Idea I can help on this, for last last item track:
https://github.com/Samsung/TizenRT/pull/2018

rzr:
Update:

- [x] Change the TizenRT base to the tag: 2.0_Public_M2 and update IoT.js.

Travis is ok, image can be eventually updated, who can help on that ?
same for remote tests?

For last part: 
This extra change is also welcome for TizenRT master (post 2.0):
https://github.com/Samsung/iotjs/pull/1808

PR #1805, opened 2018-11-20
tizenrt: Use TizenRT version to 2.0 update defconfigs

rzr:
Align to TizenRT's default config, but disable SSS for ARTIK053
also preserve debug options for debug configuration.

Because current version of TizenRT in IoT.js docker image is outdated,
let's fetch tag from upstream at build time,
to speed up build time, this part can be eventually removed
if image is updated with latest TizenRT's sources.

jerryx extra lib has been dropped for now,
this should be reintroduced later (once TizenRT upgraded IoT.js ?)

Bug: https://github.com/Samsung/iotjs/issues/1777
Forwarded: https://github.com/Samsung/iotjs/pull/1780


tools/travis_script.py
```diff
@@ -151,10 +152,8 @@ def build_iotjs(buildtype, args=[], env=[]):
             exec_docker(DOCKER_TIZENRT_OS_PATH, [
                         'make', 'IOTJS_ROOT_DIR=' + DOCKER_IOTJS_PATH,
                         'IOTJS_BUILD_OPTION='
-                        '--profile=test/profiles/tizenrt.profile',
```
daeyeon: Any reason we should drop jerry-ext? It seems better to follow the above FIXME.

LaszloLango: I think it can be deleted because the `jerry-debugger` is not enabled by default, so `jerry-ext` is not needed after #1798

LaszloLango: (If we keep this change, then the comment must be deleted as well, because it is not valid after this patch)

rzr:
Yes this part was removed because lib is no more built:

https://travis-ci.org/Samsung/iotjs/jobs/4596535933#

arm-none-eabi-ld: cannot find -ljerry-ext


I am updating patch and remove the header comment as suggested

PR #1812, opened 2018-11-30
Modify test_fs_write_stream.js to use writable memory area on TizenRT.

rtakacs:
On TizenRT, only the `/mnt` folder is writable.

### DIFF
--- a/test/run_pass/test_fs_write_stream.js
+++ b/test/run_pass/test_fs_write_stream.js
@@ -20,7 +20,8 @@ if (fs.createReadStream === undefined || fs.createWriteStream === undefined) {
   process.exit(0);
 }
 
-var outputFile = process.cwd() + '/tmp/test_fs5.txt';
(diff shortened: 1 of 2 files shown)


test/run_pass/test_fs_write_stream.js
```diff
@@ -20,9 +20,13 @@ if (fs.createReadStream === undefined || fs.createWriteStream === undefined) {
   process.exit(0);
 }
 
-var outputFile = process.cwd() + '/tmp/test_fs5.txt';
-var testData = 'WriteStream test';
+if (process.platform === 'tizenrt') {
+  var outputFile = '/mnt/test_fs5.txt';
+} else {
+  var outputFile = process.cwd() + '/tmp/test_fs5.txt';
+}
```
galpeter: Would it make sense to do this?:
```js
var prefix = (process.platform === 'tizenrt') ? '' : process.cwd();
var outFile = prefix + '/tmp/test_fs5.txt/';
```

rtakacs: @galpeter nice!

PR #1814, opened 2018-12-03
Fix an issue with writable streams on nuttx platform

DanielBallaSZTE:
There's an issue on nuttx systems causing a Buffer to be converted to a string everytime a `Writable stream` ends. This is obviously not a good practice, especially when working with binary buffers.
This patch fixes the issue by differentiating Buffers from strings.

### DIFF
--- a/src/js/net.js
+++ b/src/js/net.js
@@ -25,6 +25,13 @@ function createTCP() {
   return _tcp;
 }
 
+// Expected end message on nuttx platform.
+var expectedEnding;
+
+if (process.platform == 'nuttx') {
+   expectedEnding = new Buffer('\\e\\n\\d');
+}
+
 
 function SocketState(options) {
   // 'true' during connection handshaking.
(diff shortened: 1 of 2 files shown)


DanielBallaSZTE:
@zherczeg Please review again, I pushed another fix to `net.js`

src/js/net.js
```diff
@@ -403,15 +403,20 @@ function onread(socket, nread, isEOF, buffer) {
       return;
     }
 
-    var str = buffer.toString();
+    // We know for sure the last 6 characters are going to be the ending
```
rtakacs: Missing dot character at the end of the sentence ;)

src/js/net.js
```diff
@@ -403,15 +403,20 @@ function onread(socket, nread, isEOF, buffer) {
...
+    // Lets create a buffer with those 6 characters without toString conversion.
+    var eofLength = 6;
+    var bufferLength = buffer.length;
+    var ending = new Buffer(eofLength);
+    buffer.copy(ending, 0, bufferLength - eofLength, bufferLength);
+    var expectedEnding = new Buffer('\\e\\n\\d');
     var eofNeeded = false;
-    if (str.length >= 6
-      && str.substr(str.length - 6, str.length) == '\\e\\n\\d') {
+    if (bufferLength >= 6 && ending.compare(expectedEnding) == 0) {
```
rtakacs: The magic number can be replaced to `eofLength`.

src/js/net.js
```diff
@@ -403,15 +403,20 @@ function onread(socket, nread, isEOF, buffer) {
...
     }
 
-    var str = buffer.toString();
+    // We know for sure the last 6 characters are going to be the ending
+    // Lets create a buffer with those 6 characters without toString conversion.
+    var eofLength = 6;
+    var bufferLength = buffer.length;
+    var ending = new Buffer(eofLength);
+    buffer.copy(ending, 0, bufferLength - eofLength, bufferLength);
+    var expectedEnding = new Buffer('\\e\\n\\d');
```
galpeter: Do we really need to create this buffer every time the `onread` is triggered?

src/js/net.js
```diff
@@ -403,15 +403,20 @@ function onread(socket, nread, isEOF, buffer) {
       return;
     }
 
-    var str = buffer.toString();
+    // We know for sure the last 6 characters are going to be the ending
+    // Lets create a buffer with those 6 characters without toString conversion.
+    var eofLength = 6;
+    var bufferLength = buffer.length;
+    var ending = new Buffer(eofLength);
+    buffer.copy(ending, 0, bufferLength - eofLength, bufferLength);
```
galpeter: So this is basically a slice?

src/js/net.js
galpeter: Would it be feasible to read the last bytes from the buffer and compare their values? That way we would not need extra buffers.

src/js/net.js
DanielBallaSZTE: I moved it to a global value, good point

DanielBallaSZTE:
@galpeter PR updated, thank you for the suggestions.

rtakacs:
LGTM (informal)

PR #1815, opened 2018-12-04
Fix typo in fs.ReadStream and fs.WriteStream

pmarkee:


pmarkee:
There was a typo in both fs stream classes, so the streams didn't close properly.

PR #1817, opened 2018-12-04
Enable crypto, http-signature, websocket modules.

rtakacs:
Note: on NuttX, the http-signature and websocket cannot be enabled because of mbedtls dependency.


rtakacs:
Please wait for #1816.

PR #1825, opened 2019-01-02
Avoid possible null pointer dereferencing in tcp connection

robertsipka:
The `iotjs_jval_get_object_native_handle` function call might return with NULL, so there is a possible null pointer dereferencing in the `uv_accept` call.


rtakacs:
LGTM (informal)

PR #1826, opened 2019-01-03
Introduce a minimal REPL for IoT.js

galpeter:
The change adds a REPL tool which can be used for testing purposes.
It is not intended to be used as a module at this point.

### DIFF
--- a/tools/repl.js
+++ b/tools/repl.js
@@ -1,0 +1,110 @@
+/* Copyright 2019-present Samsung Electronics Co., Ltd. and other contributors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


tools/repl.js
```diff
@@ -0,0 +1,110 @@
...
+  fs.writeSync(1, this._prompt_msg, 0, this._prompt_msg.length);
+};
+
+REPL.prototype.run_code = function(line) {
+  var result;
+  try {
+    result = eval.call(this._repl_context, line);
+    console.log(result);
+  } catch (ex) {
+    console.log(ex);
```
LaszloLango: What do you think about using `console.err()` here?

PR #1828, opened 2019-01-08
Travis CI config maintenance

akosthekiss:
- `sudo` key has been deprecated.
- Added proper name to the jobs (instead of having it in an
  environment variable).
- Removed a legacy workaround from the Coverity Scan job as it
  isn't necessary to fiddle with the cerificates of
  scan.coverity.com anymore.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,82 +2,91 @@ language: c
 
 os: linux
 dist: trusty
-sudo: false
 
 services:
   - docker
 
-before_install:
-  - if [[ "$RUN_DOCKER" == "yes" ]]; then docker pull iotjs/ubuntu:0.9; fi
-
-script:
-  tools/travis_script.py
+before_install: if [[ "$RUN_DOCKER" == "yes" ]]; then docker pull iotjs/ubuntu:0.9; fi
+script: tools/travis_script.py
 
 matrix:
   include:
-    - env:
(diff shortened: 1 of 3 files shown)


yichoi:
Could you please update ``sonar-project.properties`` also as like the one in the pando-project/escargot ?

akosthekiss:
@yichoi I've updated the PR as suggested.

akosthekiss:
Would it be possible to have the Sonar-related changes in a separate PR? There are multiple changes that need to be applied:
- The `addons.sonarcloud.organization` key in `.travis.yml` needs to be updated. (As mentioned.)
- The `sonar.projectKey` in `sonar-project.properties` needs to be updated. (As mentioned.)
- The links in `README.md` also need to be updated.
- But the project also has to be created at sonarcloud under pando-project. Right now, https://sonarcloud.io/dashboard?id=pando-project_iot.js gives me "The requested project does not exist."

Because the sonar-related changes seem like an atomic changeset to me, I wonder whether those changes that are in the current commit (to `sonar-project.properties` and `check_sonarqube.sh`) should also go with this new PR?

yichoi:
@akosthekiss I fine with your plan. BTW "https://sonarcloud.io/dashboard?id=pando-project_iot.js" will appear after when the first analysis is completed.

PR #1838, opened 2019-02-07
Fix unhandled promise error in iotjs.c

DanielBallaSZTE:
There was an issue not freeing and checking if a promise value returned with an error.
This patch fixes it.


LaszloLango:
@DanielBallaSZTE Is it possible to add a test for this issue? Can it be reproduced with a small example?

DanielBallaSZTE:
@LaszloLango At the moment our promise implementation can not throw such an error, therefore I can not provide a test, however in the future it will definitely change, and will cause issues if it stays like it currently is.

PR #1839, opened 2019-02-15
Removed an unnecessary 'do { } while(0);' from HTTP parser module.

LaszloLango:
Usually 'do { } while(0)' without a semicolon is used in a macro when we want
to force the semicolon for the certain macro. The compiler does fails if the
developer forgot the semicolon after the macro call. Using this technique with
a semicolon after the 'while' has no sense ('...while(0);').

### DIFF
--- a/src/modules/iotjs_module_http_parser.c
+++ b/src/modules/iotjs_module_http_parser.c
@@ -457,14 +457,12 @@ static void http_parser_register_methods_object(jerry_value_t target) {
 
   jerry_value_t method_name;
   uint32_t idx = 0;
-#define V(num, name, string)                                         \
-  do {                                                               \
-    method_name = jerry_create_string((const jerry_char_t*)#string); \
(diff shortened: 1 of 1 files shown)


LaszloLango:
@daeyeon, unfortunately we cannot just drop the semicolon here. The implementation of `HTTP_METHOD_MAP` makes it impossible, but it is in the `http-parser` sub-module. Furthermore it cannot be modified, because would lead into other errors (eg.: semicolon between elements of enum, which is a syntax error).  We have some different options:
1) leave as is and close the PR
2) Do not use the `HTTP_METHOD_MAP`, but define a new one in `http-parser` with a proper semicolon usage. This wouldn't be too good, because it leads to code duplications and makes worse the readability of the code.
3) Use empty block (`{ }`) instead of `do { } while(0);`
4) Remove the `do { } while(0);` (as this PR does).

PR #1843, opened 2019-03-07
Modified the exports of the event module.

LaszloLango:
Both `require('events')` and `require('events').EventEmitter`
should return with the EventEmitter like in NodeJS. This improves
the compatibility with existing third-party NodeJS modules.

### DIFF
--- a/src/js/events.js
+++ b/src/js/events.js
@@ -21,7 +21,8 @@ function EventEmitter() {
   this._events = {};
 }
 
-module.exports.EventEmitter = EventEmitter;
+module.exports = EventEmitter;
+EventEmitter.EventEmitter = EventEmitter;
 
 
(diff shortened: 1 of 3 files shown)


src/js/events.js
```diff
@@ -21,7 +21,8 @@ function EventEmitter() {
   this._events = {};
 }
 
-module.exports.EventEmitter = EventEmitter;
+module.exports = EventEmitter;
```
galpeter: Based on the current NodeJs api, this is the preferred way to access the EventEmmiter. Thus we should also update the documentation. (And maybe the internal module also in a later PR)

LaszloLango: Good point, I forgot the documentation.

LaszloLango:
@galpeter I've updated the documentation, please check.

PR #1849, opened 2019-04-04
Add N-API async support

DanielBallaSZTE:
Based on previous work of Rokid Co., Ltd. (https://github.com/yodaos-project/ShadowNode)

### DIFF
--- a/src/napi/node_api_async.c
+++ b/src/napi/node_api_async.c
@@ -1,0 +1,169 @@
+/* Copyright 2019-present Samsung Electronics Co., Ltd. and other contributors
+ * Copyright 2018-present Rokid Co., Ltd. and other contributors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "uv.h"
+#include "internal/node_api_internal.h"
+
+static void iotjs_uv_work_cb(uv_work_t* req) {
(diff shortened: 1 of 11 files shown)


src/napi/node_api_async.c
```diff
@@ -0,0 +1,169 @@
...
+     * napi_async_work could be deleted by invocation of `napi_delete_asyncwork`
+     * in its complete callback.
+     */
+    complete(env, cb_status, data);
+    jerryx_close_handle_scope(scope);
+
+    if (iotjs_napi_is_exception_pending(env)) {
+      jerry_value_t jval_err;
+      jval_err = iotjs_napi_env_get_and_clear_exception(env);
+      if (jval_err == (uintptr_t)NULL) {
```
LaszloLango: This check is not correct. You cannot compare a `jerry_value_t` to a pointer. This leads to undefined behavoir, but with current JerryScript it is probably always false.

DanielBallaSZTE: `iotjs_napi_env_get_and_clear_exception` returns a `jerry_value_t`.
The function gets the current environments' error value with the `AS_JERRY_VALUE` macro. If it is `NULL`(`0x0`), the macro casts it to an `uintptr_t` then to a `jerry_value_t` which will hold the value of 0. Comparing `0` to `(uintptr_t)NULL` results in basically comparing `0` to `0`. If the `jerry_value_t` actually holds an error, the comparison returns false, since a jerry_value_t will guaranteed always have a value other than `0`.

LaszloLango: IMHO it is not a safe usage of the JerryScript API. `iotjs_napi_env_get_and_clear_exception` should be updated to use `jerry_create_undefined` or `jerry_create_null`.

DanielBallaSZTE: I can definitely do an update then.
Just a note:
There are other locations in the code (ie: node_api_function.c) using the same method.

LaszloLango: @DanielBallaSZTE we should fix them in a followup patch.

legendecas: Yes, it's unsafe. But there is no `Empty` value in JerryScript AFAIK, it's just a not so even bad choice.

Also note that we could throw any value in JS so as `undefined` and `null`.

LaszloLango: You convinced me. Leave as is for now.

PR #1854, opened 2019-04-15
fix an error Build-for-x86-Linux.md

coloka:
changed command line parameter  from '--memstat' to '--mem-stats'.


coloka:
add more sign-off messages.

PR #1855, opened 2019-04-16
Implemented 'napi_create_symbol' and added ES2015 related tests for N-API

LaszloLango:
IoT.js-DCO-1.0-Signed-off-by: László Langó llango.u-szeged@partner.samsung.com

PR #1856, opened 2019-04-16
Add Promise support to N-API

DanielBallaSZTE:

### DIFF
--- a/src/napi/node_api_value.c
+++ b/src/napi/node_api_value.c
@@ -16,10 +16,17 @@
 
 #include "jerryscript-ext/handle-scope.h"
 #include "jerryscript.h"
+#include <stdlib.h>
 #include "internal/node_api_internal.h"
 #include "modules/iotjs_module_buffer.h"
 
 #include <math.h>
+
+/* Feature missing error messages */
+const char* napi_err_no_promise = "Promise is not supported by this build.";
+const char* napi_err_no_symbol = "Symbols are not supported by this build.";
+const char* napi_err_invalid_deferred =
+    "Invalid deferred object. Please refer to the documentation.";
 
 static void iotjs_napi_buffer_external_free_cb(void* native_p) {
   iotjs_buffer_external_info_t* info = (iotjs_buffer_external_info_t*)native_p;
@@ -250,8 +257,7 @@ napi_status napi_create_symbol(napi_env env, napi_value description,
 
   if (!jerry_is_feature_enabled(JERRY_FEATURE_SYMBOL)) {
     NAPI_ASSIGN(result, NULL);
-    NAPI_RETURN(napi_generic_failure,
-                "Symbols are not supported by this build.");
+    NAPI_RETURN(napi_generic_failure, napi_err_no_symbol);
   }
 
   JERRYX_CREATE(jval, jerry_create_symbol(AS_JERRY_VALUE(description)));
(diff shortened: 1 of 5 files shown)


src/napi/node_api_value.c
```diff
@@ -483,3 +484,53 @@ napi_status napi_strict_equals(napi_env env, napi_value lhs, napi_value rhs,
...
+
+napi_status napi_create_promise(napi_env env, napi_deferred* deferred,
+                                napi_value* promise) {
+  jerry_value_t jpromise = jerry_create_promise();
+  if (jerry_value_is_error(jpromise)) {
+    NAPI_INTERNAL_CALL(napi_throw(env, AS_NAPI_VALUE(jpromise)));
+    NAPI_RETURN(napi_pending_exception);
+  }
+  napi_assign_nvalue(jpromise, promise);
+  *deferred = malloc(sizeof(napi_value*));
```
legendecas: out value in napi_* functions could be null though the case might mean nothing.

test/testsets.json
```diff
@@ -1298,6 +1298,16 @@
         "napi"
       ]
     },
+    {
+      "name": "test_napi_promise.js",
+      "required-modules": [
+        "napi"
+      ],
+      "skip": [
```
LaszloLango: No need to always skip this test, just add `"required-features": [ "Promise" ]`

src/napi/node_api_value.c
DanielBallaSZTE: Do you mean the user can give a non-NULL ptr as the deferred value?

DanielBallaSZTE: The API requests the user to do the following:
create a `napi_deferred deferred;`
then pass its memory address to the `napi_create_promise` function. It then creates the deferred object which will be freed with `napi_resolve/reject_deferred` function.

src/napi/node_api_value.c
```diff
@@ -483,3 +484,53 @@ napi_status napi_strict_equals(napi_env env, napi_value lhs, napi_value rhs,
 
   return napi_assign_bool(jerry_get_boolean_value(is_equal), result);
 }
+
+napi_status napi_create_promise(napi_env env, napi_deferred* deferred,
+                                napi_value* promise) {
+  jerry_value_t jpromise = jerry_create_promise();
```
LaszloLango: What happens when promise is not supported by the build? IMHO we should use `jerry_is_feature_enabled` in these functions.

src/napi/node_api_value.c
LaszloLango: I agree with @legendecas, guard the assignment to prevent null pointer dereference.

src/napi/node_api_value.c
LaszloLango:
```c
  if (!jerry_is_feature_enabled(JERRY_FEATURE_PROMISE)) {
    NAPI_ASSIGN(result, NULL);
    NAPI_RETURN(napi_generic_failure,
                "Promise is not supported by this build.");
  }
```

DanielBallaSZTE: If promise is not supported JerryScript simply returns an error value.

LaszloLango: It looks like we have multiple choices of throwing an exception. I used the `napi_generic_failure` in #1855 (see my previous comment). It looks less costly then your method. Which one do you prefer?

DanielBallaSZTE: If you used that one I think we can stick to that, so it will be consistent. I'll update my PR

PR #1862, opened 2019-04-24
Add debug build option to napi modules

DanielBallaSZTE:


src/modules/iotjs_module_process.c
```diff
@@ -383,6 +383,21 @@ jerry_value_t InitProcess(void) {
...
+
+  prop_desc.is_value_defined = true;
+  prop_desc.is_writable_defined = true;
+  prop_desc.is_configurable_defined = true;
+  prop_desc.value = jerry_create_boolean(true);
+
+  jerry_value_t property_name =
+      jerry_create_string((jerry_char_t*)IOTJS_MAGIC_STRING_DEBUG);
+  jerry_value_t prop_ret_val =
+      jerry_define_own_property(process, property_name, &prop_desc);
```
LaszloLango: Why don't you use `iotjs_jval_set_property_boolean` instead?

DanielBallaSZTE: Because the user could overwrite its value, which I don't think would be a good practice. This way the value is not writable, nor configurable.

src/iotjs_magic_strings.h
```diff
@@ -103,6 +103,7 @@
 #if ENABLE_MODULE_UART
 #define IOTJS_MAGIC_STRING_DATABITS "dataBits"
 #endif
+#define IOTJS_MAGIC_STRING_DEBUG "debug"
```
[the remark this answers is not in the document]
DanielBallaSZTE: True, thanks

PR #1874, opened 2019-05-08
Updated JerryScript submodule.

LaszloLango:
IoT.js-DCO-1.0-Signed-off-by: László Langó llango.u-szeged@partner.samsung.com

PR #1875, opened 2019-05-09
Fixed N-API support on Windows

LaszloLango:


LaszloLango:
Note: #1874 is also needed to make N-API work on Windows platform.

cmake/iotjs.cmake
```diff
@@ -504,16 +504,24 @@ else()
...
+    if(USING_MSVC)
+      foreach(NODE_SYMBOL ${NODE_SYMBOLS})
+        set(NODE_SYMBOLS_LINK_FLAGS
+            "${NODE_SYMBOLS_LINK_FLAGS} /INCLUDE:_${NODE_SYMBOL}")
+      endforeach()
+    else()
+      set(NODE_SYMBOLS_LINK_FLAGS "-Wl")
+      foreach(NODE_SYMBOL ${NODE_SYMBOLS})
+        set(NODE_SYMBOLS_LINK_FLAGS
+            "${NODE_SYMBOLS_LINK_FLAGS},-u,${NODE_SYMBOL}")
```
galpeter: hmm maybe we could reduce the duplication a bit if the `,-u,` and `/INCLUDE:_` strings are from variables, thus we would end up with something like this:

```
if (MSVC)
  set(LINKER_SYMBOL_INCLUDE "/INCLUDE:_")
else()
  set(LINKER_SYMBOL_INCLUDE ",-u,")
endif()
foreach(NODE_SYMBOL ${NODE_SYMBOLS})
        set(NODE_SYMBOLS_LINK_FLAGS
            "${NODE_SYMBOLS_LINK_FLAGS}${LINKER_SYMBOL_INCLUDE}${NODE_SYMBOL}")
endforeach()
```

src/iotjs_util.c
```diff
@@ -115,7 +115,8 @@ void iotjs_buffer_release(char* buffer) {
 }
 
 void print_stacktrace(void) {
-#if defined(__linux__) && defined(DEBUG) && !defined(__OPENWRT__)
+#if !defined(NDEBUG) && defined(__linux__) && defined(DEBUG) && \
```
galpeter: How is this related to the windows change?

src/js/module.js
```diff
@@ -275,7 +275,12 @@ Module.load = function(id, parent) {
     source = Builtin.readSource(modPath);
     module.exports = JSON.parse(source);
   } else if (dynamicloader && ext === 'node') {
-    module.exports = dynamicloader(modPath);
+    if (process.platform === 'windows') {
+      module.exports = dynamicloader(modPath.replace(new RegExp('/', 'g'),
```
galpeter: Why is this needed, windows uses the `/` for paths?

cmake/iotjs.cmake
LaszloLango: This won work correctly, because `set(NODE_SYMBOLS_LINK_FLAGS "-Wl")` is also needed for the start of the linker command in case of GCC.

src/iotjs_util.c
LaszloLango: There were a code duplication in `napi_fatal_error`. I changed it to use this function, but this change is needed to avoid linker error in release builds.

src/js/module.js
LaszloLango: [`LoadLibrary`](https://docs.microsoft.com/en-us/windows/desktop/api/libloaderapi/nf-libloaderapi-loadlibrarya) WINAPI function only works with windows paths and it it easier to replace the forward slashes in JS. Quote from the reference:

> When specifying a path, be sure to use backslashes (), not forward slashes (/). For more information about paths, see [Naming a File or Directory](https://msdn.microsoft.com/121cd5b2-e6fd-4eb4-99b4-b652d27b53e8).

LaszloLango:
@galpeter thanks for the review. I've updated PR.

src/js/module.js
galpeter: I see. then the RegExp should be only constructed once and not every time the require is called.

LaszloLango: You are right. I updated the `path` object with `toUnixPath` and `toWindowsPath`.

PR #1881, opened 2019-05-17
Unified function and variable names in C source codes.

LaszloLango:
'CamelCase' and 'underscore_case' were mixed together, but
'unerscores_case' should be used always for function and
variable names in the C codes. Most of the exiting code path
used underscores. This patch fixes the remained wrong cases.

### DIFF
--- a/src/modules/iotjs_module_adc.c
+++ b/src/modules/iotjs_module_adc.c
@@ -47,7 +47,7 @@ static void adc_worker(uv_work_t* work_req) {
   }
 }
 
-JS_FUNCTION(AdcCons) {
+JS_FUNCTION(adc_constructor) {
   DJS_CHECK_THIS();
   DJS_CHECK_ARGS(1, object);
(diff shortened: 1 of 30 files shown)


LaszloLango:
@rtakacs @Yuyupo could you test this PR on the supported devices?

pmarkee:
I have tried peripheral (gpio, i2c etc) tests on rpi2+Linux, it runs the same as the current `master`, so this patch should be fine.

src/modules/iotjs_module_adc.c
```diff
@@ -47,7 +47,7 @@ static void adc_worker(uv_work_t* work_req) {
   }
 }
 
-JS_FUNCTION(AdcCons) {
+JS_FUNCTION(adc_onstructor) {
```
hs0225: Please correct the typos.(adc_onstructor ->adc_constructor)

LaszloLango: Good catch. I've fixed it.

PR #1891, opened 2019-06-14
improve and bugfix websocket module

jan4984:


src/modules.json
```diff
@@ -395,7 +395,7 @@
                        "modules/iotjs_module_websocket.c"],
       "init": "iotjs_init_websocket",
       "js_file": "js/websocket.js",
-      "require": ["crypto", "events", "net", "util"]
+      "require": ["crypto", "events", "net", "util", "tls"]
```
galpeter: If I remember correctly the websocket can be used over simple net connections also ("ws://") and for that there is no tls requirement. Did you have a sample where the unsecure websocket connection was not working without the tls module? Of course having using websocket with tls should be the norm but having a hard requirement on tls (that is: the tls could not be turned off) would increase the binary size if the tls in not required in a target application.

src/modules/iotjs_module_websocket.c
```diff
@@ -601,7 +601,8 @@ JS_FUNCTION(ws_receive) {
       if (current_buffer + sizeof(uint16_t) > current_buffer_end) {
         break;
       }
-      payload_len = (uint16_t)(current_buffer[0] << 8 | current_buffer[1]);
+      unsigned char* len_buffer = (unsigned char*)current_buffer;
```
galpeter: The commit message ("converting singed char to unsigned short changes the value") does not describe why was this needed. I guess this is needed because shifting signed values would cause incorrect calculations thus these values should be unsinged?

galpeter: Also there is a bit of style problem: before the `*` there should be a single space (please see travis log: https://travis-ci.org/pando-project/iotjs/jobs/545645417#L841 )

src/modules.json
jan4984: `websocket.js` always runs to fail because it require `tls` at start of source and `tls` native module not found.  I agree it's better to let `websocket` require `tls` only in `wss` scheme, we should change `websocket.js`.

galpeter: Ohh... I've misunderstood somewhere. I've just checked the websocket without tls and there is indeed a bit of problem. What we can do is add a `try-catch` when the tls module is required (just like in the module.js) and also fix the tls usages in the websocket.js. Could you update the PR or should I create one with the fix?

jan4984: OK I re-made the two commit

jan4984:
not know why ci not work...

galpeter:
It seems that the project was moved and the CI was not reactivated (yet).

src/js/websocket.js
```diff
@@ -16,7 +16,13 @@
  var net = require('net');
  var util = require('util');
  var EventEmitter = require('events').EventEmitter;
- var tls = require('tls');
+ var tls = (function() {
```
galpeter: There is no need for the function. We can do the following:
```
var tls;
try {
  tls = require(...);
} catch (e) {
  tls = {};
}
```

Could you update the PR to remove the function?

src/js/websocket.js
```diff
@@ -16,7 +16,12 @@
  var net = require('net');
  var util = require('util');
  var EventEmitter = require('events').EventEmitter;
- var tls = require('tls');
+ var tls;
+ try {
+   tls = require(...);
```
galpeter: In my previous comment the `...` was meant to represent that you need to replace it with the `'tls'` string. This is a syntax error currently.  It should be: `require('tls')`

daeyeon:
@jan4984 Thank you for your contribution. By the way, can you please squash the two commits into a single commit? It's currently asked to squash multiple commits down to one, or else discrete changesets before submitting a pull request.

PR #1892, opened 2019-06-14
build: Fix mistake while testing TARGET_BOARD

rzr:
Preprocessor can't use == on strings",
so new symbols are introduced to test configurations
at build time.

Note, this mistake was harmless
because there weren't any spefic parts among boards,
but this will change in upcoming change.

Change-Id: I45d87cccc086af05661eaf7b762a4a0274fb2ede
Relate-to: https://github.com/rzr/webthing-iotjs/issues/3


galpeter:
Thanks! LGTM

PR #1902, opened 2019-06-27
Add a "Hello IoT.js!" program to the tests

akosthekiss:
The current code base does not contain any "first JavaScript
program" for those who build IoT.js. The existing tests are just
that: tests, giving passed/failed information but nothing eye candy.

This patch adds a simple JS script that can act as a warm welcome
to anyone after the first successful build trying to run something
proving that IoT.js works.


LaszloLango:
console module is always required by default just like in NodeJS

akosthekiss:
@LaszloLango True. I've updated the PR by removing the superfluous require.

PR #1921, opened 2019-07-12
Improve WebSocket server handshake handling

galpeter:
The WebSocket server incorrectly used http headers:
* The 'Connection' header field value can have multiple elements, not just 'Upgrade'.
* The header field names should be handled in a case-insensitive way.
* If there is an error, reporting a 400 error should also terminate the connection
  and correctly add http header termination lines.

### DIFF
--- a/src/js/websocket.js
+++ b/src/js/websocket.js
@@ -88,33 +88,37 @@ function parseServerHandshakeData(data, client, server) {
   var res = data.split('\r\n');
   var method = res[0].split(' ');
 
-  var headers = { 'Connection': '',
-                  'Upgrade': '',
-                  'Host': '',
-                  'Sec-WebSocket-Key': '',
-                  'Sec-WebSocket-Version': -1,
+  // All header keys are converted to lower case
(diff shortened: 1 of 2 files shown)


src/js/websocket.js
```diff
@@ -89,33 +89,38 @@ function parseServerHandshakeData(data, client, server) {
...
 
   var response = '';
   if (method[0] === 'GET' &&
       method[2] === 'HTTP/1.1' &&
       method[1] === server.path &&
-      headers['Connection'] === 'Upgrade' &&
-      headers['Upgrade'] === 'websocket' &&
-      headers['Sec-WebSocket-Version'] === '13') {
+      headers['connection'] &&
+      headers['connection'].indexOf('Upgrade') !== -1 &&
```
daeyeon: The tokens, 'websocket' and 'Upgrade', looks also case-insensitive.

galpeter:
@daeyeon Thanks! I've updated the PR with lower-case conversion for the 'upgrade' and 'websocket' tokens.

PR #1848, opened 2019-03-29
Implementation of basics of N-API for building native addons

LaszloLango:
N-API Reference: https://nodejs.org/docs/latest-v10.x/api/n-api.html

Implemented features:
  * Basic N-API data Types
  * Error handling
  * Object lifetime management
  * Functions for create, convert and compare JS values
  * Functions to work with JavaScript properties
  * Object wrapping

Missing features:
  * ES2015 features (Symbol, TypedArrays, ArrayBuffer, Promise, etc.)
  * Asynchronous Operations and Thread-safe Function Calls
  * Implementation of 'napi_external' type

Based on previous work of Rokid Co., Ltd. (https://github.com/yodaos-project/ShadowNode)

Co-authors: Daniel Balla <dballa@inf.u-szeged.hu> and Istvan Miklos <imiklos2@inf.u-szeged.hu>
IoT.js-DCO-1.0-Signed-off-by: László Langó llango.u-szeged@partner.samsung.com

PR #1923, opened 2019-07-16
build: Install to system's path (prefixed by destdir)

rzr:
This was introduced first in #1134 and then droped in #1848

This will be helpful for packaging,
or example on Tizen install using RPM macro: "%make_install"
of Debian's debhelper.

Work along CMAKE_INSTALL_PREFIX,
or/and can be overloaded individualy to alternate locations.

Forwarded: https://github.com/Samsung/iotjs/pull/
Relate-to: https://github.com/Samsung/iotjs/pull/1134
Change-Id: Ib969a1d5c2e7bc402b455377fc57a94654aa74dc

### DIFF
--- a/cmake/iotjs.cmake
+++ b/cmake/iotjs.cmake
@@ -566,12 +566,17 @@ if("${BIN_INSTALL_DIR}" STREQUAL "")
   set(BIN_INSTALL_DIR "bin")
 endif()
 
+if("${INC_INSTALL_DIR}" STREQUAL "")
+  set(INC_INSTALL_DIR "include/iotjs")
+endif()
+
 # Configure the iotjs executable
 if(NOT BUILD_LIB_ONLY)
   set(TARGET_IOTJS iotjs)
   message(STATUS "CMAKE_BINARY_DIR        ${CMAKE_BINARY_DIR}")
-  message(STATUS "BINARY_INSTALL_DIR      ${INSTALL_PREFIX}/bin")
-  message(STATUS "LIBRARY_INSTALL_DIR     ${INSTALL_PREFIX}/lib")
+  message(STATUS "BINARY_INSTALL_DIR      ${INSTALL_PREFIX}/${BIN_INSTALL_DIR}")
+  message(STATUS "LIBRARY_INSTALL_DIR     ${INSTALL_PREFIX}/${LIB_INSTALL_DIR}")
+  message(STATUS "INCLUDE_INSTALL_DIR     ${INSTALL_PREFIX}/${INC_INSTALL_DIR}")
 
   add_executable(${TARGET_IOTJS} ${ROOT_DIR}/src/platform/linux/iotjs_linux.c)
   set_target_properties(${TARGET_IOTJS} PROPERTIES
@@ -581,10 +586,10 @@ if(NOT BUILD_LIB_ONLY)
   target_include_directories(${TARGET_IOTJS} PRIVATE ${IOTJS_INCLUDE_DIRS})
   target_link_libraries(${TARGET_IOTJS} ${TARGET_LIB_IOTJS})
   install(TARGETS ${TARGET_IOTJS} ${TARGET_LIB_IOTJS}
-          RUNTIME DESTINATION "${INSTALL_PREFIX}/bin"
-          ARCHIVE DESTINATION "${INSTALL_PREFIX}/lib"
-          LIBRARY DESTINATION "${INSTALL_PREFIX}/lib"
-          PUBLIC_HEADER DESTINATION "${INSTALL_PREFIX}/include/iotjs")
+          RUNTIME DESTINATION "${BIN_INSTALL_DIR}"
+          ARCHIVE DESTINATION "${LIB_INSTALL_DIR}"
+          LIBRARY DESTINATION "${LIB_INSTALL_DIR}"
+          PUBLIC_HEADER DESTINATION "${INC_INSTALL_DIR}")
 else()
   install(TARGETS ${TARGET_LIB_IOTJS} DESTINATION ${LIB_INSTALL_DIR})
 endif()


daeyeon:
1. The install dir of lib on the `else` branch is different.
2. In my second thought, using constant normal names (e.g: lib and bin) along with CMAKE_INSTALL_PREFIX looks enough.
3. The paths for lib and bin are editable but not for include.
4. INSTALL_PREFIX can be removed if CMAKE_INSTALL_PREFIX is guided.

cmake/iotjs.cmake
```diff
@@ -566,24 +566,24 @@ endif()
 if(NOT BUILD_LIB_ONLY)
   set(TARGET_IOTJS iotjs)
   message(STATUS "CMAKE_BINARY_DIR        ${CMAKE_BINARY_DIR}")
-  message(STATUS "BINARY_INSTALL_DIR      ${INSTALL_PREFIX}/bin")
-  message(STATUS "LIBRARY_INSTALL_DIR     ${INSTALL_PREFIX}/lib")
+  message(STATUS "BINARY_INSTALL_DIR      ${INSTALL_PREFIX}/${BIN_INSTALL_DIR}")
```
akosthekiss, line 569: Either this message is incorrectly stating that the binary install directory is `${INSTALL_PREFIX}/${BIN_INSTALL_DIR}` or lines 580 and 585 miss the `${INSTALL_PREFIX}/`. Which one? (If the prefix is not needed for "bin", why is that, i.e., why is it still needed for libraries?)

rzr: yes this is intentional check the updated commit message:

```
Install directory is prefixed by destdir to install in staging directory.

This change is useful for debian packaging.

Runtime destination rely on RUNTIME_OUTPUT_DIRECTORY property

This was introduced first in jerryscript-project#1134 and then droped in jerryscript-project#1848

This will be helpful for packaging,
or example on Tizen install using RPM macro: "%make_install"
of Debian's debhelper.

Work along CMAKE_INSTALL_PREFIX,
or/and can be overloaded individualy to alternate locations.

Email has been updated for community support.

Forwarded: jerryscript-project#1923
Bug: jerryscript-project#1945
Relate-to: jerryscript-project#1134
Change-Id: Ib969a1d5c2e7bc402b455377fc57a94654aa74dc
Origin: https://github.com/TizenTeam/iotjs/tree/sandbox/rzr/review/install/master
Bug-Debian: https://bugs.debian.org/957364
Update: 2020-07-24
IoT.js-DCO-1.0-Signed-off-by: Philippe Coval rzr@users.sf.net
```

I just verified this patch with other debian packaging patches on current base:

https://github.com/jerryscript-project/iotjs/compare/master...TizenTeam:sandbox/rzr/next/snapshot/master?expand=1

rzr:
unsure this failure is related to this change:
https://travis-ci.org/github/jerryscript-project/iotjs/jobs/736333817

rzr:
This change is released in:
https://tracker.debian.org/news/1184873/accepted-iotjs-10715-1-source-into-unstable/

cmake/iotjs.cmake
akosthekiss: OK, I've reviewed this commit again. And I still cannot grasp what `${INSTALL_PREFIX}` is for. I've updated my local copy of iotjs and searched for `INSTALL_PREFIX`. All I could find was either `CMAKE_INSTALL_PREFIX` (usually in jerryscript-related files) or `INSTALL_PREFIX`, but this latter one only in this file.

My problem is:
- the messages print all install directories prefixed with `${INSTALL_PREFIX}/`
- `install()` sets all destinations without that prefix.

It may happen that `INSTALL_PREFIX` is not set by anyone at all and is only a superfluous path element? But still, the current master is at least consistent in the sense that it prints the expression that it uses for installing. On the contrary, the commit keeps `INSTALL_PREFIX` in the messages but drops them from install destinations.

If `INSTALL_PREFIX` is unused, it should be dropped from the messages, too, IMO.

akosthekiss (changes requested):
Currently, there are two commits in this PR. Please, squash them.

akosthekiss:
OK, I just realized that the second commit may not belong to this here at all but could have infiltrated this PR from #1948 . That's not good either. This PR should contain the relevant commit only.

LaszloLango:
@rzr #1948 is landed, please rebase this PR.

P.S.: sorry about the slow response times, this repository is not very active these days. :(

rzr:
Hi, I've rebased it if It's merged i will rebase debian version on this base.

Next it will be helpful to update jerryscript with this additional change in:
https://github.com/jerryscript-project/jerryscript/pull/4667

LaszloLango (approved):
LGTM

rzr:
Thanks for feedback , anyone available to review and merge this one and then I'll update the debian package 

Cc: @akosthekiss , @zherczeg , @daeyeon , @galpeter, @yadd ...

zherczeg (approved):
LGTM

zherczeg:
Note: there is a change request, until that is resolved, this cannot be landed by github

rzr:
> OK, I just realized that the second commit may not belong to this here at all but could have infiltrated this PR from #1948 . That's not good either. This PR should contain the relevant commit only.

yes that's why I suggested to review the other one 1st
now it's merged so it can be rebased ...

cmake/iotjs.cmake
```diff
@@ -566,12 +566,17 @@ if("${BIN_INSTALL_DIR}" STREQUAL "")
...
+
 # Configure the iotjs executable
 if(NOT BUILD_LIB_ONLY)
   set(TARGET_IOTJS iotjs)
   message(STATUS "CMAKE_BINARY_DIR        ${CMAKE_BINARY_DIR}")
-  message(STATUS "BINARY_INSTALL_DIR      ${INSTALL_PREFIX}/bin")
-  message(STATUS "LIBRARY_INSTALL_DIR     ${INSTALL_PREFIX}/lib")
+  message(STATUS "BINARY_INSTALL_DIR      ${INSTALL_PREFIX}/${BIN_INSTALL_DIR}")
+  message(STATUS "LIBRARY_INSTALL_DIR     ${INSTALL_PREFIX}/${LIB_INSTALL_DIR}")
+  message(STATUS "INCLUDE_INSTALL_DIR     ${INSTALL_PREFIX}/${INC_INSTALL_DIR}")
```
galpeter, line 579: Shouldn't we replace the `INSTALL_PREFIX` with  `CMAKE_INSTALL_PREFIX` ?

rzr: well it was introduced in 99ec37101 may this be changed in other change ?

My install files relies on debhelper DESTDIR there is no need tweak:

https://cmake.org/cmake/help/latest/variable/CMAKE_INSTALL_PREFIX.html

galpeter: So the main thing in this PR would be the new `INC_INSTALL_DIR` variable? If so I'm fine with this change. We'll adapt the INSTALL_PREFIX later on.

rzr: > So the main thing in this PR would be the new `INC_INSTALL_DIR` variable? If so I'm fine with this change.

Thanks the problem is that it's blocked by @akosthekiss change request which can be only acknowledged by himself but I am not able to reach him, can anyone help ?

galpeter: I'm not it.

akosthekiss (approved):
LGTM

rzr:
I guess this can be now merged thx

Issue #1907, opened 2019-07-05
Assertion !uv__io_active(&handle->io_watcher, POLLIN | POLLOUT) in uv__udp_finish_close


renatahodovan:
###### IoT.js version:

```
Checked revision: bc9a5da

Build command: CC=clang-7 \
tools/build.py --clean \
--buildtype=debug \
--compile-flag="-D'IOTJS_ASSERT(x)=assert(x)'" \
--compile-flag=-O2 --compile-flag=-fno-common --no-snapshot \
--compile-flag=-fsanitize=address --compile-flag=-fno-omit-frame-pointer \
--jerry-cmake-param=-DFEATURE_SYSTEM_ALLOCATOR=ON --target-arch=i686 \
--profile=test/profiles/host-linux.profile --jerry-profile=es2015-subset \
--jerry-cmake-param=-DEXTERNAL_COMPILE_FLAGS=-Wno-conversion
```
###### OS:

```
Linux-4.15.0-54-generic-x86_64-with-Ubuntu-18.04-bionic
```

###### Test case:

```javascript
var dgram = require('dgram')
var $ = Object.freeze(dgram.createSocket({ $: $, type: 'udp4' }).bind()).close()
```

###### Backtrace:

```
iotjs: iotjs/deps/libtuv/src/unix/udp.c:106: void uv__udp_finish_close(uv_udp_t *): Assertion `!uv__io_active(&handle->io_watcher, POLLIN | POLLOUT)' failed.

Thread 1 "iotjs" received signal SIGABRT, Aborted.
0xf7fd3939 in __kernel_vsyscall ()
(gdb) bt
#0  0xf7fd3939 in __kernel_vsyscall ()
#1  0xf7c90182 in raise () from /lib/i386-linux-gnu/libc.so.6
#2  0xf7c7a2b6 in abort () from /lib/i386-linux-gnu/libc.so.6
#3  0xf7c7a1c1 in ?? () from /lib/i386-linux-gnu/libc.so.6
#4  0xf7c87fd9 in __assert_fail () from /lib/i386-linux-gnu/libc.so.6
#5  0x082d31b5 in uv__udp_finish_close (handle=0xf5d01f30) at iotjs/deps/libtuv/src/unix/udp.c:106
#6  0x082a9c40 in uv__finish_close (handle=0xf5d01f30) at iotjs/deps/libtuv/src/unix/core.c:234
[...]

PR #1925, opened 2019-07-17
[WIP] Possible fix for closing a frozen handle

DanielBallaSZTE:
So #1907 is a pretty complex issue.
A bound udp socket gets frozen, and on that frozen object a close() is called.
Due to the freeze the object's properties can not be changed, including the native handle, namely ._handle

This fix is not nice IMHO, but solves the above issue.
I'm open to any suggestions on this PR.


DanielBallaSZTE:
Also, this issue might be reproducible with other IO handles.

hs0225:
@DanielBallaSZTE If you can catch an exception when compiling a JavaScript, can you solve it by catching an exception there and throwing an error?

DanielBallaSZTE:
This issue can also be fixed with the help of https://github.com/jerryscript-project/jerryscript/pull/3128 .
Will update the PR soon.

DanielBallaSZTE:
Closing this patch, since a bigger rework would be needed for those handles.

PR #1947, opened 2020-07-24
build: Set noarch as default TARGET_ARCH

rzr:
This will help for debian upgrade

Change-Id: I2991d324b887e340cb676f7de8ea0dda2ea7c050
Forwarded: https://github.com/jerryscript-project/iotjs/pull/rzr
Bug: https://github.com/jerryscript-project/iotjs/issues/1945
Bug-Debian: https://bugs.debian.org/957364
Origin: https://github.com/TizenTeam/iotjs/tree/sandbox/rzr/cmake/review/master
IoT.js-DCO-1.0-Signed-off-by: Philippe Coval rzr@users.sf.net

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -107,6 +107,10 @@ if(EXPERIMENTAL)
 endif()
 
 # Add arch-dependant flags
+if(NOT DEFINED TARGET_ARCH)
+  message(WARNING "Use generic flags since TARGET_ARCH is not defined")
+  set(TARGET_ARCH "noarch")
+endif()
 if("${TARGET_ARCH}" STREQUAL "arm")
   iotjs_add_compile_flags(-D__arm__ -mthumb -fno-short-enums -mlittle-endian)
 elseif("${TARGET_ARCH}" STREQUAL "i686")


CMakeLists.txt
```diff
@@ -107,6 +107,10 @@ if(EXPERIMENTAL)
 endif()
 
 # Add arch-dependant flags
+if(NOT DEFINED TARGET_ARCH)
+  message(WARNING "Use generic flags since TARGET_ARCH is not defined")
+  set(TARGET_ARCH "noarch")
+endif()
```
galpeter, line 113: If we are adding a `TARGET_ARCH` to be defined as "noarch" if it is not defined then the last `else` in the following check will never trigger. Is it not possible to set this via debian rules?

rzr: well I used to overide in debian/rules but I thought it would be better to have it set default upstream to avoid relying on undefined values,  
I could remove the else warning since it's added in my patch.

galpeter: ohh wait. I misunderstood the `if`-s here. The change is ok. If the user specified an incorrect arch then a warning will be printed out bellow, but if there is no arch configured then this will be triggered.

rzr:
yes this is same as other boad patch

PR #1949, opened 2020-07-29
Update JerryScript submodule to post 2.3 version

galpeter:
Based on: #1933

Original author IoT.js-DCO-1.0-Signed-off-by: bence gabor kis kisbg@inf.u-szeged.hu


cmake/iotjs.cmake
```diff
@@ -467,18 +467,18 @@ message(STATUS "EXTERNAL_LIBS            ${EXTERNAL_LIBS}")
 message(STATUS "EXTERNAL_MODULES         ${EXTERNAL_MODULES}")
 message(STATUS "IOTJS_LINKER_FLAGS       ${IOTJS_LINKER_FLAGS}")
 message(STATUS "IOTJS_PROFILE            ${IOTJS_PROFILE}")
-message(STATUS "JERRY_DEBUGGER           ${FEATURE_DEBUGGER}")
-message(STATUS "JERRY_HEAP_SIZE_KB       ${MEM_HEAP_SIZE_KB}")
-message(STATUS "JERRY_MEM_STATS          ${FEATURE_MEM_STATS}")
-message(STATUS "JERRY_PROFILE            ${FEATURE_PROFILE}")
+message(STATUS "JERRY_DEBUGGER           ${JERRY_DEBUGGER}")
+message(STATUS "JERRY_HEAP_SIZE_KB       ${JERRY_GLOBAL_HEAP_SIZE}")
```
akosthekiss, line 471: Is this intentionally inconsistent? JERRY_HEAP_SIZE_KB vs JERRY_GLOBAL_HEAP_SIZE

tools/build.py
```diff
@@ -191,9 +191,9 @@ def init_options():
         action='store_true', default=False,
         help='Enable JerryScript heap statistics')
     jerry_group.add_argument('--jerry-profile',
-        metavar='FILE', action='store', default='es5.1',
+        metavar='FILE', action='store', default='es.next',
```
akosthekiss, line 194: Open question: shall the bump to JerryScript 2.3 also include a bump in the supported ES version?

cmake/iotjs.cmake
galpeter: At the moment I did not want to break the defines/options provided by IoT.js, but I think the name(s) should be normalized in a later change.

tools/build.py
galpeter: Ohh. I've missed this. I've just wanted to change the comment below. Will update this asap to keep it at es5.1 for now.

cmake/iotjs.cmake
akosthekiss: Hm, what I mean is that the name that is printed is not the actual name of the variable that's printed. Shouldn't this be `message(STATUS "JERRY_GLOBAL_HEAP_SIZE       ${JERRY_GLOBAL_HEAP_SIZE}")`?

galpeter: Ahh sorry (was confused by the macro names), true . The original one was also misnamed, but it makes sense to use it the way you mentioned.

galpeter:
@akosthekiss thanks for the input I've updated the PR.

rzr:
For the record this change fix:

https://nvd.nist.gov/vuln/detail/CVE-2020-13649

https://github.com/jerryscript-project/jerryscript/pull/3793

PR #1951, opened 2020-09-02
Use CMAKE_DL_LIBS only when building shared library

alexkalmuk:
Use `CMAKE_DL_LIBS` only if it's actually required, that is when building a shared library.

IoT.js-DCO-1.0-Signed-off-by: Alex Kalmuk alexkalmuk@gmail.com


alexkalmuk:
Ah, my bad, it looks like it's actually required for `uv_dlopen` here in `deps/libtuv/src/unix/dl.c`, not for `--create-shared-lib` option of course. By the way, this file is commented out in `deps/libtuv/cmake/option/option_unix_common.cmake`, so that's why I faced this issue for a target which lacks dlopen() and thought that the problem was in `--create-shared-lib` handling.

alexkalmuk:
I had this issue when linking iotjs binary against libdl. It looks like for nuttx and tizenrt `--buildlib` option is used which disables iotjs binary creation, so libdl is not actually used. So maybe I have to use the same option.

